PR #14, opened 2016-08-06
Adding Introduction, Abstract and Scope

nicola:
I actually dont like it, it should be much simpler and shorter than that, will update


nicola:
Closing this issue since a new PR is coming

PR #22, opened 2016-08-19
wrong hash

hackervera:


hackervera:
dupe of https://github.com/ipld/specs/pull/21

PR #54, opened 2018-02-26
ipld/docs: Adds context about `dag`s

AndrewSB:
I had to look up what a dag was, and after that this doc made way more sense.


vmx (approved):
Thanks a lot!

PR #69, opened 2018-08-11
add go-car impl url

daviddias:
@whyrusleeping can you give a pass through the spec and see if it matches your implementation and update if needed?

### DIFF
--- a/CAR.md
+++ b/CAR.md
@@ -106,3 +106,7 @@ we deal with them?
 So, a topological sort makes it really easy to traverse the CAR, even when
 streaming. However producing a topologically sorted DAG is a bit trickier. Note:
 whatever we choose, it won't have any affect on the asymptotic runtime (memory or time).
+
+## Implementations
+
+- [go](https://github.com/ipfs/go-car)


whyrusleeping:
@diasdavid no, this does not match my impl. This is not a spec, it is a draft, and was never finalized (why is it merged?)

cc @Stebalien

whyrusleeping:
Ah, you merged it @diasdavid. See comments here: https://github.com/ipld/specs/pull/51#issuecomment-389268468

Stebalien:
Yeah, @whyrusleeping's impl is just tar. Not the best solution but it works, exists, and very much doesn't suffer from NIH.

daviddias:
Got it. Added the label WIP to it. https://github.com/ipld/specs/commit/d2f7b50646b9061e4a5e09bda6c00006610a1dd0

rvagg:
The CAR spec was removed in #72 so this is no longer a valid PR so I'm closing this. However, I think we might need to bring it back since it's in active use for Filecoin genesis. I've also created a ZIP CAR that's very similar but conforms to the Datastore interface and I have implementations in both [JavaScript](http://github.com/rvagg/js-ds-zipcar) and [Go](http://github.com/rvagg/go-ds-zipcar), mainly for cross-language test fixtures, but I might spec that up too.

PR #72, opened 2018-09-28
[WIP] Refactor all IPLD specifications

mikeal:
Following up on the [new direction](https://github.com/ipld/ipld/issues/50) I've started to refactor the specifications.

Based on comments by @diasdavid I'm also removing the out of date or deprecated specs as I go.


README.md
```diff
@@ -1,45 +1,97 @@
...
+                |             | |                             |     +----------+
+                +-------------+-+-----------------------------+     |          |
+                |                                             |     |  unixfs  |
+                |          Complex Data Structures            |     |    v1    |
+                |                                             |     |          |
++------------------------------------------------------------------------------+
+|               |                                             |     |          |
+|               |        dag-json         dag-cbor            |     |          |
+|               |                                             |     |          |
+|    Codecs     +---------------------------------------------+ git |  dag-pb  |
```
lanzafame: @mikeal I may be misunderstanding a lot here so sorry in advance but `git` seems to be out of place here? If you were to classify `ipld-eth`, `ipld-btc`, and `ipld-zcash` would they also go here or would they go above `Complex Data Structures`?

lanzafame: I feel like it may be related to this issue: https://github.com/ipld/ipld/issues/16

README.md
```diff
@@ -54,3 +106,149 @@ Small note: If editing the README, please conform to the [standard-readme](https
 ## License
 
 This repository is only for documents. All of these are licensed under the [CC-BY-SA 3.0](https://ipfs.io/ipfs/QmVreNvKsQmQZ83T86cWSjPu2vR3yZHGPm5jnxFuunEB9u) license, © 2016 Protocol Labs Inc.
```
vmx, line 109: License should just be `CC-BY`.

Codecs/DAG-JSON.md
```diff
@@ -0,0 +1,25 @@
...
+All simple types except binary are supported natively by JSON.
+
+Contrary to popular belief, JSON as a format supports Big Integers. It's only
+JavaScript itself that has trouble with them. This means JS implementations
+of `dag-json` can't use the native JSON parser and serializer.
+
+#### Binary Type
+
+```javascript
+{"/": { "base64": String }}
```
vmx, line 18: I'm not sure I like that. I think `/` made sense for links, but I've not sure about if that's a good one for `reserved key space`. I wanted to bring it up, but I don't think this discussion should be part of this PR, but a separate one, once this is merged.

README.md
mikeal: git, eth, btc and zcash are all in the same level as `git` is. these are all pre-existing formats that don't support all the simple types necessary to implement **our** higher order data-structures like `hamt`. That is not to say they can't implement some sort of hamt, they may already, but they can't implement the generic hamt data-structure we are going to spec out. 

perhaps we should label that section `ipld-git` and put a few more in the same box like `ipld-btc`.

README.md
mikeal: This was unchanged from the prior version. Did the new IP Policy land yet @ianjdarrow?

Codecs/DAG-JSON.md
mikeal: So, I definitely only want to reserve a single address space. If we add a new address space for each type then we are growing the list of forms that are no longer representable.

However, it has a come up a few times that `{/:}` was a hack and wasn't something we intended to be permanent. If we have a better form we can migrate to for both links and other types I need to represent in JSON that's great but I can't think of anything better than `{/:}` at the moment.

README.md
lanzafame: > perhaps we should label that section ipld-git and put a few more in the same box like ipld-btc.

This would make it clearer :+1:.

To clarify my understanding a little further, if I were to implement an 'ipld-format' similar to `ipld-git` now, where should it go? Should I be looking to implement something in the way that `ipld-git` or `ipld-btc` does?

mikeal: @lanzafame it depends :)

But just to make this a little clearer. The decoder APIs, like the ones we use in IPFS, don't really differ between formats with the full data model and those without. The API for that layer will remain simple so that it's easy to implement future codecs like `ipld-git` for formats that already exist.

However, if you want to build a codec that supports the full data model so that you can use all the complex data-structures we're building on top of that (like `hamt`) then you'll want to support the full data model in your encoder. This just means being able to encode and decode a number of standard types.

daviddias: Not trying to paint a bikeshed, but can we s/Codecs/Formats? We've been calling these formats for a while now.

README.md
```diff
@@ -1,45 +1,97 @@
...
+       IPLD Dependency Graph
+
++---+                          +-----+        +---+
+|CID+-----------+-------------->Block+-------->Raw|
++---+           |              +--+--+        +---+
+         +------v-------------+   |
++----+   |Links (Conceptually)|   |
+|Path|   +------+-------------+   |             +-----------+
++-+--+          |                 +------------->Replication|
+  |    Codecs   |                 |             +-----------+
```
daviddias, line 56: Same, s/Codecs/Formats

README.md
lanzafame: Speaking only from personal experience, I found that `Format` sent me down a path to a different understanding to that which `Codec` does. So I found the rephrasing to `Codec` useful. Just my 2 cents.

eocarragain: It would be great to have some definitions and examples to illustrate the IPFS/IPLD usage of: codecs; formats; types/typing; schema. I've seen them all used in IPLD discussions but depending on your background these can have different meanings and connotations that don't exactly match IPLD/IPFS usage.

lanzafame:
There is the start of a glossary at the bottom of this PR.

README.md
mikeal: @diasdavid the CID spec uses "Codec" consistently and that's the most consistent use I could find. I've seen format used interchangeably with codec in a few places but I've rarely seen codec refer to something *inside* a format.

If we are borrowing from prior definitions then "format" would need to be inside of "codec" as codecs traditionally compress data *into* a "format."

warpfork: I've been calling these sorts of things "IPLD-Adjunct" formats: data that we can read and address as part of the merkleforest, but it requires special code to do so, rather than being an "IPLD-native" system which works only with our totally reusable cbor/json/etc systems.

warpfork: I'll toss this into the ring: "Representations".

I did some research spike code in a vacuum last week and that's what I came up with.

ISTM this might fit well because it allows us to consider representations that aren't actually serial at all.  For example, when I have an in-memory `Node` interface hanging onto a bunch of data... it's in IPLD Data Model.  I can query and handle it like IPLD.  But it's not in CBOR... yet.  This in-memory is a valid "representation"; like any other representations, e.g. JSON, it has to be converted to a serializable representation (namely, CBOR) to be shared and hashed, but it's still definitely a representation of IPLD.

mikeal: @warpfork so, you're recommending "representation" for the term used to describe in-memory de-serialized nodes? 

I like it, I don't think we've had a consistent term for that yet.

mikeal:
Ok, just pushed lots of fixes.

I'd like to move this towards a merge. It isn't perfect, but it's a lot closer to how we're now talking about IPLD than the current state of the repo.

Codecs/DAG-JSON.md
mikeal: I'm going to table this conversation for now. `dag-json` is still a WIP and I have a pretty dramatic new proposal for the format I'm going to start on after this gets merged.

README.md
mikeal: I'm going to resolve this for now as I still haven't heard from @ianjdarrow and I'm not comfortable changing the license until I do.

README.md
mikeal: New visualization is better :)

mikeal:
@diasdavid I disagreed with the codec/format terminology and you haven't responded to my prior reply so I resolved that conversation *for now*. If you want to continue to discuss we can start a new thread, PR, or we can just have a quick call to resolve it.

Codecs/DAG-CBOR.md
```diff
@@ -0,0 +1,37 @@
...
+unpathable (TODO: drop this? IMO, we should support path escaping out of the
+box).
+
+## Canonical DagCBOR
+
+Canonical DagCBOR should:
+
+1. Use no tags other than the CID tag (42). Other tags may be lost in
+   conversion.
+2. Should use the canonical CBOR encoding and field ordering. Other orderings
```
vmx, line 34: There should be a section about numbers. I'd like to have specified what the right thing for cases like https://github.com/ipld/interface-ipld-format/issues/9#issuecomment-431029329 are.

IPLD-Path.md
```diff
@@ -0,0 +1,20 @@
+# [WIP] IPLD Path v1
+
+An IPLD Path is a string identifier used for deep references into IPLD
+graphs.
+
+IPLD Path's are constructed following the same constraints as [URI Paths](https://tools.ietf.org/html/rfc3986#section-3.3).
```
vmx, line 6: Does it imply that `#` is also reserved?

vmx, line 6: Do we also percent encode for escaping slashes?

IPLD-Path.md
```diff
@@ -0,0 +1,20 @@
...
+
+Similarly, the string `?` is reserved for future use as a query separator.
+
+# Path Resolution
+
+Path resolution is broken into two parts: full path resolution and block level resolution.
+
+Block level path resolutionis defined by individual codecs.
+For most common codecs each path segment is a property lookup.
```
vmx: What do you mean by "property lookup"?

README.md
```diff
@@ -54,3 +107,153 @@ Small note: If editing the README, please conform to the [standard-readme](https
...
+| |                              |  |                  |
+| +------------------------------+  |                  |
+|                                   |                  |
++-----------------------------------+------------------+
+```
+
+### IPLD Path
+
+A string identifier used for deep references into IPLD
+graphs. Follows similar escape and segmentation rules as URI Paths.
```
vmx, line 211: Link to IPLD Path spec.

README.md
```diff
@@ -1,45 +1,98 @@
...
-- [**Roadmap**](/ROADMAP.md)
-- **Specifications:**
-  - ![](https://img.shields.io/badge/status-draft-yellow.svg?style=flat-square) [`IPLD`](/IPLD.md) - spec about the data model, pointers and link formats
-  - ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) `IPLD Selectors` - spec about simple language to select multiple unknown nodes in a graph
-  - ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) `IPLD Transformations` - spec about the language to trasform an IPLD graph into another
-  - ![](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square) [`CID (Content IDentifier)`](https://github.com/ipld/cid)
-  - ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) [`IPLD Formats`](https://github.com/ipld/interface-ipld-format) - interface definition for adding support to different formats
-  - ![](https://img.shields.io/badge/status-draft-yellow.svg?style=flat-square) [`CAR`](/CAR.md) - Content Addressable Archives
++---+                          +-----+        +---+
+|CID+-----------+-------------->Block+-------->Raw|
```
vmx: I still think the arrow between "Block" and "Raw" should be the other way round. A "CID" and "Raw" data form a "Block".

vmx (commented):
A few comments, though I can also open issues if you want to merge without making any changes.

Codecs/DAG-CBOR.md
mikeal: This was pulled in from the Dag-CBOR PR. Regardless, I'd like to merge this and then log an issue to make this improvement. I think there will end up being some discussion around this particular point.

IPLD-Path.md
mikeal: Yes. The fact is, if we don't go "all in" on URI path semantics we can't use existing URI path parsers and escape libraries. I also think that any deviation we make will come back to bite us later when we do more browser integration.

Also, you can still have all of these special characters in keys that are referenced by paths as long as you escape them properly.

mikeal: Yes.

IPLD-Path.md
mikeal: The path segment will match a property of the decoded object. Would "attribute" be a better term than "property" here?

README.md
mikeal: Maybe I should change this to read "Raw Block." What I'm intending to reference is a full raw block, not just a CID with a "raw" codec value or just a binary blob.

IPLD-Path.md
vmx: I would just remove that sentence. I don't see why this really matters. For CBOR the path is a the nested property.

If you keep it, I'd go for "property".

IPLD-Data-Model-v1.md
```diff
@@ -0,0 +1,18 @@
+# [WIP] IPLD Data Model
+
+## Simple Types
+
+* Boolean
+* Null
+* String
+* Integer
+  * Must support BigInt
```
warpfork: I'd rather avoid promising bigint *as a core type* and revisit that in more detail later.

mikeal: This was something @jbenet was pretty adamant about. We don't want to lose precision when mutating a graph from a language or codec that didn't implement the type completely.

IPLD-Path.md
warpfork: +1 to being a little confused by this segment and possibly just skipping it.

mikeal: Ok, I'll remove it.

IPLD-Path.md
```diff
@@ -0,0 +1,20 @@
+# [WIP] IPLD Path v1
+
+An IPLD Path is a string identifier used for deep references into IPLD
+graphs.
+
+IPLD Path's are constructed following the same constraints as [URI Paths](https://tools.ietf.org/html/rfc3986#section-3.3).
+
+Similarly, the string `?` is reserved for future use as a query separator.
+
+# Path Resolution
```
warpfork, line 10: (Sorry, :bike: :building_construction: warning) Fwiw, I would refer to this as "traversal" rather than resolution.

And would rather refer to link traversal as loading another IPLD object (or node), rather than use the "block" terminology.  I think the "block" terminology is super confusing when you try to think about IPLD outside of IPFS; "block" terms come from IPFS's chunking, and IPLD should make just as much sense when applied to some plain file or HTTP stream or etc.

IPLD.md
```diff
@@ -1,553 +0,0 @@
...
-- **IPLD Canonical Format**: a deterministic description on a serialized format that ensures the same _logical_ object is always serialized to _the exact same sequence of bits_. This is critical for merkle-linking, and all cryptographic applications.
-
-## Intro
-
-### What is a _merkle-link_?
-
-A _merkle-link_ is a link between two objects which is content-addressed with the _cryptographic hash_ of the target object, and embedded in the source object. Content addressing with merkle-links allows:
-
-- **Cryptographic Integrity Checking**: resolving a link's value can be tested by hashing. In turn, this allows wide, secure, trustless exchanges of data (e.g. git or bittorrent), as others cannot give you any data that does not hash to the link's value.
-- **Immutable Datastructures**: data structures with merkle links cannot mutate, which is a nice property for distributed systems. This is useful for versioning, for representing distributed mutable state (eg CRDTs), and for long term archival.
```
warpfork, line 20: I dunno if some of this lives elsewhere already so this is just a dedup-removal, but if not, I think these couple of lines are pretty gold for the very high level motivation explainers.

IPLD-Path.md
mikeal: The intent of the section was to cover resolution and not just traversal. The part I'm missing, which I'll go in and add, is that once the path is traversed we return the **representation** of the value for that path.

We've defined Block consistently in the glossary in the README and actually plan on bringing the block level projects into the IPLD org rather than having them at the IPFS layer. We need a consistent term for "CID + Binary Data Value" and terms like "node" are already used in too many places to use them here.

README.md
warpfork: I'd actually propose "IPLD Representation" for *all of*: dag-cbor, dag-json, and go-inmem, etc.  I can traverse, etc, each of those with the standard IPLD Node interfaces.  They'll all be confined by the IPLD Data Model core types.

The first two just happen to be an IPLD *Serial* Representation.

IPLD-Data-Model-v1.md
warpfork: :grimacing:

IPLD.md
mikeal: We have new educational materials for decentralized data-structures being written for ProtoSchool that cover most of this. There's also the DWeb Book that Matt is re-writting.

IPLD-Data-Model-v1.md
warpfork: I'm just gonna come out and say right now I'm uncomfortable promising to actually implement that.  Computers tend to have this weird hangup about numbers being made out of countable number of bits.  Fighting it tends to not result in code that's fast or APIs that anyone is happy about.

I'm afraid if we commit to this, then we're going to end up with implementations that ignore this part of the standard *anyway* and those will become the more popular ones, because nobody likes to program with `MySpecialBigInt` type wrappers rather than their language's native `int`.  And people will constantly cast things to their language's `int` anyway when doing operations.

We can and should have some definitions of BigInt that we support well up at the "Types" level, higher.  But I'm extremely dubious that doing it at the *Data Model* layer is even viable.

mikeal: It's interesting to consider here what guarantees or requirements we can actually make at this spec level.

I'm comfortable saying that we can make requirements of the *format*. But when it comes to the **representation** codecs give in each language there's a limit to the guarantees we can make. Obviously codecs should make every effort possible to support the type but it may not always be possible.

For `dag-json` in JavaScript I'm confident I can support this even though the language is not particularly helpful. JSON already supports BigInt theoretically, it's JavaScript's number type that is problematic. But since I own the `dag-json` serializer/deserializer I can look for BigInt types and for serialized numbers that are too large and represent them with BigInt types rather than regular numbers types.

README.md
```diff
@@ -1,45 +1,98 @@
...
-- [**Roadmap**](/ROADMAP.md)
-- **Specifications:**
-  - ![](https://img.shields.io/badge/status-draft-yellow.svg?style=flat-square) [`IPLD`](/IPLD.md) - spec about the data model, pointers and link formats
-  - ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) `IPLD Selectors` - spec about simple language to select multiple unknown nodes in a graph
-  - ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) `IPLD Transformations` - spec about the language to trasform an IPLD graph into another
-  - ![](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square) [`CID (Content IDentifier)`](https://github.com/ipld/cid)
-  - ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) [`IPLD Formats`](https://github.com/ipld/interface-ipld-format) - interface definition for adding support to different formats
-  - ![](https://img.shields.io/badge/status-draft-yellow.svg?style=flat-square) [`CAR`](/CAR.md) - Content Addressable Archives
++---+                          +-----+        +---------+
+|CID+-----------+-------------->Block+-------->Raw Block|
```
vmx: I think you explained it to me two times already. I still don't get the direction of the arrow. But I don't want to block things because of such a small thing. If everyone else gets the idea, that's good enough for me :)

vmx (approved):
For me it's ready to be merged, further changes can be discussed on separate issues.

README.md
mikeal: It's like a reverse dependency graph. One thing is required to build the next thing. The only purpose of the "Raw Block" illustration is to note that raw blocks don't have a codec but are still part of replication.

vmx: Oh, I think I got it now. You mean a "Raw Block" as a "block with the raw codec". It doesn't contain links, hence it doesn't belong to the other codecs. So it could also be drawn like this:

```
+---+                                        +-----+
|CID+-----------+--------------------+-------+Block|
+---+           |                    |       +-----+
        +-------v------------+  +----v----+
        |Links (Conceptually)|  |Raw Block|
+----+  +-------+------------+  +--+------+ 
|Path|          |                  |          +-----------+
+----+          |                  +---------->Replication|
  |    Codecs   |                  |          +-----------+
+-v-------------v------------------+--+
|                                     |
```

IPLD-Path.md
warpfork: Fair enough re: some other terms are already too overloaded to use.

I'm still dubious about "block" though.  You don't stack them on top of each other; they're not very blocklike :)  "Block" comes from taking a large stream of bytes and *cutting it apart* into *smaller blocks*... which isn't a thing IPLD does; IPLD is a format for building merkle-friendly data structures.  The term "block" term wouldn't have evolved if we had designed IPLD in a vacuum without first having spent a lot of time thinking about the unrelated problem of the practicalities of trying to break apart large binary files for efficient storage.

/2c

vmx: @warpfork  I see block coming from block chain. You link up things and chain them together.

Though I agree and I'd be happy to have a better name. Though we should have this discussion on a separate issue. .o0(IPLD Atom, the smallest unit IPLD works with).

IPLD-Data-Model-v1.md
warpfork: I'm just not convinced that this either A) solves real problems B) in *a way that people will actually use, for human and API usability reasons* and C) is worth the effort in making thoroughly nonstandard parsers for any of the underlying serial representations that we use which are already specified and lack this capability (e.g. all of them).

I definitely buy that we can write custom parsers that keep (for example) reading another base ten ascii character in JSON and in its internal model racks up a value that's well beyond the 4 billion range.

I *don't* buy that it's plausibly usable to then store that number in some blessed-at-the-Data-Model-level BigInt type that's exposed to the user and which will rule all BigInt types.  The prior art for BigInt types is large; there's many serial implementations of them; and I don't think we should try to wade into that holy war at the Data Model level.  (The "Types" level, above?  Sure.  But not the Data Model, where we'll be stuck with it in perpetuity and have limited options supporting variations.)

I don't buy that we'll have very friendly APIs if any integer is implicitly `Either<int,BigInt>` -- that might fly in a fairly untyped language like javascript, but it'll still produce utterly bizarre runtime errors when users actually encounter it; and it's not going to fly well *at all* in more strongly typed languages.  In particular, this will yield a fractal of trouble if we make other choices like allowing numeric map keys; `map[Either<int,BigInt>]interface{}` is :nauseated_face:   I don't think we've specified whether numeric map keys are allowed yet, so we certainly can make other choices to avoid this problem... but I think it's critical to make sure we've fully considered these kinds of kinds of choices before committing to a definition of numerics will require us to allow bigints anywhere we allow ints.

I don't buy that IPLD having native BigInt everywhere is either necessary nor (more importantly) *sufficient* for making lossless transformations of IPLD data the norm.  As long as people write code in languages that don't have native BigInt, they'll be writing bindings of IPLD to types like `struct{ x int }`, doing their logical transformations on those structures, and emitting IPLD again.  In such usage, it doesn't matter in the slightest if we've gone through tons of work to make native bigints; the least expansive primitive types used anywhere in the transformation code are going to be the dominant factor; and for the forseeable future that means anyone designing a durable protocol that works smoothly in many modern languages is still going to want to consciously constrain themselves to int64 or int32.  (http://reasonablypolymorphic.com/blog/protos-are-wrong/ has some commentary similar to this; namely, it's rare in practice to see programs which inspect only one field out of a protobuf type; the vast majority of programs that people write will load, transform in that language's types, perform some computation, and emit again; this is the path that we should concern ourselves with.)  So, I would argue that the necessary thing for lossless transformations of IPLD data is more to do with the implementation and *having a language* that uses bigints for its basic numeric types than anything else.

---

tl;dr; I think we'd be better off saying "implementations SHOULD support the largest integer ranges possible and MUST support at least int32 ranges; protocols SHOULD NOT attempt to use any kind of "INT_MAX" values as sentinels"... and trying to solve infinite portability problems at the level where they're actually most likely to be problematic: transformation code and the languages its written in.

mikeal:
Ok, I think we're ready to merge now. All pending discussions have been closed.

mikeal:
What teams have merge access in this repo?

vmx:
@mikeal I've added the IPLD JS Team.

PR #73, opened 2018-10-30
[WIP] dag-cbor: add a draft spec

Stebalien:


mikeal:
Thanks!

vmx:
Not sure if it belongs to this spec. Should it also contain information on how to encode numbers (for more information about problems with numbers see https://github.com/ipld/interface-ipld-format/issues/9#issuecomment-431029329)?

Stebalien:
@vmx kind of. Canonical CBOR says "use the smallest representation".

vmx:
@Stebalien: smallest representation for integers. We need to specify if we treat `5.0` as in integer or not.

Stebalien:
Ah, you're right. I assumed that also applied to floats.

vmx:
This one got already merged with https://github.com/ipld/specs/pull/72.

PR #75, opened 2018-11-01
Proposal: GraphSync (B)

jbenet:
This is an alternative proposal for graphsync.

Status of this proposal:
- This was written around 2018-10-16 ([video presentation](https://drive.google.com/file/d/1NbbVxZQFKXwW6mdodxgTaftsI8eID-c1/view))
- This document is unfortunately far from complete. :( I want to finish this by EOQ.
- But this document provides enough information for an implementation to be made by someone who has already implemented bitswap (or understands it well).
- It relies heavily on an understanding of bitswap as it is now. It likely won't be useful to people without a good understanding of how Bitswap works at the moment.
- This requires IPLD Selectors to exist and be implemented.


jbenet:
This PRs on top of #74

jbenet:
BTW: on another look, this doc is crap right now. (sorry!) I needs a lot of work to be useful as a proposal. But i think it is meaningful to people who have implemented Bitswap, and they may be able to write an implementation from it.

jbenet:
See the video presentation here: https://drive.google.com/file/d/1NbbVxZQFKXwW6mdodxgTaftsI8eID-c1/view

graphsync/graphsync.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Downloading Package Dependencies
+
+- Give me everything within `<hash>/foo/v1.0.0`
+
+### Loading content from deep within a giant dataset
+
+- Give me the nodes for the path `<hash>/a/b/c/d/e/f/g`
+
+### Loading a large video optimizing for playback and seek
```
mikeal, line 133: The problem with these approaches that say "give me [indefinite number] of blocks" is that they don't parallelize well.

The client making a request of indefinite size, by definition, doesn't know the size and shape of the request. At each layer of depth we lose the ability to grab that content from other peers. Depending on the shape of the graph and the network conditions of the different peers this approach is rarely the fastest and for many edge cases ends up being incredibly punitive. If we instead just focus on requesting more blocks concurrently once we have the knowledge to do so we actually *gain* concurrency vectors at each depth of the tree and can optimize our requests across many peers.

The inability to request a large number blocks *serially* from a single peer should not be a source of performance issues right now. Currently, we think the largest bottlenecks are:

* inability to grab a deep node without a roundtrip for every branch.
* lack of parallelization when grabbing blocks from a peer (which, on the JS side at least, is apparently just a bug).
* lack of storage knowledge for each CID (basically, we need to note when we have the full graph for a CID so that when the client is syncing an updated graph it knows if it needs to continue traversing down a particular path).

The rough outline of a fix for this was:

* Create a new RPC call that asks for the merkle proof of a given PATH.
* Request large numbers of blocks in parallel from known peers (I think someone said this isn't actually a problem on the Go side).
* Create a layer on top of BlockStore that has a boolean index for whether or not the local repo has the entire graph referenced by that CID.

That's enough to drastically improve over what we have today and avoids a lot of very complex logic when trying to spread out these indefinite requests over multiple peers.

graphsync/graphsync.md
```diff
@@ -0,0 +1,151 @@
...
+    int32 id = 1;       // unique id set on the requester side
+    bytes selector = 2; // ipld selector to retrieve
+    bytes extra = 3;    // aux information. useful for other protocols
+    int32 priority = 4;	// the priority (normalized). default to 1
+    bool  cancel = 5;   // whether this cancels a request
+  }
+
+  message RequestList {
+    repeated Request reqs = 1; // a list of requests
+    bool full = 2;			   // whether this is the full RequestList. default to false, for incremental requests.
```
Stebalien, line 73: Should probably call this something like `replaceRequests`.

graphsync/graphsync.md
```diff
@@ -0,0 +1,151 @@
...
+
+  message Request {
+    int32 id = 1;       // unique id set on the requester side
+    bytes selector = 2; // ipld selector to retrieve
+    bytes extra = 3;    // aux information. useful for other protocols
+    int32 priority = 4;	// the priority (normalized). default to 1
+    bool  cancel = 5;   // whether this cancels a request
+  }
+
+  message RequestList {
```
Stebalien, line 71: We can probably remove this indirection and embed the requests direly in the message.

graphsync/graphsync.md
```diff
@@ -0,0 +1,151 @@
...
+- `response` - the content sent from `responder` to `requester` fulfilling a `request`.
+- `requester` - the peer which initiates a `request` (wants content).
+- `responder` - the peer receiving a `request`, and providing content in a `response` (provides content).
+- `request process` - a request and its fulfillment is a sub-process, a procedure call across peers with the following phases (at a high level):
+  - (1) The `requester` initiates by sending a request message (`req`) to the `responder`, specifying desired content and other request parameters.
+  - (2) Upon receiving a request message, the `responder` adds the request to a set of active requests, and starts processing it.
+  - (3) The `responder` fulfills the request by sending content to the `requester` (the `response`) .
+  - (4) The `responder` and `requester` can terminate the request process at any time.
+  - Notes:
+    - We are explicitly avoiding the `client-server` terminology to make it clear that `requester` and `responder` are "roles" that any peer might play, and to avoid failling in the two-sided client-server model of the web..
```
warpfork, line 32: :heart: :heart_decoration: :heart:  Words matter.  :heart: :heart_decoration: :heart:

Stebalien:
Merging as this is the current "working draft".

PR #85, opened 2018-12-20
add a document covering requirements on IPLD

Stebalien:
This document covers parts of IPLD that must exist and must not change to allow building things like type systems, programming languages, etc. on top of IPLD.

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -6,7 +6,7 @@ ensure the success of future improvements (especially type systems).
 **Definitions**
 
 * Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
-* Patch: A piece of an IPLD DAG. Blocks contain patches.
+* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
 * Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
 * Link: A link is an IPLD Node that points to another IPLD Node.
 * Path: A paths a human readable pointer to an IPLD Node.
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -41,10 +41,12 @@ definition of `foo` would not be a valid IPLD block:
 
 ```
 var foo = {
+  // points outside of the current block, into the parent's "baz" field.
   "baz": {"/": "../../baz"}
 }
 var bar = {
   "foo": CidOf(foo),
+  // `/foo/baz` points here.
   "baz": "something"
 }
 
@@ -53,6 +55,9 @@ Resolve("/ipld/${CidOf(bar)}/foo/baz/")
 ```
 
 For the same reason, IPLD links can't rely on an authority (e.g., a blockchain).
+
+Note: Links like this can still be encoded at the application level but they
+won't be handled by the IPLD resolver (and won't get the special "link" type).
 
 **Motivation:** IPLD needs to be easy to reason about.
 
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -1,0 +1,155 @@
+# IPLD Requirements
+
+This document outlines parts of IPLD that should and should not be changed to
+ensure the success of future improvements (especially type systems).
+
+**Definitions**
+
+* Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
+* Patch: A piece of an IPLD DAG. Blocks contain patches.
+* Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
+* Link: A link is an IPLD Node that points to another IPLD Node.
+* Path: A paths a human readable pointer to an IPLD Node.
+
+## Linked
+
+IPLD must support linking to any IPLD node (even if the node is in the middle of
+a block). That is, IPLD must support arbitrary IPLD paths in links.
+
+**Motivation:** Considering this in the context of programming languages, not
+being able to *store* a pointer to a struct *inside* of another struct would be
+severely limiting.
+
+NOTE: We don't currently support arbitrary paths but, in the context of
+programming, we really need to.
+
+## Immutable
+
+IPLD links must be immutable. We'll likely define a mutable link spec on top of
+IPLD but there needs to be an immutable layer at the bottom.
+
+**Motivation:** *Having* an immutable layer is important for a lot of analysis,
+memoization, type checking, etc.
+
+## No Non-Local Reasoning
+
+An IPLD block should never be interpreted in the context of *anything* not
+contained in the block (and CID).
+
+For example, assuming we add support for relative links, the following
+definition of `foo` would not be a valid IPLD block:
+
+```
+var foo = {
+  "baz": {"/": "../../baz"}
+}
+var bar = {
+  "foo": CidOf(foo),
+  "baz": "something"
+}
+
+// resolution throug block `foo` depends on block `bar`.
+Resolve("/ipld/${CidOf(bar)}/foo/baz/")
+```
+
+For the same reason, IPLD links can't rely on an authority (e.g., a blockchain).
+
+**Motivation:** IPLD needs to be easy to reason about.
+
+**Caveat:**
+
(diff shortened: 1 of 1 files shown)


Stebalien:
This covers a lot of pretty obvious requirements but also some interesting requirements that only came up in the context of programming (e.g., array slicing).

REQUIREMENTS.md
```diff
@@ -0,0 +1,155 @@
+# IPLD Requirements
+
+This document outlines parts of IPLD that should and should not be changed to
+ensure the success of future improvements (especially type systems).
+
+**Definitions**
+
+* Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
+* Patch: A piece of an IPLD DAG. Blocks contain patches.
```
mikeal: Can we find another term for this? Patch is commonly used as a verb for selective edits of a structure. There are even existing threads about adding a "patch" API to the existing dag API's in IPFS.

REQUIREMENTS.md
```diff
@@ -0,0 +1,155 @@
...
+* Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
+* Link: A link is an IPLD Node that points to another IPLD Node.
+* Path: A paths a human readable pointer to an IPLD Node.
```
mikeal, line 12: I like the distinction this is starting to draw between blocks and nodes/links/paths. These lines have been pretty blurry in prior docs and I like this way of drawing the distinction.

REQUIREMENTS.md
```diff
@@ -0,0 +1,155 @@
...
+Resolve("/ipld/${CidOf(bar)}/foo/baz/")
+```
+
+For the same reason, IPLD links can't rely on an authority (e.g., a blockchain).
+
+**Motivation:** IPLD needs to be easy to reason about.
+
+**Caveat:**
+
+We *may* want to relax this if we want to move schemas into separate,
```
mikeal, line 72: We can break this into layers to make it clearer. For instance, you *can* decode the block with nothing but the block data and the CID (which tells you what codec to use). In this sense, "interpret" is a little loaded because you *can* turn it into a somewhat understandable in-memory structure, you just might need to fetch additional linked data in order to do more than that, like traverse paths through it.

REQUIREMENTS.md
```diff
@@ -0,0 +1,155 @@
...
+   This makes it easier to correctly and securely implement some graph
+   algorithms, even on potentially untrusted data in a distributed system.
+2. Consistency: Without complex hacks, it's impossible to create link cycles
+   *between* hash-linked blocks. Allowing link cycles one one level (e.g., in
+   relative links within a single block) but not on a larger scale (between
+   blocks) is inconsistent (even though there's nothing we can do about it).
+
+## Stable Pathing
+
+An IPLD path always means the same thing, everywhere, every time. Importantly,
```
mikeal, line 99: We may want to qualify this to a particular version of "Path." We're talking about a few iterations on pathing that *will* change how it works.

REQUIREMENTS.md
```diff
@@ -0,0 +1,155 @@
...
+resolution works. They should be able to use alternative path resolution
+algorithms for *non*-IPLD paths (e.g., IPFS paths) but those should be built
+on top of IPLD.
+
+**Motivation:** Deterministic computations on top of a IPLD need to produce the
+same result every time.
+
+## Link Transparent Pathing
+
+Path resolution must transparently traverse links.
```
mikeal, line 110: 💯

REQUIREMENTS.md
```diff
@@ -0,0 +1,155 @@
...
+
+**Motivation:** Deduplication and composability. If links are *not* transparent,
+programmers will have to either make many small objects (lots of hashing) or
+inline data into large objects (lots of duplication and copying).
+
+## Multicodecs Are Not Types
+
+It's impossible to understand IPLD data at a *structural* level if we don't know
```
mikeal: This is very clarifying, I think we might want to move it up to before the "non-local resolution" section because it's useful for understanding that section.

mikeal:
Overall, this is great and I love it. Comments in line :)

REQUIREMENTS.md
Stebalien: We can probably just use subgraph.

REQUIREMENTS.md
Stebalien: This came out of a discussion with @jbenet in London (which never made it to GitHub).

REQUIREMENTS.md
Stebalien: I mean a path in the `/ipld` namespace. When we add new resolution algorithms, we'll have to introduce a new prefix.

Stebalien:
Thanks for the review!

Stebalien:
(fast review :heart:)

REQUIREMENTS.md
vmx: Or Fragment (coming from HTML: https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment).

REQUIREMENTS.md
vmx: We'd need to clarify how to represent namespaces. Would every path start with `/ipld/`?

REQUIREMENTS.md
```diff
@@ -0,0 +1,155 @@
...
+
+An IPLD block should never be interpreted in the context of *anything* not
+contained in the block (and CID).
+
+For example, assuming we add support for relative links, the following
+definition of `foo` would not be a valid IPLD block:
+
+```
+var foo = {
+  "baz": {"/": "../../baz"}
```
vmx, line 51: It took me a while to grasp it. So perhaps adding a comment like `// points outside the current block` might make things clearer.

warpfork: +1 to more comments or other way to communicate the focus here.  It's also not super obvious that we're looking at a Link (IMO -- I spend most of my time looking at code that doesn't have links encoded in JSON, right?  So).

I'd maybe like to salt on a little more examples here.  It's bad if a Link wanted to point out of the currently hash-covered scope... but if some piece of data wants to declare it has semantic content with the string "../.." (e.g. symlinks in unixfsv2), that's fine (or rather, it's a tricky situation to carefully solve... but somewhere else!  Not an IPLD data model problem).

REQUIREMENTS.md
```diff
@@ -0,0 +1,155 @@
...
+
+Working through this, I realized we have a few things we really need to finish a few things before we can
+call IPLD ready.
+
+* **Path links.** Pointers that can only point to objects at block boundaries
+  are useful but severely gimped. We've been fine up till now because we
+  generally don't *edit* complicated datastructures but this will change.
+  ([#83](https://github.com/ipld/specs/issues/83))
+* **Slice links.** For the same reason, we really should support
+  `/ipld/QmID/start..stop` as a syntax for slicing an array. Most programming
```
warpfork, line 155: Is this not more of a selectors thing?

I thought links always pointed to one node.

warpfork:
:heart:

REQUIREMENTS.md
Stebalien: It is more of a selector thing. My worry is that most programming languages support this so we should consider it as well.

On the other hand, we could implement this as a higher level type.

REQUIREMENTS.md
Stebalien: Regardless of how we do it, we'd still lose the self-contained nature of IPLD blocks. We may find we don't have a choice but we need to be careful.

But I agree we should *always* maintain the ability to understand a block, even if it's just as `(type CID, content)`.

vmx (approved):

Stebalien:
Merging. We can iterate on this in-tree.

REQUIREMENTS.md
warpfork: +1 for higher level type.  That seems to resonate well with the general concept of "advanced layouts" I'm carrying around in my head.  An "advanced layout" object that has some links and offset information in it and acts like an array when viewed through `typed.Node` rather than directly as the Data Model layer data is :ok_hand:

PR #89, opened 2019-02-08
graphsync: rework protobufs a bit

Stebalien:
1. Remove the indirection (`RequestList`).
2. Rename `full` to `completeRequestList` and document it. I'm still not happy with the name but it's hopefully a bit less confusing.
3. Avoid abbreviated names.

### DIFF
--- a/graphsync/graphsync.md
+++ b/graphsync/graphsync.md
@@ -68,11 +68,6 @@ message GraphsyncMessage {
     bool  cancel = 5;   // whether this cancels a request
   }
 
-  message RequestList {
-    repeated Request reqs = 1; // a list of requests
-    bool full = 2;			   // whether this is the full RequestList. default to false, for incremental requests.
-  }
-
   message Response {
     int32 id = 1;     // the request id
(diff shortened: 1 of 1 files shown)


hannahhoward:
Does completeRequestList have meaning in graphsync land? Particularly at the moment? (where there's no broadcasting of request lists -- the current interfaces are direct peer to peer)

hannahhoward:
But non-blocking for me, otherwise LGTM.

hannahhoward (approved):

Stebalien:
> Does completeRequestList have meaning in graphsync land? Particularly at the moment? (where there's no broadcasting of request lists -- the current interfaces are direct peer to peer)

It means "throw away all requests not listed in this message". Really, it _shouldn't_ be necessary if everyone is behaving correctly but it exists to get of funky states. For example, if you send me a response for some long-dead request, I may decide to send you a message with all open requests and `completeRequestList` set.

Note: For now, we just need to _handle_ this on the receiver's side. We never need to set it.

(we can also rework it if it isn't appropriate).

Issue #90, opened 2019-02-08
Selector Verify Interface


Stebalien:
I'm thinking something like:

```go
type Verifier interface {
	// Expects returns the next expected CID and a boolean indicating if we *are* expecting a CID.
	Expects() (cid.Cid, bool)
	// Receive a block, advancing the selector.
	Receive(Block) error
	// Skip the block.
	Skip() error
}
```

As we receive messages via graphsync, we'd (in graphsync):

1. For every response in the message.
  a. If the response is "I don't have", call `Skip` on the verifier.
  b. If the response is content, partial content, etc., while `verifier.Expects()` is in the set of blocks in the message, keep feeding those blocks into `verifier.Receive()`.

This shouldn't require any additional information.

cc @hannahhoward, @warpfork, @vmx, @mikeal.

hannahhoward:
As I look at this, it makes sense to me provided we can assume traversing a selector is deterministic, based on the the selector and content being traversed, even if parts of the content are missing.

Essentially, for a selector A, when applied to a DAG of nodes B, the order I will visit nodes to apply selector A will always be the same. This will be true even if I have to skip some nodes in DAG B cause I don't have them (or the remote does not have them).

If this is the case I only envision two components sent over the wire:
1. A block
2. A CID that is part of the traversal that the remote did not have

Of course, the verifier struct @Stebalien implies that Receive gets blocks that don't necessarily match the selector but part of the traversal. (Or are necessary to verify a sub branch does not match the traversal)

So I'd like to make a slightly modified proposal:

Overall call signature:

```go
func VerifySelection(root CID, selector Selector, visit VisitFn) chan Node 
```

returns a channel of verified, matching Nodes

and the VisitFn looking like this:

```go
 type VisitFn func(visitor VerifyVisitor) {
}
```

and the VerifyVisitor {

```go
type VerifyVisitor interface {
  Cid() cid.Cid
  Receive(Node) error
  Skip() error
}
```

Essentially the visit functions job is to call receive or skip, letting the traversal continue based of the selector being applied client-side.

hannahhoward:
^^ @Stebalien  @warpfork  @vmx @mikeal

hannahhoward:
The visitFn would be supplied by go-graphsync and would essentially block until it got a response from the server regarding the given CID of whether it had it or not.

Stebalien:
[...]

PR #91, opened 2019-02-11
Protocol Revision Proposal: Root CID, Partial Response, Block Metadata

hannahhoward:
This proposal makes three changes to the wire protocol to support current thinking on how the first implemented version of the selector interface will work.

1. Add a root CID to the request. Current thinking suggests a root CID is neccesary to begin a request, and absolute selectors will not be supported in the first version of IPLD selector interfaces (at least in go)

2. Add a response code for an in-progress partial response. The responder may want to begin sending response blocks and information before it completes the request, so that it doesn't a huge amount of blocks in memory for a large selection. This is different than a request acknowledged response code (10) in that it contains actual blocks and metadata. It is different that a terminal response with partial data (20), in that it indicates more data is coming-- this is not the last response this responder will send.

3. Add a block metadata structure. This serves two purposes. If a responder is servicing more than one request, it should not send the same block twice for bandwidth reasons. The same is true if the same block appears multiple times in a selection for whatever reason. However, simply sending the blocks provides little information about how they tie to a request, where they sit in the IPLD path relative to the root, or any other information that we may want to communicate to the requester. We may also need to communicate information about a block we don't have -- so that the requester knows that is missing in the response and it can go seek it elsewhere. The metadata block contains:

```
1. A request ID
2. A CID for a block
3. The path for that block in the context of a selection
4. A control message about that block for additional information that can be decoded by IPLD.
```

Importantly, there can be more than one metadata for a given block over the course of the request.

From @warpfork in https://github.com/ipld/specs/issues/90#issuecomment-462029414 -- provides some of the rationale for adding this metadata

```
It's possible that the state machines on both sides of this process could be implemented without extra control messages to mark things like these blocks which we're including despite not matching. I'm uncommitted to whether or not that's a _good_ idea even if it's _possible_ -- this seems like a situation where a little more contextualization information might be extremely useful for development and debugging, sanity checks, meaningful error reports, and future proofing.
```

### DIFF
--- a/graphsync/graphsync.md
+++ b/graphsync/graphsync.md
@@ -68,11 +68,6 @@ message GraphsyncMessage {
     bool  cancel = 5;   // whether this cancels a request
   }
 
-  message RequestList {
-    repeated Request reqs = 1; // a list of requests
-    bool full = 2;			   // whether this is the full RequestList. default to false, for incremental requests.
-  }
-
   message Response {
     int32 id = 1;     // the request id
     int32 status = 2; // a status code.
@@ -85,9 +80,10 @@ message GraphsyncMessage {
   }
 
   // the actual data included in this message
-  RequestList reqlist = 1;
-  repeated Response reslist = 2;
-  repeated Block data = 2;
+  bool completeRequestList    = 1; // This request list includes *all* requests, replacing outstanding requests.
+  repeated Request  requests  = 2; // The list of requests.
+  repeated Response responses = 3; // The list of responses.
+  repeated Block    data      = 4; // Blocks related to the responses
 }
 ```
 
--- a/graphsync/graphsync.md
+++ b/graphsync/graphsync.md
@@ -105,6 +105,7 @@ message GraphsyncMessage {
 11   Additional Peers. PeerIDs in extra.
 12   Not enough vespene gas ($)
 13   Other Protocol - info in extra.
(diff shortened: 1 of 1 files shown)


hannahhoward:
also includes @Stebalien 's changes cause they are likely to make it in.

vmx:
I would *not* send the CID with the request, but with the selector. This way we can have selectors like a multi-selector which requests several CIDs in the future.

I also wouldn't include the path of block. It's more work to implement and I don't see any practical use (rather than a theoretical nice to have).

I personally would also not allow multiple requests/responses, but always only send a single block as this would make the protocol way easier, but I guess there's some people not agreeing with me on trying to make the first version very simple (memcached protocol like).

hannahhoward:
@vmx -- re: CID in selector -- I thought there was an agreement on not implementing absolute selectors in the first version? All of the interfaces I've seen for IPLD selector require a root. Again, if we can support absolute selectors in go-selector and avoid passing roots, then I support that.

vmx:
@hannahhoward You are right, that currently all selectors we need require to be rooted in a CID. My request for making it part of the selector is, that I'd like to add a selector which can request multiple CIDs at the same time (kind of a stripped down multi-select) rather sooner than later.

So it's for keeping this open for future extensions and (IMHO) it even makes things simpler for you.

PR #95, opened 2019-02-18
Specify selector types

vmx:
Add a description of selector types we want to support. There isn't
a nice syntax yet, but there is a schema of the serialised form of
the selectors.

### DIFF
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -346,7 +346,7 @@ type SelectorNonRecursive union {
 } representation keyed
 
 # Paths are split into their individual segments
-type PathSegement String
+type PathSegment String
 
 # Selects a specific full path
 type SelectPath PathSegment
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -340,7 +340,7 @@ type SelectorNonRecursive union {
   | SelectArrayAll "selectArrayAll"
   | SelectArrayPosition "selectArrayPosition"
   | SelectArraySlice "selectArraySlice"
-  | SelectArrayAll "selectMapAll"
+  | SelectMapAll "selectMapAll"
 } representation keyed
 
 # Paths are split into their individual segments
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
 - Allow language-specific implementations of selectors (parsers, execution, etc).
 - (IMPORTANT) Well-designed set of test vectors representing a variety of use cases for IPLD Selectors.
 - A recommended structure for implementing a selector type, with an easy to use test suite.
+
+
+Selector types
+--------------
+
+### CID Selector
+
+Selects a single CID.
+
+
+### Path Selector
+
+Selects a single path segment.
+
+
+### Array Selector
+
+Selects a slice or single item of an array of items.
+
+
+### Recursive Selector
+
+Uses another Selector to select things recursively. It also takes a limit on how often this recursion should happen (the limit can be set to "infinity" to keep traversing until there are no matching paths left.
+
+The recursion may also be stopped by a given CID.
+
+
+### Selector Query
+
+The Selector Query is a list of Selectors. The first element will be a CID Selector which points to the CID your Selector is rooted at.
+
+
+Use cases
+---------
+
+### Deeply nested path
+
+In some DAG you want to get one specific value you know the path of. Let's say you want to get the birth year of a specific character of a specific show.
+
+Example data (as JSON) with CID `cidabcdef`. To keep it simple it's represented as a single JSON structure, but imagine e.g. the characters are linked by CID.
+
+```json
+{
+  "show": "start-trek-voyager",
+  "characters": {
+    "kathrin-janeway": {
+      "birthday": {
+        "day": 20,
+        "month": 5,
+        "year": 2328
+      },
+      "rank": "vice admiral"
+    }
+  }
+}
+```
+
+The final Selector could look like this:
+
+```json
+{
+  "cidRootedSelector": {
+    "root": "cidabcdef",
+    "selectors": [
+      {"selectPath": "characters"},
+      {"selectPath": "kathryn-janeway"},
+      {"selectPath": "birthday"},
+      {"selectPath": "year"}
+    ]
+  }
+}
+```
+
+
+### Getting a certain number of parent blocks in a blockchain
+
+You want to get a certain number of parents from a certain block.
+
+The shape of a block could look like this (in JSON):
+
+```json
+{
+  "parent": "parentcid",
+  "time": 1549641260,
+  "none": 3423545
+}
+```
+
+If you know you want five parents you could use Path Selectors:
+
+```json
+{
+  "cidRootedSelector": {
+    "root": "cidabcdef",
+    "selectors": [
+      {"selectPath": "parent"},
+      {"selectPath": "parent"},
+      {"selectPath": "parent"},
+      {"selectPath": "parent"},
+      {"selectPath": "parent"}
+    ]
+  }
+}
+```
+
+But there's also a better way with using a Recursive Selector with a limit:
+
+```json
+{
+  "cidRootedSelector": {
+    "root": "cidabcdef",
+    "selectors": [
+      {"selectRecursive": {
+        "follow": [
+          {"selectPath": "parent"}
+        ],
+        "depthLimit": 5
+      }}
+    ]
+  }
+}
+```
+
+
+### Getting changes up to a certain one
+
+This use case is inspired by CRDTs, where you have a chain of changes. You observe a new change and want to get all the previous changes up to the one that you have already observed. It is a recursive query with a CID as stopping condition.
+
+The shape of a change could look like this (in JSON):
+
+```json
+{
+  "prev": "prevcid",
+  "timestamp": 1549641260,
+  "value": "abc"
+}
+```
+
+It will be a Recursive Selector following a long a certain field (`prev` in this case):
+
+```json
+{
+  "cidRootedSelector": {
+    "root": "cidabcdef",
+    "selectors": [
+      {"selectCid": "cidabcdef"},
+      {"selectRecursive": {
+        "follow": [
+          {"selectPath": "prev"}
+        ],
+        "cidLimit": "somecid"
+      }}
+    ]
+  }
+}
+```
+
+
+### Getting a full sub-DAG
+
+For getting a full file from UnixFSv1 you need to retrieve a full sub-DAG.
+
+An example selector to get the full sub-DAG rooted at a certain CID:
+
+
+```json
+{
+  "cidRootedSelector": {
+    "root": "cidabcdef",
+    "selectors": [
+      {"selectRecursive": {
+        "follow": [
+          {"selectPath": "Links"},
+          {"selectArrayAll": null},
+          {"selectPath": "multihash"}
+        ]
+      }}
+    ]
+  }
+}
+```
+
+If it's a file in some directory, you can also start at a deeper level:
+
+```json
+{
+  "cidRootedSelector": {
+    "root": "cidabcdef",
+    "selectors": [
+      {"selectCid": "cidabcdef"},
+      {"selectPath": "with"},
+      {"selectPath": "some"},
+      {"selectPath": "subdirectory"},
+      {"selectRecursive": {
+        "follow": [
+          {"selectPath": "Links"},
+          {"selectArrayAll": null},
+          {"selectPath": "multihash"}
+        ]
+      }}
+    ]
+  }
+}
+```
+
+IPLD Schema
+-----------
+
+```ipldsch
+# This is the main entry point for the current selectors
+type CidRootedSelector struct {
+  root Cid
+  # Each element matches a path segement (or several, in some cases)
+  selectors [Selector]
+}
+
+# A catch all for all types of selectors. They are split between recursive and
+# non-recursive ones as a recursive selector can't have another recursive
+# selector embedded
+type Selector union {
+  | SelectPath "selectPath"
+  | SelectArrayAll "selectArrayAll"
+  | SelectArrayPosition "selectArrayPosition"
+  | SelectArraySlice "selectArraySlice"
+  | SelectMapAll "selectMapAll"
+  | SelectRecursive "selectRecursive"
+} representation keyed
+
+# This is a subset of the selectors that can be used within a recursive selector
+type SelectorNonRecursive union {
+  | SelectPath "selectPath"
+  | SelectArrayAll "selectArrayAll"
+  | SelectArrayPosition "selectArrayPosition"
+  | SelectArraySlice "selectArraySlice"
+  | SelectArrayAll "selectMapAll"
+} representation keyed
+
+# Paths are split into their individual segments
+type PathSegement String
+
+# Selects a specific full path
+type SelectPath PathSegment
+
+# Selects all elements of an array, it's similar to a `/*`
+type SelectArrayAll Null
+
+# Selects a specific item from an array
+type ArrayPosition Int
+
+# Selects a slice of items out of an array
+type ArraySlice struct {
+  start optional Int
+  end optional Int
+}
+
+# Selects all keys one level deep, it's similar to a `/*`
+type SelectMapAll Null
+
+# Follow a selector recursively
+type SelectRecursive struct {
+  # Can be used to follow a more complex path, e.g. for UnixFSv1
+  follow [SelectorNonRecursive]
+  # Stop recursing after a certain amount of iterations
+  depthLimit optional Int
+  # Stop recursing once a specific CID is visited
+  cidLimit optional Cid
+}
+```
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -135,9 +135,9 @@ Uses another Selector to select things recursively. It also takes a limit on how
 The recursion may also be stopped by a given CID.
 
 
-### Selector Query
+### CID Rooted Selector
 
-The Selector Query is a list of Selectors. The first element will be a CID Selector which points to the CID your Selector is rooted at.
+The CID Rooted Selector consists of a CID where the selection starts and a list of Selectors specifying the traversal.
 
 
 Use cases
@@ -253,7 +253,6 @@ It will be a Recursive Selector following a long a certain field (`prev` in this
   "cidRootedSelector": {
     "root": "cidabcdef",
     "selectors": [
-      {"selectCid": "cidabcdef"},
       {"selectRecursive": {
         "follow": [
           {"selectPath": "prev"}
@@ -297,7 +296,6 @@ If it's a file in some directory, you can also start at a deeper level:
   "cidRootedSelector": {
     "root": "cidabcdef",
     "selectors": [
-      {"selectCid": "cidabcdef"},
       {"selectPath": "with"},
       {"selectPath": "some"},
       {"selectPath": "subdirectory"},
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -153,7 +153,7 @@ Example data (as JSON) with CID `cidabcdef`. To keep it simple it's represented 
 {
   "show": "start-trek-voyager",
   "characters": {
-    "kathrin-janeway": {
+    "kathryn-janeway": {
       "birthday": {
         "day": 20,
         "month": 5,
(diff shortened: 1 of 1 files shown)


selectors/selectors.md
```diff
@@ -110,295 +110,269 @@ These components imply or expand into the following things:
...
-```
+Selects a single CID.
 
-**TODO**: put selectors into the grand magnificent global universal multicodec table...
 
-### Cid Selector
+### Path Selector
 
-This is a basic selector, that means exactly a single CID, and thus a single graph node. Nothing more.
+Selects a single path segment.
```
momack2, line 123: why rephrase the description to something less descriptive?

momack2: old one was "This is a selector w/ a path. Absolute ones must start with a CID"

selectors/selectors.md
```diff
@@ -110,295 +110,269 @@ These components imply or expand into the following things:
...
-```go
-// Binary: <selector-code><length><path-selector-code><binary-cid><binary-path>
-//     or: <path-selector-code><length><binary-path>
-// String: /sel/path/<multibase-encoded-cid>/a/b/c/d/...
-type PathSelector struct {
-  Path Path
-}
-```
 
-### Glob Selector
```
momack2: why the change from glob to array? AFAICT they don't behave exactly the same - are you really proposing a different selector, or just trying to rename the same one initially described here with regexs for paths?

selectors/selectors.md
```diff
@@ -110,295 +110,269 @@ These components imply or expand into the following things:
...
+### Recursive Selector
 
-This is a selector with a [unix shell glob](https://en.wikipedia.org/wiki/Glob_(programming)) style thing. Absolute ones must start with a CID. It supports the following syntax:
+Uses another Selector to select things recursively. It also takes a limit on how often this recursion should happen (the limit can be set to "infinity" to keep traversing until there are no matching paths left.
```
momack2, line 133: nice description =]

selectors/selectors.md
momack2: how are path and array selectors different from each other? what does a "slice" of items mean vs a "path segment" - could be useful to describe these a little more and explain their usage to help resolve questions like this

selectors/selectors.md
```diff
@@ -110,295 +110,269 @@ These components imply or expand into the following things:
...
-```go
-// Binary: <selector-code><length><glob-selector-code><binary-cid><binary-path>
-//     or: <glob-selector-code><length><binary-path>
-// String: /sel/path/<multibase-encoded-cid>/a/*/**
-type GlobSelector struct {
-    Glob Path
-}
-```
 
-### Multi Selector
```
momack2: I found multi selectors confusing, but if "selector queries" are supposed to be a replacement for them we need to enhance this definition. I'd really like to understand what I'd use selector queries for and how they'd work (do they union/intersect? do I get to choose?)

selectors/selectors.md
```diff
@@ -110,295 +110,269 @@ These components imply or expand into the following things:
...
+        "year": 2328
+      },
+      "rank": "vice admiral"
+    }
+  }
+}
 ```
-<sel-sexp>  ::= <varint-2+><operator>(<selector>)+
+
+The final Selector could look like this:
```
momack2, line 168: I thought selectors were normally described as strings? do you want to include the notation that a client might use to say look up the rank of Katherin?

selectors/selectors.md
```diff
@@ -110,295 +110,269 @@ These components imply or expand into the following things:
...
+      {"selectPath": "birthday"},
+      {"selectPath": "year"}
+    ]
+  }
+}
 ```
 
-Option 2: more complicated at the `sel-sexp` level, but avoids having to embed the multiselector preamble.
 
+### Getting a certain number of parent blocks in a blockchain
```
momack2, line 185: 👍

selectors/selectors.md
```diff
@@ -110,295 +110,269 @@ These components imply or expand into the following things:
...
-    return nil
+IPLD Schema
+-----------
+
+```ipldsch
+# This is the main entry point for the current selectors
+type CidRootedSelector struct {
+  root Cid
+  # Each element matches a path segement (or several, in some cases)
+  selectors [Selector]
```
momack2, line 322: why is this plural?

selectors/selectors.md
```diff
@@ -110,295 +110,269 @@ These components imply or expand into the following things:
...
 
-func (d *traversedDag) GetNode(path Path) Node {
-    if d.s.IncludesPath(path) {
-        return d.d.GetNode(path)
-    }
-    return nil
+# A catch all for all types of selectors. They are split between recursive and
+# non-recursive ones as a recursive selector can't have another recursive
+# selector embedded
+type Selector union {
```
momack2, line 328: you define selector types here, but not the functions the selector exposes (the string/byte representation, the type, etc) - intentional?

momack2 (commented):
Thanks for the writeup, @vmx. I think the selector types could use some more description and have a few questions about title, usage, etc, but appreciate you writing it up!

selectors/selectors.md
vmx: In this spec, there isn't the idea of "one selector with a certain type" anymore. All selectors can be combined. So if you want to access something with a deeply nested path, it is a list of path selector, as a single path selector is one segment of the path only.

selectors/selectors.md
vmx: The functionality is described above. This is a description on how they are encoded within the IPLD data model.

vmx:
@momack2 I've rebased this commit. Sorry for the earlier version it was confusing. It wasn't meant to be diffed with the old version. The overlap with it was just coincidentally. This PR should be read as a new document, not as changes to the old ones. I hope things make more sense then.

Hence I just replied to the questions you had within this new context. If you still have questions, feel free to ask.

selectors/selectors.md
vmx: No, currently there is no simple string encoding, it's always a structure that can e.g. be represented as JSON. So find out her rank would be:

```json
{
  "cidRootedSelector": {
    "root": "cidabcdef",
    "selectors": [
      {"selectPath": "characters"},
      {"selectPath": "kathryn-janeway"},
      {"selectPath": "rank"}
    ]
  }
}
```

Though one future string representation will most likely look like `cidabcdef/characters/kathryn-janeway/rank`.

selectors/selectors.md
warpfork: also, in case it's that the syntax may not yet be familiar: `[Selector]` means an array of things of type `Selector`.  (It's a syntax roughly inspired from GraphQL schemas.)

selectors/selectors.md
warpfork: Has to do with how these are now a composable model.

If you want to start at an absolute CID and then navigate to some path within it, you... combine a CID selector with a Path selector.

So the whole "absolute" thing just kinda went away.  Or rather, It's not that it's gone, it's just that it's not a weird special case of this particular feature now.

warpfork: (There are some nomenclature issues to be sorted out about how we refer to single selector components vs the whole aggregated selector, to be sure...)

warpfork (approved):

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+  | SelectArrayPosition "selectArrayPosition"
+  | SelectArraySlice "selectArraySlice"
+  | SelectArrayAll "selectMapAll"
+} representation keyed
+
+# Paths are split into their individual segments
+type PathSegement String
+
+# Selects a specific full path
+type SelectPath PathSegment
```
hannahhoward, line 350: If this is a full path, should it be `type SelectPath [PathSegment]`?

hannahhoward: actually curious cause the spec above says a path selector is only to select a segment.

hannahhoward (commented):
Minor nits, LGTM

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+
+In some DAG you want to get one specific value you know the path of. Let's say you want to get the birth year of a specific character of a specific show.
+
+Example data (as JSON) with CID `cidabcdef`. To keep it simple it's represented as a single JSON structure, but imagine e.g. the characters are linked by CID.
+
+```json
+{
+  "show": "start-trek-voyager",
+  "characters": {
+    "kathrin-janeway": {
```
rvagg: *nit: `kathryn-janeway` for correctness and to match the selector just below

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+
+
+### Recursive Selector
+
+Uses another Selector to select things recursively. It also takes a limit on how often this recursion should happen (the limit can be set to "infinity" to keep traversing until there are no matching paths left.
+
+The recursion may also be stopped by a given CID.
+
+
+### Selector Query
```
rvagg: is "query" the best name for this? could it just be "Selector List" or is there context I'm missing (very possible!)?

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+```
+
+It will be a Recursive Selector following a long a certain field (`prev` in this case):
+
+```json
+{
+  "cidRootedSelector": {
+    "root": "cidabcdef",
+    "selectors": [
+      {"selectCid": "cidabcdef"},
```
rvagg: is this necessary since it's the root anyway?

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+```
+
+If it's a file in some directory, you can also start at a deeper level:
+
+```json
+{
+  "cidRootedSelector": {
+    "root": "cidabcdef",
+    "selectors": [
+      {"selectCid": "cidabcdef"},
```
rvagg: same question as above, is this serving an important purpose or demonstrating the possible redundancy?

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+  | SelectRecursive "selectRecursive"
+} representation keyed
+
+# This is a subset of the selectors that can be used within a recursive selector
+type SelectorNonRecursive union {
+  | SelectPath "selectPath"
+  | SelectArrayAll "selectArrayAll"
+  | SelectArrayPosition "selectArrayPosition"
+  | SelectArraySlice "selectArraySlice"
+  | SelectArrayAll "selectMapAll"
```
rvagg: `SelectArrayAll` -> `SelectMapAll`

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+} representation keyed
+
+# Paths are split into their individual segments
+type PathSegement String
+
+# Selects a specific full path
+type SelectPath PathSegment
+
+# Selects all elements of an array, it's similar to a `/*`
+type SelectArrayAll Null
```
rvagg, line 353: Couldn't this just be rolled up in to one of the other two array types? Maybe position with a `-1` or slice without bounds specified.

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+type ArrayPosition Int
+
+# Selects a slice of items out of an array
+type ArraySlice struct {
+  start optional Int
+  end optional Int
+}
+
+# Selects all keys one level deep, it's similar to a `/*`
+type SelectMapAll Null
```
rvagg, line 365: There seems to be a linguistic disconnect here between 'path' and 'map', could it be brought in better alignment? `SelectMapAll` -> `SelectPathAll` or `SelectPath` -> `SelectKey`?

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+type PathSegement String
+
+# Selects a specific full path
+type SelectPath PathSegment
+
+# Selects all elements of an array, it's similar to a `/*`
+type SelectArrayAll Null
+
+# Selects a specific item from an array
+type ArrayPosition Int
```
rvagg: `ArrayPosition` -> `SelectArrayPosition`

selectors/selectors.md
```diff
@@ -108,3 +108,271 @@ These components imply or expand into the following things:
...
+
+# Selects a slice of items out of an array
+type ArraySlice struct {
```
rvagg: `ArraySlice` -> `SelectArraySlice`

selectors/selectors.md
rvagg: Further on the map types: this is likely my ignorance speaking, but is it strictly necessary to treat them different to other properties for the purposes of selecting or is it a matter of needing strict typing for values? You could do:

* array positions as paths/keys, just using integers as your selector value
* "all" for both map and array are essentially the same
* slice is distinct to array types

selectors/selectors.md
vmx: Good catch! The comment is wrong. A full path is formed by the `cidRootedSelector` which has an array of selectors, hence can have an array of `SelectPath`, hence a full path.

selectors/selectors.md
vmx: It is not. Thanks for catching, it must be a leftover from an older version.

selectors/selectors.md
vmx: Having `SelectArrayAll` is more explicit (which I like), it also plays nice with type systems.

Have a separate type for "Map" helps for better error checking/messages. You should normally know if it's an array or a map. Though it might make sense to add a "I don't know/care" in the future if it turns out it makes sense.

selectors/selectors.md
vmx: Thanks for bringing this up. I see this as a broader issue (even IPLD wide), hence I've opened https://github.com/ipld/specs/issues/103 and will leave it like that for now, so we can get it merged.

selectors/selectors.md
vmx: Good catch, this is now the `CidRootedSelector`. That also explains why the examples below were wrong :)

vmx:
Thanks @hannahhoward and @rvagg all issues should be addressed. Please give your thumbs up/approve this PR so that we can merge it soon.

selectors/selectors.md
mikeal: best captain, no contest.

don’t @ me :P

rvagg (approved):

hannahhoward (approved):

whyrusleeping:
Without getting into the details of the schema for encoding selectors, or their human readable representations, the expressibility of this scheme seems good enough. @jbenet is gonna be upset that selectors arent always rooted in a CID, but I think separating them makes the composition story cleaner. That said, do we have a term that describes 'selector rooted with a CID' that isnt that long? I think the bulk of juans argument for having selectors always be CID rooted is that it provides an easy term to talk about that particular thing.

mikeal:
I vote for “rooted selector” and “un-rooted selector” :)

mikeal:
I’ll also quickly note that having them separated makes for a very nice and more advanced pinning system around selectors, where you create a pin for an “un-rooted selector” and then you simply set and re-set the “root” of the selector.

hannahhoward:
@vmx @momack2 can we go ahead and merge this? who else needs to sign off? (It would be nice to be able to reference the updated file in some stuff that I am doing)

selectors/selectors.md
```diff
@@ -108,3 +108,269 @@ These components imply or expand into the following things:
...
+
+
+### Recursive Selector
+
+Uses another Selector to select things recursively. It also takes a limit on how often this recursion should happen (the limit can be set to "infinity" to keep traversing until there are no matching paths left.
+
+The recursion may also be stopped by a given CID.
+
+
+### CID Rooted Selector
```
Stebalien, line 138: Hm. Shouldn't we have two?

1. A CID selector (which we have).
2. A sequential selector that selects each selector one after the other?

selectors/selectors.md
```diff
@@ -108,3 +108,269 @@ These components imply or expand into the following things:
...
+
+
+### Path Selector
+
+Selects a single path segment.
+
+
+### Array Selector
+
+Selects a slice or single item of an array of items.
```
Stebalien, line 128: Wouldn't the "single item" selector just be the path selector? That is, I assume the array selector will always return an array.

selectors/selectors.md
Stebalien: Alternatively, we could drop the CID selector and have:

1. A CID rooted selector. `{"root": Cid, selector: ...}`. Use with no selector (or the identity selector?) to select the root cid.
2. A sequential selector.

selectors/selectors.md
```diff
@@ -108,3 +108,269 @@ These components imply or expand into the following things:
...
+
+
+### Recursive Selector
+
+Uses another Selector to select things recursively. It also takes a limit on how often this recursion should happen (the limit can be set to "infinity" to keep traversing until there are no matching paths left.
+
+The recursion may also be stopped by a given CID.
```
Stebalien, line 135: What if we just had a general purpose "stop" condition selector?

selectors/selectors.md
Stebalien: Actually, isn't this more of a "range" selector? That is, we're selecting a "range" of keys (from an object or an array).

selectors/selectors.md
```diff
@@ -108,3 +108,269 @@ These components imply or expand into the following things:
...
+
+Selects a single path segment.
+
+
+### Array Selector
+
+Selects a slice or single item of an array of items.
+
+
+### Recursive Selector
```
Stebalien, line 131: Let's say that I want to lookup a specific path on each block along a chain. I think we'll need a way to fork. Maybe something like:

```
{
  "recursive": {
    "follow": ...,
    "select": ...,
    "stop": ...,
  }
}
```

Stebalien (commented):
We should look at other graph query systems (and SQL). We don't have to invent this _entirely_ ourselves and we're definitely not going to make the best thing we can if we try.

selectors/selectors.md
vmx: I introduced the CID Rooted Selector so that we can have other ones, like a sequential one (i'd call it multi selector) in the future. It can be added to the spec later on. I just wanted to keep things as minimal as possible to make agreement easier.

vmx: You can use `{"root": CID, selector: []}` to select only a CID.

selectors/selectors.md
vmx: That's actually a good point. I named it after the data structures it operates on. But it makes sense to name them after the result (what they actually return). Is it OK if I create a PR once this one is merged (so that we can then discuss it there if anyone has objections, instead of postponing this one even more).

selectors/selectors.md
vmx: I'll open an issue for that once merged.

selectors/selectors.md
vmx: In the current spec we don't have any filtering capability at all yet. Currently we just return everythig we traverse. I'll open an issue for that once this spec is merged.

vmx:
I merge this spec now, I think we found agreement on the general direction. @Stebalien has pointed out several things which I'll address in subsequent PRs/issues.

selectors/selectors.md
vmx: Actually this part is outdated. We don't have a general "Array Selector" anymore. We have a single item selector, slice and all elements selector now. I'll update the spec.

vmx: https://github.com/ipld/specs/pull/104

selectors/selectors.md
vmx: https://github.com/ipld/specs/issues/105

selectors/selectors.md
vmx: https://github.com/ipld/specs/issues/106

selectors/selectors.md
vmx: About the sequential selector: https://github.com/ipld/specs/issues/107

Stebalien: My point was to keep each selector as minimal as possible. The CID rooted selector does two things: selects a CID and a sequence of follow-on selectors.

vmx: Yes. Currently it's the only "entry point". So the CID Rooted Selector is kind of the only proper Selector, all other ones are "Selector Components" or something like that.

Issue #96, opened 2019-02-19
Clarifying what Graphsync means to @jbenet and @whyrusleeping


vmx:
I thought I finally have a good understanding of what Graphsync means to @jbenet and @whyrusleeping. But I don't. Hence I open this issue to see if we can clarify things for me once more. Below is my current view.

### The vision

Graphsync is about synchronising remote DAGs. It's a way to make sure you have the data you need locally available. In order to get there, you send requests to other peers which will then reply with the missing bits. Once fully implemented, it works across a peer-to-peer network.

### The short-cut for now

In order to get there, we've simplified things down to:

 - Instead of working across a full peer-to-peer network, the networking is restricted to two peers talking to each other.
 - Instead of being able to request very specialised sub-sets of the DAG we restrict it to simple, broader selectors.
 - Instead of being able to upgrade/downgrade/change an already running request, you can only cancel a request.

Thought reverse, we want to build the non-simplified thing in the future.

### What we are building now

 - A wire protocol that supports the full system, but we only leverage/implement the parts needed for the short-cut.
 - Simpler selectors that are useful for the use-cases we have right now.
 - No state on the server, once a traversal is started, it either finishes, or gets cancelled. It's never changed while running.

### What we are not building

 - A smaller system that is useful by its own, before the full system is ready.
 - A wire protocol that serves the short-cut version better and could be extended later on (e.g. to optimise things).
 - A system that can be used as building block for larger ones (e.g. a system that could be used to implement Graphsync proposal (a/c)).

whyrusleeping:
graphsync is basically just an upgraded bitswap, with the ability to ask for selectors instead of just single hashes. At the same time, we want to include a few improvements to the protocol, to allow for extensions and better logic to be written around it (the 'I dont have that' error message in particular would have been really helpful for bitswap).

Ideally, graphsync is a drop-in replacement for bitswap, and could be used everywhere bitswap is used. In addition, it would be able to ask for more complicated things, if the caller requests it.

vmx:
Thanks @whyrusleeping for the reply. Although I heard Graphsync being "a better Bitswap" floating around, but articulated so clearly is news to me. I find the current proposal reaching for a much bigger thing that hasn't much in common with Bitswap. Though it might be my lack of a deep understanding of Bitswap.

vmx:
[...]

PR #97, opened 2019-02-19
Proposed changes for initial go implementation: Partial response code & @stebalian changes

hannahhoward:
This proposal covers all proposed changes required for the initial go implementation.

It includes:

1. @Stebalien 's changes, because they make sense, outlined here: https://github.com/ipld/specs/pull/89

2. One additional response code, 14, called a partial response. Here is the rationale:

The original proposal includes an acknowledgement of the request, and two terminal response success codes - one for a completed request with all content, one for a completed request with only partial content, because the server was unable to satisfy all of the potential paths covered by the requested selector.

However, in order to send all of the found blocks in a terminal response code, the server would buffer all of them in memory as it traverses the graph before sending them. Similarly, in order to verify these blocks actually match the selector requested, the client would need to keep all of them in memory as the response is processed.

The proposal for the initial go implementation, in order to optimize memory usage, is to instead have the server send blocks as soon as it finds them, as well as information about what blocks are not available as soon as it knows. Then the client will verify and return blocks to the graphsync caller as soon as it receives them.

To support this, I'm proposing adding a "partial response" code. When this response code is transmitted, the "extra" field of the response becomes metadata about what has been transmitted -- either CIDs of blocks that are included in the response, or CIDs the server knows it doesn't have, or other control messages. For the moment, the content and structure of this extra data is left intentionally ambiguous, because more potential use cases may emerge during the actual implementation of go-graphsync and the IPLD selector library. It's structure will likely be specified more concretely once an implementation has been completed.

---

This proposal does not address the larger potential issues with the protocol outlined in https://github.com/ipld/specs/issues/96

### DIFF
--- a/graphsync/graphsync.md
+++ b/graphsync/graphsync.md
@@ -95,6 +95,7 @@ message GraphsyncMessage {
 11   Additional Peers. PeerIDs in extra.
 12   Not enough vespene gas ($)
 13   Other Protocol - info in extra.
+14   Partial Response w/ metadata, may include blocks, metadata in extra
 
 # success - terminal
 20   Request Completed, full content.
--- a/graphsync/graphsync.md
+++ b/graphsync/graphsync.md
@@ -68,11 +68,6 @@ message GraphsyncMessage {
     bool  cancel = 5;   // whether this cancels a request
   }
 
-  message RequestList {
-    repeated Request reqs = 1; // a list of requests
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
LGTM but probably needs a quick signoff from @jbenet and/or @whyrusleeping (although I don't expect any objections).

whyrusleeping (approved):
seems fine to me

PR #104, opened 2019-03-19
Update Selector types overview

vmx:
The selector types section was outdated and didn't describe the Selectors
that are defined in the schema.

### DIFF
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -113,19 +113,23 @@ These components imply or expand into the following things:
 Selector types
 --------------
 
-### CID Selector
-
-Selects a single CID.
-
-
 ### Path Selector
 
 Selects a single path segment.
 
 
-### Array Selector
+### Array Selectors
 
-Selects a slice or single item of an array of items.
+There's three selectors that operate on arrays:
+
+ - Array All Selector: selects all items of an array
+ - Array Slice Selector: selects a given range of an array
+ - Array Position Selector: selects a single element from an array
+
+
+### Map All Selector
+
+Selects all elements of a map structure.
 
 
 ### Recursive Selector


selectors/selectors.md
```diff
@@ -113,19 +113,23 @@ These components imply or expand into the following things:
...
 
-### Path Selector
+### Array Selectors
 
-Selects a single path segment.
+There's three selectors that operate on arrays:
+
+ - Array All Selector: selects all items of an array
+ - Array Slice Selector: selects a given range of an array
+ - Array Position Selector: selects a single element from an array
```
Stebalien, line 127: 1. Isn't the array "all" selector just a "glob" selector? How is this different from the map all selector?
2. Isn't the array position selector just a path selector?

If we're trying to do type checking _and_ path selection, I think this is the wrong place to do so. This is one of the cases where we need some kind of "where" selector: `select X where typeof(X) == array`.

vmx: > If we're trying to do type checking _and_ path selection, I think this is the wrong place to do so.

Yes, this was the intention. I would expect that the one writing the query knows whether it's a map or an array and would rather prefer getting an error message than weird results.

Stebalien: I don't really see how the results will be weird. We're basically just selecting different paths through the graph. If we select a subset of an array, we get a subset of the array.

vmx: And if it isn't an array, you would get nothing back instead of an error.

Well, perhaps the implementations could be smart enough that if you want to do a subset and it's not an array it will give you an error. But then again, do you always want an error in that case or just skip it?

rvagg:
superseded by more recent work that's been landed

Issue #67, opened 2018-07-17
Spec refining: specify valid merklepath segments and encoding


warpfork:
## Mission

It's important to specify precisely what is a valid merklepath segement in IPLD.

The spec currently contains a "TODO: list path resolving restrictions" and this could be improved :)

---

## Why

First, a quick clarification: "merklepath segments" are a distinct concept from "IPLD Selectors".  Merklepaths are a specific and limited implementation of IPLD Selectors; they can only specify a traversal to a single object; and importantly, we want them to be serializable in a way that's easy for humans to operate.  To quote the current spec for motivations:

> IPLD paths MUST layer cleanly over UNIX and The Web (use `/`, have deterministic transforms for ASCII systems).

(Perhaps "ASCII" is a little over-constrained there.  The spec *also* says "
IPLD paths MUST be universal and avoid oppressing non-english societies (e.g. use UTF-8, not ASCII)" -- we might want to refine those two lines after we tackle the rest of this issue.)

Second of all, just a *list* of other issues that are fairly closely related to a need for clarity on this subject:

 - https://github.com/ipfs/go-ipfs/issues/1710 -- "**IPFS permits undesirable paths**" -- lots of good discussion on encodings, and examples of valid but problematic characters
- https://github.com/ipld/specs/issues/59 -- "**Document restrictions on keys of maps**"
- https://github.com/ipld/specs/issues/58 -- "**Non-string keys**" -- this one has a particularly interesting detail quoted: "The original intention was to actually be quite restrictive: map keys must be unicode strings with no slashes. However, we've loosened that so that they can contain slashes, it's just that those keys can't be pathed through.".  (n.b. *this* issue here is named "merklepath segments", not "IPLD keys", specifically to note this same distinction.)
- https://github.com/ipld/specs/issues/55 -- "**Spec out DagPB path resolution**"
- https://github.com/ipld/specs/issues/37 -- "**Spec refining: make sure that an attribute cannot be named `.`**"
- https://github.com/ipld/specs/issues/20 -- "**Spec refining: Merkle-paths to an array's index**"
- https://github.com/ipld/specs/issues/15 -- "**Spec refining: Terminology IPLD vs Merkle**" -- basically, am I titling this issue correctly by saying "merklepath"?  Perhaps not ;)
- https://github.com/ipld/specs/issues/1 -- "**Spec refining: Relative paths in IPLD**" -- may require reserving more character sequences as special
[...]

PR #112, opened 2019-04-12
Data Model doc expanded, and in terms of Kinds.

warpfork:
This is part of a motion towards being able to introduce additional
new terminology as a part of work on Schemas (which will follow in
(many) additional separate commits), and is also alluded to in the
content already.

(There's also a dangling link to a 'schema' folder that doesn't exist
yet, but will quite presently.  Mea cupla; gotta pick *some* order to
get started in!)

### DIFF
--- a/IPLD-Data-Model-v1.md
+++ b/IPLD-Data-Model-v1.md
@@ -50,28 +50,32 @@ The _scalar kinds_ (the complement of recursive) are:
 these Data Model kinds can be called the "_representation kinds_",
 while the additional kinds introduced in the Schema layer are "_perceived kinds_".)
 
+### Kinds Reference
 
-### Null kind
+Each of the following sections provides details about each of the kinds
+introduced in the summary table above.
+
+#### Null kind
 
 Null is a scalar kind.  Its cardinality is one -- the only value is 'null'.
 
-### Boolean kind
+#### Boolean kind
 
 Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or the value 'false'.
 
-### Integer kind
+#### Integer kind
 
 
-### Float kind
+#### Float kind
 
 
-### String kind
+#### String kind
 
 String is a scalar kind.  Its cardinality is infinite -- strings do not have a length limit.
 
 Strings should be UTF-8 text, in [NFC](https://www.unicode.org/reports/tr15/#Norm_Forms) canonicalization.
 
-### Bytes kind
+#### Bytes kind
 
 Bytes is a scalar kind.  Its cardinality is infinite -- byte sequences do not have a length limit.
 
@@ -79,28 +83,26 @@ Bytes are distinct from strings in that they are not considered to have any char
 generally expected to be printable as human-readable text.
 In order to print byte sequences as text, additional effort such as Base64 encoding may be required.
 
-### List kind
+#### List kind
 
 List is a recursive kind.
 
 Values contained in lists can be accessed by their ordinal offset in the list.
 
-### Map kind
+#### Map kind
 
 Map is a recursive kind.
 
 Values in maps are accessed by their "key".  Maps can also be iterated over,
 yielding key+value pairs.
 
-### Link kind
+#### Link kind
 
 A link represents a link to another IPLD Block. The link reference
 is a [`CID`](CID.md).
 
 Link is a scalar kind -- however, when "loaded", may become another kind, either scalar or recursive!
 
+### Kinds Implementation References
 
-Implementation References
--------------------------
-
-- in Go: https://github.com/ipld/go-ipld-prime/blob/master/kind.go
+- Kinds in Go: https://github.com/ipld/go-ipld-prime/blob/master/kind.go
--- a/IPLD-Data-Model-v1.md
+++ b/IPLD-Data-Model-v1.md
@@ -69,7 +69,7 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or 
 
 String is a scalar kind.  Its cardinality is infinite -- strings do not have a length limit.
 
-Strings are generally understood to be UTF-8 text, in [NFC](https://www.unicode.org/reports/tr15/#Norm_Forms) canonicalization.
+Strings should be UTF-8 text, in [NFC](https://www.unicode.org/reports/tr15/#Norm_Forms) canonicalization.
 
 ### Bytes kind
 
--- a/IPLD-Data-Model-v1.md
+++ b/IPLD-Data-Model-v1.md
@@ -1,17 +1,103 @@
-# [WIP] IPLD Data Model
+IPLD Data Model
+===============
 
-## Simple Types
+The IPLD Data Model is "layer 1" of the IPLD core models.
+("Layer 0" are the [Codecs](Codecs); "layer 2" is called [Schemas](schema).)
 
-* Boolean
+The Data Model specifies what categories of data are representable in IPLD --
+for example, booleans, and integers, and textual strings.
+[Codecs](Codecs) specify exactly how these data are transcribed into serialized bytes.
+
+
+Kinds
+-----
+
+The following is the list of essential _kinds_ (or more formally, _representation kinds_)
+of data representable in the IPLD Data Model:
+
 * Null
-* String
+* Boolean
 * Integer
 * Float
-* Array
-* Object (Hash Map)
-* Binary
+* String
+* Bytes
+* List
+* Map
+* Link
 
-## Link Type
+(Note that we use the term "_kinds_" here to disambiguate this from "_types_",
+which is a term we'll use at the [Schemas](schema) level.)
 
-This type represents a link to another IPLD Block. The link reference
+The _recursive kinds_ are:
+
+* List
+* Map
+
+The _scalar kinds_ (the complement of recursive) are:
+
+* Null
+* Boolean
+* Integer
+* Float
+* String
+* Bytes
+* Link
+
+(Note that [Schemas](schema) introduce a few more kinds -- when clarification is necessary,
+these Data Model kinds can be called the "_representation kinds_",
+while the additional kinds introduced in the Schema layer are "_perceived kinds_".)
+
+
+### Null kind
+
+Null is a scalar kind.  Its cardinality is one -- the only value is 'null'.
+
+### Boolean kind
+
+Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or the value 'false'.
+
+### Integer kind
+
+
+### Float kind
+
+
+### String kind
+
+String is a scalar kind.  Its cardinality is infinite -- strings do not have a length limit.
+
+Strings are generally understood to be UTF-8 text, in [NFC](https://www.unicode.org/reports/tr15/#Norm_Forms) canonicalization.
+
+### Bytes kind
+
+Bytes is a scalar kind.  Its cardinality is infinite -- byte sequences do not have a length limit.
+
+Bytes are distinct from strings in that they are not considered to have any character encoding nor
+generally expected to be printable as human-readable text.
+In order to print byte sequences as text, addition effort such as Base64 encoding may be required.
+
+### List kind
+
+List is a recursive kind.
+
+Values contained in lists can be accessed by their ordinal offset in the list.
+
+### Map kind
+
+Map is a recursive kind.
+
+Values in maps are accessed by their "key".  Maps can also be iterated over,
+yielding key+value pairs.
+
+### Link kind
+
+A link represents a link to another IPLD Block. The link reference
 is a [`CID`](CID.md).
+
+Link is a scalar kind -- however, when "loaded", may become another kind, either scalar or recursive!
+
+
+Implementation References
+-------------------------
+
+- in Go: https://github.com/ipld/go-ipld-prime/blob/master/kind.go
--- a/IPLD-Data-Model-v1.md
+++ b/IPLD-Data-Model-v1.md
@@ -1,12 +1,15 @@
 IPLD Data Model
 ===============
 
-The IPLD Data Model is "layer 1" of the IPLD core models.
-("Layer 0" are the [Codecs](Codecs); "layer 2" is called [Schemas](schema).)
+The IPLD Data Model is a core part of the IPLD specification,
+which describes what data is representable in IPLD --
+for example, booleans, integers, textual strings, maps and lists, etc.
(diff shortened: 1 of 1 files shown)


IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,103 @@
...
+Bytes are distinct from strings in that they are not considered to have any character encoding nor
+generally expected to be printable as human-readable text.
+In order to print byte sequences as text, addition effort such as Base64 encoding may be required.
+
+### List kind
+
+List is a recursive kind.
+
+Values contained in lists can be accessed by their ordinal offset in the list.
+
```
warpfork, line 91: This area could probably be improved by lifting @rvagg's "Singular" vs "Associative" terminology from https://github.com/ipld/specs/pull/110/files...

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,103 @@
...
+
+### Float kind
+
+
+### String kind
+
+String is a scalar kind.  Its cardinality is infinite -- strings do not have a length limit.
+
+Strings are generally understood to be UTF-8 text, in [NFC](https://www.unicode.org/reports/tr15/#Norm_Forms) canonicalization.
```
warpfork: n.b. I'm stating this one as a fact, because afaik, the last time string normalization was discussed was in https://github.com/ipld/specs/issues/67, and there seems to be consensus that NFC is the most widely accepted answer.

I also find the Golang approach of "strings are just bytes, actually; glhf" to be quite practical, which is why I used the somewhat loose phrasing of "generally understood".  We might want to lock things down even more explicitly when it comes to map keys (for pathability reasons), but that's a bigger issue than we probably want to engage within the scope of this PR.

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,103 @@
-# [WIP] IPLD Data Model
+IPLD Data Model
+===============
 
-## Simple Types
+The IPLD Data Model is "layer 1" of the IPLD core models.
```
mikeal: I've been thinking of Layer 0 as the Block Layer, then Codecs, then Data Model, then Schema.

I think it's important to have a Block layer separate from codecs because a lot of things operate purely on CID's and never actually decode the data.

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,103 @@
...
+* Boolean
+* Integer
+* Float
 * String
+* Bytes
+* List
+* Map
+* Link
+
+(Note that we use the term "_kinds_" here to disambiguate this from "_types_",
```
mikeal, line 31: For some reason I instinctively consider "type" lower level than "kind", maybe that's just me tho. I totally agree we need a way to distinguish between these two things, I think I've said "primitive types" and "base types" before and those terms don't distinguish between them enough.

IPLD-Data-Model-v1.md
mikeal: I would change "generally considered" to the RFC "MUST" term. I don't think we can leave any ambiguity here.

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,103 @@
...
+Values in maps are accessed by their "key".  Maps can also be iterated over,
+yielding key+value pairs.
+
+### Link kind
+
+A link represents a link to another IPLD Block. The link reference
 is a [`CID`](CID.md).
+
+Link is a scalar kind -- however, when "loaded", may become another kind, either scalar or recursive!
+
```
mikeal, line 105: We should put a note in here about identity multi-hashes just so that people are aware that there are cases where the CID isn't necessarily a "remote" reference.

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,103 @@
...
+
+Implementation References
+-------------------------
+
+- in Go: https://github.com/ipld/go-ipld-prime/blob/master/kind.go
```
mikeal: Should probably also link to all the Codecs that support the full data model.

IPLD-Data-Model-v1.md
warpfork: I'd like to think air-quotes'd "loaded" is already sufficiently mystic about that.  The statement that it may become another kind is accurate, at least.

Does the CID page already handle mentioning identity multihash?  That seems like a more appropriate place to put that detail.

IPLD-Data-Model-v1.md
warpfork: Oh.  Uh, yeah, I should reduce the heading power level... this belonged under the "kinds" heading.  Good catch.

IPLD-Data-Model-v1.md
mikeal: Ya, I guest the most important place to note this will be wherever we describe paths and selector traversal because that’s where it mostly comes into play.

IPLD-Data-Model-v1.md
warpfork: I'll go for "should".  In lowercase, but correct to interpret in RFC 2119 terms.

I've noticed some other modern specs, even when using RFC 2119 terms (and explicitly denoting it in their preambles) are dropping the capslock on the terms.  If we do use RFC 2119, I'd vote for doing the same.  The capslock adds a lot of "line noise", IMO; and I've been reading the specs that drop the caps without any confusion.

We might upgrade this to "must" at some point, but I'd rather give that it's own discussion space.  Ideally some concrete ideas about programmatically how we'd enforce that, some intuitions about the performance impact of it, and most importantly how we'd recommend raising the errors would all be good to have in hand before going into that.

IPLD-Data-Model-v1.md
warpfork: Okay, I'm just going to reword this to drop the numerical references for now then.  I think all of these sentences can be constructed, and possibly better, without them.

IPLD-Data-Model-v1.md
warpfork: It's possible we should split this out into a couple of files: make a `data-model/` directory, and put a lot of this new content I've added into a specific file named `kinds.md` (or `representation-kinds.md`?).

IPLD-Data-Model-v1.md
rvagg: I agree, _but_, if the schema work goes well then we're going to be talking about their "types" more than these representation "kinds", so which term would you rather be using more often?

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,108 @@
...
+* Map
+* Link
+
+(Note that we use the term "_kinds_" here to disambiguate this from "_types_",
+which is a term we'll use at the [Schemas](schema) level.)
+
+The _recursive kinds_ are:
+
+* List
+* Map
```
rvagg, line 37: :+1: to these renamings

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,108 @@
...
-* Array
-* Object (Hash Map)
-* Binary
+* String
+* Bytes
+* Link
+
+(Note that [Schemas](schema) introduce a few more kinds -- when clarification is necessary,
+these Data Model kinds can be called the "_representation kinds_",
+while the additional kinds introduced in the Schema layer are "_perceived kinds_".)
```
rvagg, line 51: wait, aren't those "types"?

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,108 @@
...
+
+Null is a scalar kind.  Its cardinality is one -- the only value is 'null'.
 
-## Link Type
+#### Boolean kind
 
-This type represents a link to another IPLD Block. The link reference
+Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or the value 'false'.
+
+#### Integer kind
```
rvagg, line 66: Should we be providing bounds (or safe bounds at least) for integers and floats? I can imagine extensions to the schema that introduce "big" number types that are different ways of encoding with the base kinds, strings, byte arrays, tuples of integers, etc.

IPLD-Data-Model-v1.md
rvagg: There's something a bit off here and in Map about using the description of _what you can do_ with them rather than _what they are_. Since a List and Map can hold arbitrary other kinds, do we have to build up the concept from first principles rather than just saying "a List is like an array" and "a Map is like an associative array"?

IPLD-Data-Model-v1.md
```diff
@@ -1,17 +1,108 @@
...
+
+List is a recursive kind.
+
+Values contained in lists can be accessed by their ordinal offset in the list.
+
+#### Map kind
+
+Map is a recursive kind.
+
+Values in maps are accessed by their "key".  Maps can also be iterated over,
```
rvagg, line 96: do we have a restriction on the kind(s) that can be used for Map keys? just strings? strings and integers? arbitrary?

rvagg:
_finally_, I keep on being tempted to have a go at this myself but then realise my mental model isn't quite complete enough yet

IPLD-Data-Model-v1.md
warpfork: each type is of some kind -- which may either be a representation kind or a perceived kind.

more on that in the other PR, but for example: a type may be of kind 'struct', which is a perceived kind (in that it doesn't exist at the Data Model layer) (and also needs to have a specified representation, which for a struct will probably be 'map' for example).

IPLD-Data-Model-v1.md
warpfork: Probably but that's a super huge issue, both of nuclear-reaction-design complexity and strangely popularized enough that it still can attract bikeshed-painting levels of casual interjections -- I'm utterly terrified of getting hung up on that and don't wanna talk about it in this PR :)

IPLD-Data-Model-v1.md
warpfork: Casually, there seems to be a consensus we're going to go with strings.  The ecosystemic costs of having non-printable map keys (e.g. bytes) seem to be... High.

Integer map keys can sometimes be useful, but anecdotally, the most common response on hearing they're disallowed is "harumph.  okay." rather than any more ardent objections; while on the costs side, integer map keys are fraught with peril and nonsense in practice in languages like javascript, and they're possible but add nontrivial implementation complexity in go...  There'd be benefits to supporting integer keys, no doubt, but the cost/benefit ratio seems rough.

On a bright side, one cool thing we can do with the schema-vs-representation-kinds relationship: supposing we spec that map keys must be of _representation kind_ string... we can in fact still allow complex keys, even structs, as map keys.  They just need a representation strategy like e.g. 'stringjoin'.

Stebalien: Personally, I'd _kind_ of like to have bytes, using multibase in paths. I believe this should "just work" and should improve some datastructures quite a bit.

(where "some datastructures" include, e.g., HAMTs)

mikeal: Most formats don’t support non-string keys so it would be quite the ask to add this to the data model layer. 

Adding it to the Map type over in schema land is more doable, but I’d still like to have a base Map type that only supports strings and we can define some fancier Map type with support for bytes as keys.

rvagg: Sounds fine. Base Map key type is strings only. Collections built on top can optionally have more sophisticated types.

So far in my collections experiments [I've just been targeting byte arrays](https://github.com/rvagg/iamap#IAMap_set) (or things that can be easily converted to them) because they are the easiest thing to hash and give you maximal flexibility. I can imagine some folks wanting specific types or cleaner interfaces for other types (i.e. without having to jump through the convert-to-byte-array hoops) and I think I've opened the door for that possibility in https://github.com/ipld/specs/pull/110, but it'll likely be an exercise for the people who want that.

rvagg:
We have a bit of a problem now, with the renaming of "Array" and "Object (Hash Map)" here to "Set" and "Map" there is a terminology conflict with the collections work @ https://github.com/ipld/specs/pull/110

To illustrate:

Screenshot 2019-04-16 16 41 33

I think this new naming is correct, we're just not differentiating between the layers. How do we make it clear to the casual reader that these things are different types building blocks?

* Call this the "block encoding data model"? (does that solve much?)
* Refer to the collection types as "multi-block utility collections" or "block-spanning utility collections"? (awkward?)
* Rename the collection types to be more specific: SpanningMap, SpanningSet, SpanningList? (awkward and would need to be done by everyone that implements something at this layer, probably not realistic)
* Just add clarification to both documents that we're talking about different things?

rvagg:
Sorry, "List" not "Set", but the point still stands and I could have listed "Maps, Lists, Sets, etc." in that stack graphic, Lists and Maps are (or will be) in both.

IPLD-Data-Model-v1.md
mikeal: That’s a good point. Type is the more well known term so I’d prefer to use that one more often.

IPLD-Data-Model-v1.md
mikeal: I’m fine leaving it ambiguous for now and maybe going back and scaling it *down* to least common denominator once we have a known way in Schema land to do improved number types.

warpfork:
> Refer to the collection types as "multi-block utility collections" or "block-spanning utility collections"?

I think I like this approach.  Alternatively, I've been calling things "advanced data layouts" or "AdvLayout" for short in some other draft documents.  Same thing, though.  I like your words for it too.

The multi-block collections should fit the same programmatic interfaces as the data model primitives of the same kind, so that will also need some clarification documents.

vmx (approved):

rvagg:
> The multi-block collections should fit the same programmatic interfaces as the data model primitives

I'm not convinced of this point, but we can defer discussion of that till later unless you want to drop into #110 and argue the point.

Stebalien:
I'm not seeing any definition of the term "kind". I assume it's kind of like a class and types are more concrete? Or are types composite? If that's the case, I'd just call these types.

mikeal:
@Stebalien I had a similar initial response but came around to using *kind* here and reserving *type* for schemas https://github.com/ipld/specs/pull/112/files#r274972214

We should document `kind` and `type` in the Terminology section of the readme.

warpfork:
I *think* all of the current comments that need immediate fixes are addressed -- and the rest of this points to more docs we'd like to write, but that can come in more PRs.  Gonna merge this, and we can keep iterating from here!

PR #113, opened 2019-04-14
Begin schema specs

warpfork:
Beginning to write down schema specs, both examples and prose.

Getting this to a satisfactory state of documentation will be a long process -- there are TODOs and blanks in here 'o plenty.  I'm pushing the first commit of this so we can start referring to it in other exploratory hacks, but it's very WIP and it'll probably be many more commits before we want to merge this.  It's a start, though.

There are several sources of existing docs which haven't been fully included, but are other good references and should probably mostly be hoisted over here:
- various docs in the [go-ipld-prime docs dir](https://github.com/ipld/go-ipld-prime/tree/4169feeb45509ce2032278c2e633f8211c6e56f1/doc)
- a large example of [the self-describing schema schema](https://github.com/ipld/go-ipld-prime/blob/4169feeb45509ce2032278c2e633f8211c6e56f1/typed/declaration/schema-schema.ipldsch) also from the go-ipld-prime repo
- the original PR landing the schema-schema -- https://github.com/ipld/go-ipld-prime/pull/1 -- has some pretty rich commentary, which I'm not sure has yet been entirely recaptured
- a somewhat [old gist](https://gist.github.com/warpfork/6df17e791936d1f9b0d5e5483678c8bf), which though now syntactically heckin' legacy already, does have some decent comments about the earlier thoughts about kinds
- some discussions on union representations in unixfsv2: https://github.com/ipfs/unixfs-v2/issues/20 which are utterly generalizable
- and similar, earlier design discussion on union representation concepts over in refmt: https://github.com/polydawn/refmt/pull/30

There are also allusions here to the Data Model and representation kinds which are currently somewhat dangling, but will make more sense after https://github.com/ipld/specs/pull/112 is merged (I didn't do PRs-on-PRs; I've a dim view of the manageability of those).


schema/representations.md
```diff
@@ -0,0 +1,256 @@
+
+Kinds and their Representations
+-------------------------------
+
+A type at the Schema level must be mapped onto a representation expressible
+within the Data Model.
+
+For all Schema kinds which have equivalents at the Data Model level, there is
+a default representation.  The default representation this is, broadly speaking,
```
rvagg: remove "this"

schema/representations.md
```diff
@@ -0,0 +1,256 @@
...
+
+(Kinds for which no meaningful customization is possible -- Null, Boolean,
+Integer, Float, String, Bytes, and List -- are elided.
+N.B., it's not ruled out that custom representations may be introduced for
+these kinds in the future; there are simply none to date.
+Link is also not described in this section, as it's a rather unique business.)
+
+- Map
+	- `map` representation -- the default -- transcribes to `map` in the Data Model.
+	- `stringpairs` representation -- transcribes to `string` in the Data Model.
```
rvagg, line 40: I think @mikeal has a visceral reaction to the mention of `stringpairs` previously. Is this something we're seeing in the wild or is it just a might-be-nice-to-have?

schema/representations.md
```diff
@@ -0,0 +1,256 @@
...
+- Union
+	- `keyed` representation -- transcribes to a single-entry `map` in the Data Model.
+	- `kinded` representation -- transcribes to varying (!) kinds in the Data Model.
+	- `envelope` representation -- transcribes to a dual-entry `map` in the Data Model.
+	- `inline` representation -- transcribes to a `map` in the Data Model (and has additional limitations).
+- Struct
+	- `map` representation -- the default -- transcribes to `map` in the Data Model.
+	- `tuple` representation -- transcribes to `list` in the Data Model.
+	- `stringpairs` representation -- transcribes to `string` in the Data Model.
+	- `stringjoin` representation -- transcribes to `string` in the Data Model.
```
rvagg, line 51: same as above comment about `stringpairs`, it's pretty icky, do we have reason to believe it'll be useful as more than a novelty?

schema/representations.md
```diff
@@ -0,0 +1,256 @@
...
+{"foo": {"froz": true}}
+```
+
+This data would also match, as the other type:
+
+```json
+{"bar": 12}
+```
+
+Note how kinded unions don't introduce any kind of wrapping object at all;
```
rvagg: *keyed unions

schema/representations.md
```diff
@@ -0,0 +1,256 @@
...
+
+```json
+12
+```
+
+Note that the syntax used in the type declaration is different for kinded
+unions versus other union representations!
+Kinded unions, unlike all other representation strategies, don't have
+any kind of string keywords at all.  Instead, they encode all of their
+discriminant information in the kind of the data itself!
```
rvagg, line 190: OK, now I get what you intended with "kinded", but it scares me because of the kind of mess that type coercion imposes on untyped languages like JavaScript. We're going to need to specify rules that differentiate the close types: int and float, if that's even possible:

```ipldsch
 type MyKindedUnion union {
	| Foo float
	| Bar int
} representation kinded
```
->
```json
1
```
or
```json
1.0
```
what do?

And how do we even handle that in JavaScript? If we decode a `1.0`, by the time it gets past the encoding layer we have some generic number type that we can't differentiate between a float and an int, it's a `1`, `typeof x === 'number'`. Is it a `MyKindedUnion#Foo` or a `MyKindedUnion#Bar`?

schema/representations.md
```diff
@@ -0,0 +1,256 @@
...
+match that type's representation kind or the schema is invalid.
+
+### union envelope representation example
+
+```ipldsch
+type MyEnvelopeUnion union {
+	| Foo "foo"
+	| Bar "bar"
+} representation kinded {
+	discriminantKey "tag"
```
rvagg, line 206: does it make sense to have default keys for envelopes? `type` and `value` maybe?

schema/schema-kinds.md
```diff
@@ -0,0 +1,159 @@
...
+**or** its place may be occupied by the null value.
+
+The `nullable` modifier is valid on map values, list values, and struct fields.
+
+The presence of a `nullable` modifier increases the cardinality of a field's
+valid members by one -- see the "Cardinality Examples" table below.
+
+### Optional Fields
+
+The `optional` modifier means that either the modified type
```
rvagg: ... incomplete sentence

rvagg:
Really great stuff! I reckon you could just fill in some of the small missing pieces and land an initial version of this that can be iterated over subsequent PRs rather than trying to max this one out.

Further to the comments about `stringjoin` and `stringpairs`: can we at least defer on these and tag them as "potential future representation types" if they have to be mentioned at all. That is unless they're unavoidable because they are in the wild today in blocks that we want to describe? Conceptually they really help illustrate the ideas but they're going to make it more complicated to get to an minimum viable implementation and they may even invite other novel representations that we won't have a good reason to reject.

schema/representations.md
```diff
@@ -0,0 +1,256 @@
...
+
+#### struct tuple representation example data
+
+```json
+```
+
+### struct stringjoin representation example schema
+
+```ipldsch
+## Fizzlebop is a pair of fields which serializes as "part1:part2" as a string.
```
vmx: That's not clear to me. I guess it will be: "value-of-a:value-of-b".

schema/representations.md
```diff
@@ -0,0 +1,256 @@
...
+
+### struct stringpairs representation example schema
+
+Say we're doing something awfully like the mount options in an /etc/fstab file:
+
+```ipldsch
+## MountOptions serializes as a string of comma-separated "k=v" entries, e.g. "k1=v1,k2=v2".
+## This serial representation is limited (n.b. The domain of keys must exclude "=" and values must exclude ","!),
+## but we've chosen it because other parts of our application protocol are drastically simplified by limiting this to a string.
+type MountOptions map {String:String} representation stringpairs {
```
vmx, line 127: The headline says "**struct** stringpairs…", but here the type is `map`.

schema/representations.md
mikeal: so, i see why having these is good in that it expresses what can be done by the schema beyond basic type bindings. my primary concern is just opening the door to “every feature ever” at this layer. like, how many of these “string*” features will we add to the spec and what is the reason for allowing at this level or not?

schema/representations.md
mikeal: we can’t make assurances about language level precision loss in this schema spec, all we can do is codify what these mean in language agnostic terms and direct each language to attempt to preserve precision as much as possible. we already do this at the data model layer in JavaScript with these types, we can’t force float types but we do try to preserve 64bit integers.

in other words, it’s fine to be as precise as possible at the schema layer as long as we understand that different languages can only make their best effort to support that precision.

schema/representations.md
warpfork: I'm not sure.

I think I'd vote against because it introduce more opinions into the core; and reduces the amount of strings which are can be expected to show up word-for-word in both the schema and the encoded data, which I think are a nice sanity synchronizer.  It's a weak vote, though.

(Also, I'll readily admit "discriminantKey" is heckin' long and a handful to type.  Maybe something terser here, or using positional arguments for this, would improve the feel.)

schema/representations.md
warpfork: Oh, boy.  That's rough.

I'm not entirely sure.  Something like this has come up before in refmt where some care was required around int vs uint bitwidth issues, but those were... more precise and more tractable issues than this.

warpfork:
One of the major virtues of `stringjoin` and `stringpairs` representation strategies on structs is that it gives us a way to have struct kinds which have `string` as their _representation_ kind.

This unlocks interesting things like being able to use *structs* as a *map key*... even though we've restricted map keys to being strings (obvious, in JSON for example, we can't have other full objects as map keys!).  By using the representation vs type kind distinction, we can define "map keys must have representation kind string", and then freely use structs there, as long as they can be suchly represented.

Another interesting place this comes up "in practice" in my experiences and wishlists of things I could use IPLD for: I have some structure which I want to pass around in command line flags.  Doing the parser dance for this manually, including all the defaulting and optional and so forth behavior, is both irritating and error-prone.  If I could make a quick jot of schema that gets all the cardinality clarified and give me the string I need, that's super useful.  (I want to use this somewhat often in parts of [the Timeless Stack](https://repeatr.io/schema/layer-2/) schemas; the filesystem options strings in `/etc/fstab` are another interesting example of this.)

I agree we'll want to keep from getting carried away here, but a handful of things that can do this representation jump could be incredibly useful.

Perhaps we can further alleviate the work required for new schema library implementors by writing some auxillary spec notes which describe tiers of support?  None of these fancy representations for structs would be in the core that you'd need to parse the schema-schema, for example; so we could kick them to another tier.  (Once one has all the other features to parse the schema-schema, I suspect adding things like the stringjoin representation should be easy-peasy; it's just daunting if it's all heaped on the todo list at once.)

schema/representations.md
warpfork: derp good catch

schema/representations.md
warpfork: I think we can get *pretty* far with some sort of rule like: the deserialization code for ambiguous cases should prefer to report the more constrained type (e.g. int over float) until forced not to by the data.  (`1.0` in JSON would be considered forcing a float.)

(Precision at the serial level is only a JSON issue.  CBOR doesn't have this problem -- it has tagged floats: there's a nibble in the serial encoding which makes them completely unambiguous from ints.)

In javascript, it's probably going to turn out required to keep a kind tag associated with the token from the deserialization parse layer.  That's not the prettiest, but I don't see how it can be avoided.

And after all that, it still might be "not recommended" to try to use a kinded union of int-or-float because of the potential for ambiguity.  Even with these rules in _our_ code, there's plenty of other stuff out there we'll interact with which might have numerical precision choices that turn a `1.0` into a `1` in JSON, and there's nothing we can do about that other than caution people to use tools wisely.

schema/representations.md
warpfork: wait, that whole paragraph is in the wrong section :thinking: how'd that get there?

Fixed, thanks for the catch

schema/representations.md
rvagg: "not recommended" sounds OK I suppose. A clear set of rules about what path is likely to be chosen when there's ambiguity might help too but a schema author is unlikely to know all possible values their code can take (or may not think this through) and might make a false assumption that all their floats will have a fractional part.
I can imagine having to let the tag bubble up from the serialization layer become a problem when we're trying to separate out these layers nicely. In fact, I'm not sure it's even possible in the JS incarnations as they exist today (old and new proposed).

schema/representations.md
```diff
@@ -0,0 +1,257 @@
...
+`struct` is not.  The kind that a union member type is tagged with must
+match that type's representation kind or the schema is invalid.
+
+### union envelope representation example
+
+```ipldsch
+type MyEnvelopeUnion union {
+	| Foo "foo"
+	| Bar "bar"
+} representation kinded {
```
rvagg: `} representation envelope {`

schema/representations.md
```diff
@@ -0,0 +1,257 @@
...
+{"tag":"bar", "msg":12}
+```
+
+### union inline representation example
+
+```ipldsch
+type MyInlineUnion union {
+	| Foo "foo"
+	| Bar "bar"
+} representation kinded {
```
rvagg: `} representation inline {`

the schema-schema skips `discriminantKey ` and uses a quote string instead of this block, so it would be `} representation inline "tag"`. Is that intended to be supported as a shortcut as well as this long-form block?

schema/representations.md
```diff
@@ -0,0 +1,257 @@
...
+{"bar": 12}
+```
+
+### union kinded representation example
+
+```ipldsch
+type MyKindedUnion union {
+	| Foo map
+	| Bar int
+} representation kinded
```
rvagg, line 163: I think we should drop the kinds from kinded unions and infer them from the info we already have on those types (we know `Foo` is a map and we know `Bar` is an int). The parsing pattern is broken when kinded unions have unquoted kinds in each element descriptor but the other union types have quotes strings for the various discriminators. This is going to make learning more difficult, particularly for example-driven folks who will keep on seeing weird discrepancies.

I agree that moving the discriminators down into the `representation` block would become tedious very fast, requiring a doubling up of each element. But having it up in the main block undermines your original ideal of keeping representation-specific information out of the main block. Maybe if we're doing a special case for unions we should consider how much of the representation block can be removed. A type can only be of one representation anyway, it's not as the main block needs to be kept clean for multiple representation forms, they're already tied together.

I can at least see any of the `field` descriptors in the representation needing to be lifted out to the main block by this same logic. And while at it, I'd put the name _after_ the type to bring them closer to unions. I'd even be tempted to quote them for the same reasons.

```
type TypeList struct {
	foo Foo
	bar Bool
	baz Int
} representation map {
	field foo optional
	field bar nullable default "false"
	field baz optional nullable
}
```

->

```
type TypeList struct {
	Foo "foo" optional
	Bool "bar" nullable omitted "false"
	Int "baz" optional nullable
} representation map
```

as a tuple:

```
type TypeList struct {
	Foo "foo"
	Bool "bar" nullable
	Int "baz" optional nullable
} representation tuple
```

Maybe the rule is that anything in the `representation` block should be required generally for the representation (discriminantKey etc.) rather than specifically for individual fields. Anything needed for the fields should go in the main block coupled with the fields.

The parser will then still have to backtrack and error if it finds field descriptors with too much, or the wrong kind of information. Front-loading a representation block doesn't seem to work (too messy), but just collapsing it in might:

```
type MyEnvelopeUnion union {
	| Foo "foo"
	| Bar "bar"
} representation envelope {
 	discriminantKey "tag"
 	contentKey "msg"
}
```

->

```
type MyEnvelopeUnion union envelope {
 	$discriminantKey "tag"
 	$contentKey "msg"
	| Foo "foo"
	| Bar "bar"
}
```

That's fantastic for parsing, but maybe I making too much of a dogs breakfast of your vision?

rvagg: OK, maybe scrap that last bit, I'm reflecting on its ugliness. I'd still like to explore the fields problem though because the representation is already mixed up in the definition and we're sitting in an awkward middle that I think we can move out of.

schema/representations.md
warpfork: > my primary concern is just opening the door to “every feature ever” at this layer.

100%.  It's a super legitimate concern.  There's clear reasons for "some" features of this form.  It would also clearly be extremely problematic if this got out of hand with too many extensions.  Every new member in this set of known representations adds implementation work to every single IPLD library that will ever be written -- and if we were to add new members after the first stable versions are released, will create partial compatibility tables between libraries (!!) -- and that should not be taken lightly.

I don't know how we police future choices other than always being clear that the discussion starts at "default: No." and require any more additions make a very strong case for themselves.

One rule of thumb that might help is "if it seems turing complete, hard No".  Similarly, "if the parser would be recursive, hard No".  I might suggest "if it requires bittwiddling or math, lean against"; "if this seems like a one-off for someone's pet project, No"; "if there aren't existing of examples of this pattern done in the wild, No"... perhaps the biggest one might be "Does this enable other features that aren't otherwise possible, or have radically different performance characteristics if carried out by the library?"    This last question is why we're here in the first place (the answer is "yes" because it lets us have complex map key types, which was otherwise intractable).

Beyond that, discretion and caution.  And (rapidly) increasing reticence over time, due to that partial compatibility concern.

schema/representations.md
warpfork: Not a fan of dropping kinds from the declaration syntax of kinded unions -- it *is* redundant, for sure, but imo it's going to be a useful redundance, and it's a useful lexically local clarity of intent (i.e. I wanna *see* it in the union defn while I'm typing it out if I accidentally use two things that are a map) and it makes for clearer error reporting (i.e. if I type out a lie there that doesn't match the member type defn's kind, it's very clear exactly which lines to point at as the error, rather than having to point at a whole range of lines and leave the user to figure out which one is conflictory.)

To the rest of it ... and per conversation in other channels... Yeah, I'm super willing to entertain (if not already full on convinced) there's some weird bits here, and maybe we could reach a better and more internally consistent thing.

Getting there is gonna be bigger than fits on the discussion of one PR though; let's land this one, call it a checkpoint, and open more, yeh?

warpfork: Just to toss this out there, one variation in syntax I've been trying out is:

```
type SomeStruct struct {
    fieldname nullable OtherType (alias "foo")
}
```

The rule here would be "things that would have gone in the representation block, but are specific to a field... go after the field, but in parens".  (Thus, modifiers like "nullable" still stay outside the parens -- they're still a cardinality changer, so they're still clearly a type-defn thing.)

But this is just another idea.  And not the one in this PR.  Just airing it here as an idea of a direction we could examine further in a future PR.

warpfork:
It looks like I've got one incomplete sentence to finish (!) and a couple of typos in the examples to fix...

And then if there's no other objections, I think I'd like to merge this shortly.  It's not "done", but it's feeling like a good checkpoint. :)

schema/representations.md
rvagg: That kind of thought process would be good to capture somewhere, not just because it'd be amusing to see you try and codify that into documentation form; these are the kinds of details that get lost with people-churn so boundaries start to slip because nobody knows where they should be.

rvagg:
Yeah, I agree, let's get this landed so we don't leave it in limbo and get tangled into PR-onto-PR land. I've aimed [js-ipld-schema](https://github.com/rvagg/js-ipld-schema) at this current form but I'd like to experiment with the per-field representation properties in parens idea that you've outlined above because I'm suspecting that solves a lot of my concerns about bulky `representation` blocks. That could happen with a separate PR for isolated discussion.

The kinded union thing I think goes back to a bigger problem of explaining the overall model. It's taken me far too long to get into the right mindset about how this all works. But that could probably be solved with clear documentation (rather than making people figure it out with a schema-schema!) so maybe I'll work on fleshing some of that out when you land this.

schema/representations.md
```diff
@@ -0,0 +1,257 @@
+
```
sansato, line 1: This is interesting information.  I'm liking some of principles in discussions.  As a follower of the project, I have some suggested adds.  It's a little hard to understand at first.  A few improvements to suggest:

 * A Background section with criteria and purpose to motivate.  For example, "IPFS applications like domain-specific applications and generic data-crawl software will traverse a graph of data found in IPFS.  They will discover sub-graphs and identify the type of schema-definitions found.  If they support,  they parse the data and continue with data traversal.  If they don't support the schema, they still identify the schema format that's not supported. Lookup service to find support?  Given a CID referencing a data block, a schema-aware IPFS client can identify the schema type from data in the block, schema for that data block, and show the found data".  And so.

 * What are evaluation criteria - moving some of the discussed principles into the document.  

 * Certain use-cases providing design purposes.  Encoding filesystem-mounting semantics?  Broad vision.  Examples of mount filesystems from IPFS/IPLD data: read-only, read-write, cache hints - specific, not generic.  IMDB data might have good examples too.  Put concrete examples in an introductory section as backstory, then reference them inline, not using "Fizzlebop".  If no Filecoin references, I'm sad.

 * Strategy for forward-compatibility: Are multiformats or in-block schema-CID refs planned?  Will a block of data be self-describing?  Will you later support new use-cases and introduce new schema capabilities?  How?  Will IPLD schema be open, so that non-PL contributors will be able to define IPLD data and software plugins for today's  encodings & data schemas?  How can metadata be encoded so that current and future software are strong with such multi-schemas?  Will they follow all data links because CBOR?  If not, Unsupported section can be useful.  Links to other documents can organize some of info that now exists.

 * How is avoided https://www.xkcd.com/927/ ?  JSON-Schema, protobuf and Avro should be mentioned.  What good or bad points, and how proposal takes their best points to improve within the IPFS ecosystem.  Can IPLD support data using these encodings?  Or unsupported? Help others know what it is/isn't and how to get there from here.

Keep the good work guys.

rvagg: :thumbsup: I think the path to doing some of this is to just get this merged and then start to fill out these pieces. This is a good TODO list though.

schema/representations.md
```diff
@@ -0,0 +1,257 @@
...
+{"tag":"foo", "msg":{"froz":true}}
+```
+
+This data would also match, as the other type:
+
+```json
+{"tag":"bar", "msg":12}
+```
+
+### union inline representation example
```
rvagg, line 226: I think that it might be the case that an inline union can only be a union of structs with map representations, it doesn't work any other way, does that sound right?

schema/representations.md
warpfork: _engages holodeck poker room_

I'll see your preemptive schadenfreude, and raise you one epistemological dread!

 . . .

More seriously, though, yes, we should try to capture that somewhere; thoughts on how to draft that would indeed by welcome.

schema/representations.md
warpfork: I'll switch to that phrasing.

schema/representations.md
warpfork: > intended to be supported as a shortcut

Unsure.  I think that line in the schema-schema might be an artifact of an earlier round of drafting.  The shortcut might be nice (if, indeed, we're sure it'll often be a single-arg use); otoh, the parser would be more consistent if we didn't add such shortcuts.  I think we're still open to going either way, and as someone who's written a grammar, I think your opinion counts more than mine here.

schema/representations.md
warpfork: confirmed.

warpfork: Apparently this is currently documented in the [schema-schema](https://github.com/ipld/go-ipld-prime/blob/f2d595797253f736988a9a2402c2a5a0a05d4e93/typed/declaration/schema-schema.ipldsch#L254-L255), but might not have made it into this text yet.  We should do a pass on that whole thing and make sure everything in it gets represented in prose somewhere over here.

warpfork:
Okay, think I've fixed all the immediate todos and wrapped up the one dangling section.  Gonna take the previous "let's get this landed so we don't leave it in limbo" comment as gold and merge this.  Lots more to do, but more PRs shall flow!

PR #114, opened 2019-04-24
add a spec for dag-pb

whyrusleeping:


Codecs/DAG-PB.md
```diff
@@ -0,0 +1,55 @@
...
+The DAG-PB IPLD format is a legacy format implemented with a single protobuf.
+
+```protobuf
+// An IPFS MerkleDAG Link
+message PBLink {
+
+  // multihash of the target object
+  optional bytes Hash = 1;
+
+  // utf string name. should be unique per object
```
Stebalien, line 16: utf-8

Codecs/DAG-PB.md
```diff
@@ -0,0 +1,55 @@
...
+  optional string Name = 2;
+
+  // cumulative size of target object
+  optional uint64 Tsize = 3;
+}
```
Stebalien, line 21: Are these the names used in IPLD? We need to be consistent with go/js here.

Codecs/DAG-PB.md
```diff
@@ -0,0 +1,55 @@
...
+This is non-standard and means that the 'Tsize' attribute of a link, and
+neither the 'Data' field nor the 'Links' field of the PBNode are available
+through the ipld query language. 
+
+## Link Format
+
+In DagCBOR, links are encoded using the raw-binary (identity, NUL) multibase in a
+field with a byte-string type (major type 2), with the tag 42. They are then
+put in the 'Hash' field of a PBLink. All links from a DAG-PB object are
+specified in the 'Links' array.
```
Stebalien, line 44: We just use the raw CID, minus the multibase.

Codecs/DAG-PB.md
```diff
@@ -0,0 +1,55 @@
...
+
+  // opaque user data
+  optional bytes Data = 1;
+}
+```
+
+The objects link names are specified in the 'Name' field of the PBLink object.
+This is non-standard and means that the 'Tsize' attribute of a link, and
+neither the 'Data' field nor the 'Links' field of the PBNode are available
+through the ipld query language.
```
Stebalien, line 37: This is a nasty wart. We _should_ just be using paths of the form `/ipld/Qm.../Links/0/Hash` etc. We could also do an implicit transform to allow `/ipld/Qm.../Links/$Name` (where `Size` can't be addressed with a path) but that's still icky.

Codecs/DAG-PB.md
```diff
@@ -0,0 +1,55 @@
...
+## Canonical DAG-PB
+
+Canonical DAG-PB must:
+
+1. Contain only the specified protobuf fields.
+2. Use standard protobuf encoding, except that the 'Links' field must be before
+   the 'Data' field in the encoding. This is due to a bug in the initial
+   protobuf encoder that was used in the first implementation of ipfs.
+3. Only use string map keys. Some implementations may not be able to
+   handle non-string keys.
```
Stebalien, line 55: What map?

Codecs/DAG-PB.md
```diff
@@ -0,0 +1,55 @@
...
+specified in the 'Links' array.
+
+## Canonical DAG-PB
+
+Canonical DAG-PB must:
+
+1. Contain only the specified protobuf fields.
+2. Use standard protobuf encoding, except that the 'Links' field must be before
+   the 'Data' field in the encoding. This is due to a bug in the initial
+   protobuf encoder that was used in the first implementation of ipfs.
```
Stebalien, line 53: The protobuf spec has further softened ordering requirements so we should be explicit: The ordering of the fields is `Links[{ Hash, Name, Tsize }, ...], Data`

Stebalien (changes requested):
(cr in pr: https://github.com/ipld/specs/pull/115)

Codecs/DAG-PB.md
vmx: Yes. When https://github.com/ipld/js-ipld-dag-pb/pull/127 the JS implementation uses only those names, so that pathing through objects on an IPLD level (not UnixFS/IPFS-level) is the same in Go and JS.

Codecs/DAG-PB.md
vmx: The JS implementation currently exposes named links as `/Qm.../<named link>/` so that it is similar to IPFS-style pathing.

Stebalien: How does it expose numbered links?

vmx: e.g. `Links/0`. Yes it's right, there can't be a link named `Links`. This is the behaviour it had before I joined and I just kept it for now.

Stebalien: Got it.

vmx:
Closing this one if favour of #201.

PR #115, opened 2019-04-25
dag-pb-spec: steb CR

Stebalien:
1. Fix link/hash encoding (not CBOR).
2. Describe a bunch of options for pathing.
3. Expand on canonical encoding.

### DIFF
--- a/Codecs/DAG-PB.md
+++ b/Codecs/DAG-PB.md
@@ -10,10 +10,10 @@ The DAG-PB IPLD format is a legacy format implemented with a single protobuf.
 // An IPFS MerkleDAG Link
 message PBLink {
 
-  // multihash of the target object
+  // binary CID (with no multibase prefix) of the target object
   optional bytes Hash = 1;
 
-  // utf string name. should be unique per object
+  // UTF-8 string name.
   optional string Name = 2;
 
   // cumulative size of target object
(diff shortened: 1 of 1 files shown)


Codecs/DAG-PB.md
```diff
@@ -32,24 +32,69 @@ message PBNode {
 ```
 
 The objects link names are specified in the 'Name' field of the PBLink object.
-This is non-standard and means that the 'Tsize' attribute of a link, and
-neither the 'Data' field nor the 'Links' field of the PBNode are available
-through the ipld query language. 
+All links in an object must either be blank or unique within the object.
```
rvagg, line 35: A link itself can be blank? Or PBNode.Links can be blank? If the former, does that mean that "optional" in all the PBLink fields are truly optional? In the JS implementation they're not, a CID is required and links created from within JS will have a default name of empty-string if none is supplied.

Codecs/DAG-PB.md
```diff
@@ -10,10 +10,10 @@ The DAG-PB IPLD format is a legacy format implemented with a single protobuf.
 // An IPFS MerkleDAG Link
 message PBLink {
 
-  // multihash of the target object
+  // binary CID (with no multibase prefix) of the target object
```
rvagg, line 13: Would "binary version 0 CID (no multibase prefix) of the target object" be accurate? It might be nice to be explicit about why this is odd.

Codecs/DAG-PB.md
```diff
@@ -32,24 +32,69 @@ message PBNode {
...
+## Pathing
 
-In DagCBOR, links are encoded using the raw-binary (identity, NUL) multibase in a
-field with a byte-string type (major type 2), with the tag 42. They are then
-put in the 'Hash' field of a PBLink. All links from a DAG-PB object are
-specified in the 'Links' array.
+TODO: See https://github.com/ipld/specs/issues/55
+TODO: Also see https://github.com/ipfs/unixfs-v2/issues/24 (specifically: the "Revise History" option).
+
+Neither go-ipfs nor js-ipfs agree so we have some room here because we're going to break something.
```
rvagg, line 42: How about for the purpose of the spec, the only required path mechanism that need be implemented is the `/Links/0/Hash` style. In js-ipld-dag-pb the names are just sugar and I'm guessing that the go implementation does the `/named` style as sugar as well and it just doesn't expose a resolution that can get into the paths array? Maybe it should and that's the consistency between the two. `/Links/named/Hash` is the JS sugar and `/named` is the Go sugar and that's just how it is.
Either way, that's a blocking issue that's going to hold this up so it needs to get resolved.

Codecs/DAG-PB.md
```diff
@@ -32,24 +32,69 @@ message PBNode {
...
+paths.
+
+It's also impossible to path through nodes that don't name their links.
+
+### Alternative: js-ipfs
+
+As far as I can tell, js-ipfs supports pathing both by name and by index with
+paths like: `/Links/$name/Hash/...` or `/Links/$idx/Hash`.
+
+### Alternative: Correct IPLD
```
rvagg, line 57: Yeah, this one. Would it be difficult to implement this in the Go version? (I can't find the code) Having this unresolved and listing "Alternative"s isn't a very helpful spec. Listing what the two versions do is helpful, it'd be nice to have some concrete middle for the spec.

Codecs/DAG-PB.md
Stebalien: Sorry, that should have said "link names".

Codecs/DAG-PB.md
Stebalien: Any CID can go here. The key part is that we're not including the multibase (which we sometimes (e.g., CBOR) do include in binary).

Codecs/DAG-PB.md
Stebalien: I'd rather not leave sugar like that. We can get rid of it but `resolve(node, '/some/path')` should behave the same way everywhere.

Codecs/DAG-PB.md
Stebalien: These sections are just here for discussion. The final spec will have just one.

Switching to "correct IPLD" will take a bit of work in go-ipfs but we'll have to break something no matter what.

vmx:
Closing this one if favour of #201.

PR #116, opened 2019-04-29
Selectors++ (recursive design)

warpfork:
This set of specs is for a more recursive strategy of selectors, and is
much more expressive.

It's also, importantly, accompanied by a working proof-of-concept
implementation in the go-ipld-prime repo, which has kicked the tires.

Credit where it's due: the previous stuff, with its linear stack of
selectors, despite being 'git blame' to vmx, was mostly me; and this
stuff, which is much more excellent, is actually closer to what vmx was
originally proposing in our earlier discussions.
So in other words, I came around eventually :)

We turn out to have union selectors *in*!  This is remarkable because
it's a feature we were trying to avoid until future rounds.
The change of heart is because it turns out that unions are actually
a feature that's *implied* by some of the other selectors (namely,
recursives) when implemented internally... so the additional work to
then just expose that feature turns out to be negligible.  Tada!

This may still not be the final resting place of this spec.
I'm baking on another round of drafts which attempt to more fully
separate "matchers" vs "explorer" parameters -- you can see some of
this concept starting to show up in the docs already in this commit,
though it's not made fully concrete anywhere yet.  These drafts are
as-yet incomplete, but fueled by the hope that it may make some parts
of recursives a bit clearer... as well as leave us better set up for
future fancier node match conditions which can use logic that's
orthagonal from the exploration guidance.  But that's a mouthful.
We'll see.

### DIFF
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -66,7 +66,7 @@ These other questions are solved in other systems, above or below IPLD Selectors
 > - How can we make selector combination efficient? (IPLD library implementations)
 > - How can we strip a dag subset from links to objects not in the subset (IPLD Transformations, user code)
 > - How can we convert one dag representation into another? (IPLD Transformations, user code)
-> - How can we agree with multiple other parties who will pin or send subets of a dag? (ipfs-cluster, graphsync)
+> - How can we agree with multiple other parties who will pin or send subsets of a dag? (ipfs-cluster, graphsync)
 
 With our problem narrowed down, we can focus on solving just that, and let all other concerns be solved elsewhere in the stack.
 
@@ -113,19 +113,29 @@ These components imply or expand into the following things:
 Selector types
 --------------
 
-### CID Selector
+### Fields Selector
 
-Selects a single CID.
+Selects each of a set of path segments.
+
+(If only a single path segment is specified, this can be called a "Path Selector", but it's the same thing.)
 
 
 ### Path Selector
 
 Selects a single path segment.
 
+Path selectors are a common pattern of usage, but not a specific feature in the selector spec --
+it's a degenerate case of a Fields Selector which happens to select only a single field.
 
-### Array Selector
 
-Selects a slice or single item of an array of items.
+### All Selector
+
+Selects every field or index in a map or list.
+
+
+### Range Selector
+
+Selects a slice or single item of a list of items.
 
 
 ### Recursive Selector
@@ -135,9 +145,10 @@ Uses another Selector to select things recursively. It also takes a limit on how
 The recursion may also be stopped by a given CID.
 
 
-### CID Rooted Selector
+### CID Selector
 
-The CID Rooted Selector consists of a CID where the selection starts and a list of Selectors specifying the traversal.
+There is no CID selector.  Asking for a specific CID doesn't involve any traversal; it's just asking for a node itself.  No selectors are necessary to ask for data by CID.
+
 
 
 Use cases
@@ -147,7 +158,7 @@ Use cases
 
 In some DAG you want to get one specific value you know the path of. Let's say you want to get the birth year of a specific character of a specific show.
 
-Example data (as JSON) with CID `cidabcdef`. To keep it simple it's represented as a single JSON structure, but imagine e.g. the characters are linked by CID.
+Example data (as JSON):
 
 ```json
 {
@@ -165,20 +176,14 @@ Example data (as JSON) with CID `cidabcdef`. To keep it simple it's represented 
 }
 ```
 
-The final Selector could look like this:
+A Selector to extract the "year" data could look like this:
 
 ```json
-{
-  "cidRootedSelector": {
-    "root": "cidabcdef",
-    "selectors": [
-      {"selectPath": "characters"},
-      {"selectPath": "kathryn-janeway"},
-      {"selectPath": "birthday"},
-      {"selectPath": "year"}
-    ]
-  }
-}
+{"selectFields":{"characters":
+	{"selectFields":{"kathryn-janeway":
+		{"selectFields":{"birthday":
+			{"selectFields":{"year":
+				true}}}}}}}}
 ```
 
 
@@ -199,37 +204,39 @@ The shape of a block could look like this (in JSON):
 If you know you want five parents you could use Path Selectors:
 
 ```json
-{
-  "cidRootedSelector": {
-    "root": "cidabcdef",
-    "selectors": [
-      {"selectPath": "parent"},
-      {"selectPath": "parent"},
-      {"selectPath": "parent"},
-      {"selectPath": "parent"},
-      {"selectPath": "parent"}
-    ]
-  }
-}
+{"selectFields":{"parent":
+	{"selectFields":{"parent":
+		{"selectFields":{"parent":
+			{"selectFields":{"parent":
+				{"selectFields":{"parent":
+					true}}}}}}}}}}
 ```
 
-But there's also a better way with using a Recursive Selector with a limit:
+This selector matches the fivth-deepest "parent" (and in the context of
+graphsync or other merkleproof applications, will yield all five nodes).
+
+But this gets a bit verbose.  We can explore the same tree in a similar
+pattern with another mechanism -- recursive exploration:
 
 ```json
-{
-  "cidRootedSelector": {
-    "root": "cidabcdef",
-    "selectors": [
-      {"selectRecursive": {
-        "follow": [
-          {"selectPath": "parent"}
-        ],
-        "depthLimit": 5
-      }}
-    ]
-  }
-}
+{"selectRecursive": {
+  "depthLimit": 5,
+  "next":
+    {"selectFields":{"parent":
+      true}}
+}}
 ```
+
+This will traverse the same set of nodes as the previous example -- however,
+it has has a *slightly* different effect!
+
+Using a recursive selector in this way matches *each* of the "parent" nodes,
+up to the depth limit -- meaning it matches five nodes, instead of the
+previous example, which matches only the last one.
+
+In terms of graphsync or other merkleproof applications, the selector will yield
+the same set, since the set of nodes *considered* for matches is the same as
+in the previous example.
 
 
 ### Getting changes up to a certain one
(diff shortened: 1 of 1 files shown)


selectors/selectors.md
```diff
@@ -199,38 +204,40 @@ The shape of a block could look like this (in JSON):
...
+  "next":
+    {"selectFields":{"parent":
+      true}}
+}}
 ```
 
+This will traverse the same set of nodes as the previous example -- however,
+it has has a *slightly* different effect!
+
+Using a recursive selector in this way matches *each* of the "parent" nodes,
```
mikeal, line 233: Can we use a different property value than “parent.” It’s not clear if this refers to a property because parent is such a broadly used term.

rvagg:
You know the [English spelling reform joke](https://www.smart-jokes.org/english-spelling-reform.html) where, over the telling of the joke, English slowly transforms into German? I'm getting shades of that with GraphQL here.

warpfork:
GraphQL was pretty cool :woman_shrugging: 

If we end up with similar-ish things, I wouldn't be mad.  Most of the beefs I have with graphQL are based on the fact it's got client-server concepts baked really deeply into its mindset.  And we don't brook with "servers".

Also GraphQL queries getting sent as one giant DSL string *inside* a JSON object makes me eyebrow-raise.

Also we're going to end up in a little different place if the plan to use selectors to also guide walk/visitor functions on data shapes up (in which case we'll have distinctions between 'match' and 'explore', which graphQL doesn't).

Okay, so there's a lot of things that are different.  But yeah, GraphQL remains cool.

selectors/selectors.md
```diff
@@ -273,104 +275,94 @@ An example selector to get the full sub-DAG rooted at a certain CID:
...
-          {"selectPath": "Links"},
-          {"selectArrayAll": null},
-          {"selectPath": "multihash"}
-        ]
-      }}
-    ]
-  }
-}
+{"selectRecursive": {
+  "depthLimit": 5,
```
vmx: The `depthLimit` shouldn't be there.

selectors/selectors.md
```diff
@@ -273,104 +275,94 @@ An example selector to get the full sub-DAG rooted at a certain CID:
...
+{"selectFields":{"with":
+  {"selectFields":{"some":
+    {"selectFields":{"subdirectory":
       {"selectRecursive": {
-        "follow": [
-          {"selectPath": "Links"},
-          {"selectArrayAll": null},
-          {"selectPath": "multihash"}
-        ]
+        "depthLimit": 5,
```
vmx: The `depthLimit` shouldn't be there.

vmx (commented):
In your proposal there is no CID Selector anymore anymore. I think that made sense to make CIDs part of the Selectors. As then (in the future) you could have Selectors that are not bound to a CID. The one I really want is e.g. the one that gets several items just by CID.

hannahhoward:
Please see #117, which is a neccessary change to Graphsync if we remove CIDs from selectors

warpfork:
We've gone back and forth on where to put the CIDs for a while... So imma try to kinda destroy this...

The one reason to be in favor of putting CIDs _in_ selectors that I'm aware of is that we often expect to see them on the wire next to each other.  I think this is a weak argument in the first place -- one of correlation rather than intent -- and it's countered by a pretty sizable list of much stronger counterarguments.

Con 1: a "CID selector" doesn't make much sense anywhere but the first position in a selector.  So it seems clear that it shouldn't be in the Selector union, at least.

Con 2: a Selector is something I want to use in the [Traverse method](https://github.com/ipld/go-ipld-prime/blob/cd9283ddd86af15b9bc4a1f5b71f00fbfc2f8b94/traversal/traverse.go#L35) to guide walks over a data structure, and that method starts with a Node, because otherwise it wouldn't be possible to sanely use Traverse() inside of the visitFn of another Traverse()!  (The visitFn receives the already-loaded Node -- and again, must, because it can land on a *mid-block Node* which might not *have* a CID -- so it would be unwieldy to make it only possible to launch another Traverse by forcing the user code to load that same thing again from CID, and sometimes not even possible.)

Con 3: Let's doubledown on that mid-block node issue: we should be able to use Selectors to traverse any part of an IPLD node graph, including starting on nodes that aren't at the root of a serialization block.

I think I could keep going on the "cons" to coupling here, but these should be strong enough already.

One more reason I can think of that would be a "pro" for CIDs-in-selector (but it's a bit of a stretch): it would make it possible to use "union"-mode selectors to do a union of CIDs and their deeper selectors.  But does this even hold up?  IMO, no, in several different ways: the implementation for it is gonna be... strange... and not look at all like any of the rest of selector internals, since it's jumping to a disconnected new graph rather than exploring a subgraph.  It would also not be a particular *good* compact way of expressing things if you wanted to apply the same selector to multiple cids -- doing this with unions as we have them here would require you to dupe the rest of the selector body; what you'd rather do is have a message at some other layer that has multiple CIDs in it and then a single selector (or CID of a selector!) that should be applied to each of them.  So overall this just turned into more counterarguments and reasons to put them elsewhere again, not a 'pro' at all.

Total score: 
- CIDs-in-selectors: +0.5; -4.  (And that final demerit is the result of a train of thought I was *trying* to turn into a "pro", even!)

---

I think it's potentially valid to consider having a *different* type for `type CIDAndSelector struct {...}`, which is *not* a member of any of the other unions in this schema, and document that as a pattern that is valid for applications to use.  Applications could then choose to use *either* the `Selector` type or the `CIDAndSelector` type on the wire.  But this is a different discussion, and compatible with "CIDs are not Selectors".

warpfork:
Okay, I'll undercut myself.  There's one user-story that would blow my "Con 1" out of the water and lead to some serious reconsiderations.

Can we think of an occasion in using selectors where we'd ever want to...

- start at some point in a dag
- traverse some stuff via path (or other selection criteria)
- and then *only* select things by CID -- as it, we won't select this CID *unless we reached it via the earlier steps*, and we won't select anything *else* either?

If we have a use for that, then... that's interesting.

But I also have a feeling that kind of usage would need significant thought and still wouldn't actually resemble a selector itself.  It seems closer to how recursive selectors have a "stop" CID as a *property* of that selector -- though this would be the inverse, of course.  Also, notice how the "it jumps to a disconnected new graph" lament in my previous comment also wouldn't apply here.

So in other words this still seems to me like we should A) take out the not-very-fully-baked concept of CID selectors now and then B) come back to this later and create a fully formed proposal as a new PR.

warpfork:
Final, tangential comment on the CIDs thing:

It's my experience that anytime I'm working with posix-y paths in nontrivial quantities (i.e., *really* munging some files, not just like "eh, load a config file and move on"), [Relative Paths](https://github.com/polydawn/rio/blob/08fb8cc73674ef1f9db75b16e1450f3d27e03ac5/fs/path.go#L15) and [Absolute Paths](https://github.com/polydawn/rio/blob/08fb8cc73674ef1f9db75b16e1450f3d27e03ac5/fs/path.go#L160) should be treated as separate types.  The sanity improvements and bug avoidance reaped by declaring separate types for these is just massive.  (You only have to have an application accidentally "rm-rf /" because it failed to distinguish between a zero-value RelPath vs a zero-value AbsPath *once* in order to have a visceral reaction to this...)

I feel this applies to CID-rooted Selectors as well.  They're like absolute paths.  Selectors are like relative paths.  And it's going to be wise to keep those distinguished at a type level.

vmx:
@warpfork Good points and I agree that the CID should *not* be part of the Selector.

Though I'd like to point out that it doesn't come from changing my point of view, but rather my perspective (no idea if that actually makes sense in English). What I mean by it: I think we were looking at Selectors from two different perspective. Your perspective is an internal one, "how do I traverse a DAG with some kind of description". Mine was rather "how am I going to use this as a user, I need to start *somewhere* to traverse".

I agree that we should concentrate on this internal view, hence not making the CID part of the Selector. Once we get to a proper user-friendly syntax (if we ever), then we probably have something which starts with a CID. But that's kind of a level above the Selector. You would first root the selector somewhere. It could be a CID or also a CID + Selector, to actually get to the Selector you wan to run (e.g. the contents of a file within a directory in UnixFS. There the recursive Selector doesn't start at a CID, but at the thing the previous Selector pointed to).

mikeal:
> Con 1: a "CID selector" doesn't make much sense anywhere but the first position in a selector. So it seems clear that it shouldn't be in the Selector union, at least.

Any data structure that only appends new state as a new root with a reference to the old state’s root would want to make use of a CID as both a “starting point” **and** an “ending point.” Typically, you sync a graph and know the old root, you find out there is a new root and in order to get the changes since that last sync you would say “give me all the data form CID-1 down a recursive path selector to CID-2.”

I know this is typical in blockchains but I can also see this being useful in any data structure that keeps track of its own history.

I can also see a similar case in which a time series data structure might have a similar structure and could make use of this.

warpfork:
Yeah, I came up with at least one application where I can imagine actually wanting to use a `./foo/**/{CID}/bar` selector.

(It's using it to do a *search*, which is pretty heckin' questionable in some other ways (namely, it's probably really high cost to evaluate it in the application I'm imagining), but, it's a use case.)

I'm still kinda suspecting this would be best implemented by adding some more fields to existing selectors though -- like how the "stop" cid is a feature of the recursive selector rather than being a selector itself.

I'll give another shot at this on the next draft.  I think the "matchers and explorers are different" approach might make this all shake out clearer.  Not 100% sure until it's written though.

mikeal:
> I'm still kinda suspecting this would be best implemented by adding some more fields to existing selectors though -- like how the "stop" cid is a feature of the recursive selector rather than being a selector itself.

That’s a good point. CID’s, and even paths to some extent, aren’t enough information by themselves to be useful as a selector, you need to know if you are meant to match against them and what you are expected to do when you do or do not match.

warpfork:
Okay, updates on the status of this:

- I have more drafts of selectors on my localhost;
- Not all of them have gotten all of the examples fully rendered as I've iterated quickly over drafts (and these examples aren't enough to hit all interesting cases anymore, either);
- Therefore at this point, moving forward probably involves composing some meta-analysis of these; and also, I'm starting to think it desperately requires breaking this file up (high-level prose in one file; schema in one file; example data in one file; rationale and history in a file; at least those four).

Since we've also had a re-org PR land since this started, I'm thinking I'm going to rebase this, land some other drafts in merge-ignore form, and then prepare to land another batch of all this stuff.  As a result, further review of this can just wait until that next round.  All the comments here have been good and insightful and remain in effect for the other drafts I now need to get ready to publish, though :)

PR #117, opened 2019-04-30
Graphsync: add root cid

hannahhoward:
This is a companion PR to #116, which removes CIDS from selectors, making them entirely recursive. If PR #116 is accepted, Graphsync will need to treat the first CID lookup as seperate concern from performing a selector traversal. 

The PR proposes adding a CID root as a member in a graphsync request. Graphsync itself would then be responsible for using the loader function to look up and transmit the first block. This would not add a second round trip -- it would just mean Graphsync had to do a bit of extra work to actually look up the root node on the responder side.

The advantage is this allows #116, which would make selectors fully recursive.

### DIFF
--- a/block-layer/graphsync/graphsync.md
+++ b/block-layer/graphsync/graphsync.md
@@ -64,10 +64,11 @@ message GraphsyncMessage {
 
   message Request {
     int32 id = 1;       // unique id set on the requester side
-    bytes selector = 2; // ipld selector to retrieve
-    bytes extra = 3;    // aux information. useful for other protocols
-    int32 priority = 4;	// the priority (normalized). default to 1
-    bool  cancel = 5;   // whether this cancels a request
+    bytes root = 2;     // a CID for the root node in the query
+    bytes selector = 3; // ipld selector to retrieve
+    bytes extra = 4;    // aux information. useful for other protocols
+    int32 priority = 5;	// the priority (normalized). default to 1
+    bool  cancel = 6;   // whether this cancels a request
   }
 
   message Response {


graphsync/graphsync.md
```diff
@@ -62,10 +62,11 @@ message GraphsyncMessage {
 
   message Request {
     int32 id = 1;       // unique id set on the requester side
-    bytes selector = 2; // ipld selector to retrieve
-    bytes extra = 3;    // aux information. useful for other protocols
-    int32 priority = 4;	// the priority (normalized). default to 1
-    bool  cancel = 5;   // whether this cancels a request
+    bytes root = 2;     // a CID for the root node in the query
```
Stebalien: Nit: could we call this `root_cid`? I'd kind of like to leave room for `root_ipns` etc.

Stebalien (approved):

graphsync/graphsync.md
hannahhoward: It's actually an encoded IPLD Link -- I'm curious @warpfork would IPNS eventually be a type of IPLD Link? Probably not but maybe? Just curious.

hannahhoward: IPLD Link = abstraction on CID that only has one implementation (CID) right now, but could have others in the future maybe?

Stebalien: I assume "IPLD Link" = CID + Path, right? Is there a document for this? I.e., how are we encoding the type?

warpfork: 0. If we don't know, speculative generality is bad.

1. 'Link' is the name of a piece of abstraction in go-ipld-prime that's mostly there to reduce the import graph tangle (go-cid pulls in ~large amounts of dependencies and it's something I want to have a Conversation about someday)

2. In theory, Link could be anything that we can make the LinkLoader interface do.  In practice, it's a CID.  I can't honestly say I've given many oodles of brainseconds to thinking about what would and would not be a _wise_ idea to do with this extensibility, so without that, I'd tend towards caution.

warpfork: I'd be fine with `root_cid` because that seems clear regardless, but I'm pretty indifferent.  This is protobuf... it's also valid to call it `root` and if there's more fields later, rename it... renaming the field to be less ambiguous later is 'free' in protobuf...

warpfork (approved):

graphsync/graphsync.md
Stebalien: > If we don't know, speculative generality is bad.

But leaving room for future improvements is kind of our motto.

> 'Link' is the name of a piece of abstraction in go-ipld-prime that's mostly there to reduce the import graph tangle (go-cid pulls in ~large amounts of dependencies and it's something I want to have a Conversation about someday)

go-cid pulls in multihash which is kind of required. Forcing users to use an entirely different type to use go-ipld-prime sounds like a bad idea.

If Link is just a fancy term for CID, lets just use CIDs. Having two things for the same thing is _really_ confusing.

> I'd be fine with root_cid because that seems clear regardless, but I'm pretty indifferent. This is protobuf... it's also valid to call it root and if there's more fields later, rename it... renaming the field to be less ambiguous later is 'free' in protobuf...

Yeah, I just want to make it clear that we may choose a different root later.

hannahhoward: Merging as is for the time being, seems like a larger discussion needs to be had about the concept of IPLD Links, and whether they exist as something other than just a CID outside the go-ipld-prime code.

hannahhoward: https://github.com/ipld/specs/issues/145

PR #121, opened 2019-05-06
Simplified Layer Model

mikeal:
I’ve been thinking about this for quite a while.

Our existing layer model is too complicated. It sort of makes sense to those of us deep in the stack but there’s too many caveats and people don’t seem to understand it.

With this more simplified model I hope to make it clearer *where* specific functionality belongs. It should answer these basic questions:

* When do I need a new codec and when do I **not** need a new codec?
* What can I do without schemas?
* What can I **only** do effectively with schemas?

As we implement more path and selector libraries I’d also like a clearer way to talk about what layers that library operates at.

* Does this path library understand HAMTs and other collections?
* Is this selector engine just looking at properties from the Data Model or does it understand more advanced layouts described in schemas?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,77 +1,64 @@
 IPLD Specifications
 ===================
 
-IPLD is not a single specification, it is a set of specifications.
+IPLD is not a single specification it is a set of specifications.
 
-```
-                                     The IPLD Stack
-
-           ┌───────────┐ ┌────────────────────────────────────┐
-           │           │ │                                    │
-           │    MFS    │ │       End-User Applications        │
-           │           │ │                                    │
-           ├───────────┤ ├─────────────────────────┬──────────┤            ┌───────────┐
-           │           │ │    Specialized Data     │          │            │           │
-           │ UnixFS v2 │ │ Structures & Utilities: │          │            │    MFS    │
-           │           │ │   VR, Geo, SQL, etc.    │          │            │           │
-           ├───────────┴─┴─────────────────────────┘          │            ├───────────┤
-           │                                                  │            │           │
-           │    Multi-block Collections: Maps, Sets, etc.     │            │ UnixFS v1 │
-           │                                                  │            │           │
-┌──────────┼──────────────────────────────────────────────────┼────────────┼───────────┤
-│          │                                                  │            │           │
-│          │            dag-json         dag-cbor             │  ipld-git  │           │
-│          │                                                  │            │           │
-│  Codecs  ├──────────────────────────────────────────────────┤  ipld-btc  │   dag-pb  │
-│          │                                                  │            │           │
-│          │                 IPLD Data Model                  │ ipld-zcash │           │
-│          │                                                  │            │           │
-└──────────┼──────────────────────────────────────────────────┴────────────┴───────────┤
-           │                                                                           │
-           │                           Block (CID, Raw Data)                           │
-           │                                                                           │
-           └───────────────────────────────────────────────────────────────────────────┘
-```
-
-The goal of this stack is to enable decentralized data-structures
+The goal of IPLD is to enable decentralized data-structures
 which in turn will enable more decentralized applications.
 
-Many of the specifications in this stack are inter-dependent.
+Many of the specifications in IPLD are inter-dependent.
+
+# IPLD Layer Model
+
+This layer model is a simplified heirarchy of IPLD concepts and requirements.
 
 ```
-                          IPLD Dependency Graph
-
-┌─────┐                           ┌────────────┐
-│ CID ├───────────┬───────────────> Raw Blocks │
-└─────┘           │               └─┬──────────┘
-          ┌───────v──────────────┐  │
-┌──────┐  │ Links (Conceptually) │  │
-│ Path │  └───────┬──────────────┘  │           ┌─────────────┐
-└──┬───┘          │                 ├───────────> Replication │
-   │              │                 │           └─────────────┘
-┌──v──────────────v─────────────────┴──────┐
-│                  Codecs                  │
-│  ┌─────┐   ┌───────────────────────────┐ │ ┌────────────────────────────┐
-│  │ Git │   │       Data Model v1       │ │ │  Multi-block Collections   │
-│  └─────┘   │ ┌──────────┐ ┌──────────┐ │ │ │ ┌────────────────────────┐ │
-│ ┌────────┐ │ │ dag-json │ │ dag-cbor │ ├─┼─> │    Specialized Data    │ │
-│ │  BTC   │ │ └──────────┘ └──────────┘ │ │ │ │ Structures & Utilities │ │
-│ └────────┘ └───────────────────────────┘ │ │ └────────────────────────┘ │
-│ ┌────────┐ ┌────────┐                    │ └───┬─────────┬──────────────┘
-│ │ Zcash  │ │ dag-pb │                    │     │         │
-│ └────────┘ └────────┘                    │     │         │
-└──────────────────────────────────────────┘     │         │
-                                                 │         │
-               ┌──────────────────────┐          │         │
-               │ File System (UnixFS) <──────────┘         │
-               └──────────────────────┘                    │
-               ┌───────────────────────┐                   │
-               │ End-User Applications <───────────────────┘
-               └───────────────────────┘
-
+┌──────────────────────────────────────────────────┐
+|                                                  | 
+|                   Schema Layer                   |
+|  (advanced types, multi-block data structures)   |
+|                                                  |
+├──────────────────────────────────────────────────┤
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,77 +1,64 @@
...
+└──────────────────────────────────────────────────┘
 ```
 
-The goal of this stack is to enable decentralized data-structures
-which in turn will enable more decentralized applications.
+## Block Layer (Layer 0)
 
-Many of the specifications in this stack are inter-dependent.
+The block layer encompasses all content addressed block formats and specifies how blocks
+are addressed,how they self-describe their codec for encoding/decoding, and how block link
```
warpfork: s/block/blocks/

README.md
```diff
@@ -1,77 +1,64 @@
...
+
+The Data Model Layer describes a set of base required types to be implemented by a subset of IPLD codecs.
+
+With these basic types authors can create various single-block data structures which can be read with predictable
+paths and selectors.
+
+With just the data model layer, several data structures can be authored and put into a single block. These data 
+structures can also link to one another, but a *single* data structures's namespace cannot be spread across many
+blocks. This limitation means that collections authored in only the Data Model layer are limited by the maximum
+block size (typically 2 megabytes).
```
warpfork: I'd split this mention of block size limits into its own paragraph, and try to phrase it as *some* transports having block size limits.

Perhaps some text to the effect that "ideally, a system designs to produce blocks of limited size: since a block is the unit of data that must be loaded into memory at once for any operation, a good size is on the order of megabytes.  larger datastructures should be sharded (and advanced data layout features such as HAMT, etc, can be used do this transparently)."   And then we can mention the example of libp2p's practical size limits as an instance of why this is good to mind.

warpfork (commented):
Definitely simpler!

README.md
```diff
@@ -1,77 +1,64 @@
...
-│ └────────┘ └────────┘                    │     │         │
-└──────────────────────────────────────────┘     │         │
-                                                 │         │
-               ┌──────────────────────┐          │         │
-               │ File System (UnixFS) <──────────┘         │
-               └──────────────────────┘                    │
-               ┌───────────────────────┐                   │
-               │ End-User Applications <───────────────────┘
-               └───────────────────────┘
+This layer alone is enough to accomplish basic graph replication across a variety of formats (eth, bitcoin, git, etc).
```
vmx: I'm unsure if we should mention formats here. As those certainly live in Layer 1. They just don't implement the full IPLD Data Model (so not everything IPLD can be encoded in them), but all types they have should be decoded in an IPLD Data Model compatible way.

README.md
```diff
@@ -1,77 +1,64 @@
...
 
-```
+This layer does not define data structures or types, although many codecs may convert these formats into native types,
+there are no type requirements or assurances about types at the block layer.
+
+## Data Model Layer (Layer 1)
+
+The Data Model Layer describes a set of base required types to be implemented by a subset of IPLD codecs.
```
vmx, line 48: I don't understand the "by a subset of IPLD codecs" part.

README.md
```diff
@@ -1,77 +1,64 @@
...
+
+With these basic types authors can create various single-block data structures which can be read with predictable
+paths and selectors.
+
+With just the data model layer, several data structures can be authored and put into a single block. These data 
+structures can also link to one another, but a *single* data structures's namespace cannot be spread across many
```
vmx: I think "data structure namespace" needs a definition.

mikeal: ya, i’m struggling with the language here because we’ve actually just over-loaded the term “data structure.”

like, we *do* have multi-block data structures we just don’t have multi-block **collections**. maybe I just need to replace some of the data structure verbiage with collections and link appropriately.

README.md
mikeal: What I mean is, “not all IPLD codecs support the full data model.”

README.md
mikeal: maybe we should just add mention of dag-pb and dag-cbor to make it clear that you can do replication with only this layer.

README.md
warpfork: I've been flirting with the idea of defining it as:

- "IPLD native formats" can represent any tree of data in the IPLD Data Model.  This basically describes dag-cbor.
- "IPLD adjunct formats" can be mapped onto the IPLD Data Model, and we can do some operations on them through that lens, but they *can't* support *any arbitrary* data from the IPLD Data Model being shoveled into them.  And this describes git, eth, etc.

(or s/formats/codecs/, I haven't been paying attention using those words very consistently myself)

README.md
```diff
@@ -1,77 +1,64 @@
...
-│          │                                                  │            │           │
-│          │                 IPLD Data Model                  │ ipld-zcash │           │
-│          │                                                  │            │           │
-└──────────┼──────────────────────────────────────────────────┴────────────┴───────────┤
-           │                                                                           │
-           │                           Block (CID, Raw Data)                           │
-           │                                                                           │
-           └───────────────────────────────────────────────────────────────────────────┘
+┌──────────────────────────────────────────────────┐
+|                                                  |
```
rvagg: use the ascii box `│` instead of `|`, it'll look betterer

README.md
```diff
@@ -1,77 +1,64 @@
...
-           └───────────────────────────────────────────────────────────────────────────┘
+┌──────────────────────────────────────────────────┐
+|                                                  | 
+|                   Schema Layer                   |
+|  (advanced types, multi-block data structures)   |
+|                                                  |
+├──────────────────────────────────────────────────┤
+|                                                  |
+|                 Data Model Layer                 |
+|   (basic types, single-block data structures)    |
```
rvagg: I like this explicit distinction: multi-block, single-block, we should be sure to emphasize that

README.md
rvagg: `s/structures's/structures'/`

rvagg:
Starter for schema layer:

```
## Schema Layer (Layer 2)

IPLD Schemas define a mapping from the Data Model Layer (Layer 1) to
instantiated data structures comprising complex layouts. The Schema Layer
adds the ability to extend the IPLD Data Model to the wide variety of types
required for typical programatic interaction with a data source without the
need to implement custom translation abstractions.

The Schema Layer will also serve as an enabling layer for complex multi-block
data structures by providing stability and consistency of data model use within
individual blocks and defined interaction points for the logic required for
building and interacting with advanced data layouts, such as multi-block Maps,
Lists and Sets.
```

Although, I'm not convinced with the inclusion of everything above the single-block data model into a single "Schema Layer", there seems like there's at least two different things in there: schemas for mapping to more complex data layouts, and multi-block data structures—with or without sophisticated logic that can't fit directly into the data model or schemas.

mikeal:
> Although, I'm not convinced with the inclusion of everything above the single-block data model into a single "Schema Layer"

Ya, @warpfork and I were having this discussion in a side channel. If I had a better name for “Schema Layer” I’d use it, because I think we’re already putting more than just “schemas” in it. However, I’d like to *try* and keep all of our “above the data model” work in a single place, hopefully in a single reserved property, even if we end up doing something other than just schema’s.

README.md
mikeal: i tweaked the wording a bit to make it even more explicit.

README.md
mikeal: I like “native” but I’m not a huge fan of “adjunct.”

But I’m wondering if we can get rid of this entire distinction. It’s the role of a codec to **present its internal format in IPLD Data Model kinds**. Even if it doesn’t support the entire Data Model the codec needs to conform to it.

Then, the distinction for a **native** codec is one that “can encode/decode data in all IPLD Data Model kinds.”

vmx:
I miss in the current "schema layer" section the simple thing of just describing the data with a schema (without talking about complex data structures).

mikeal:
I’m ready to merge this “as-is” and iterate from there. Any objections?

> I miss in the current "schema layer" section the simple thing of just describing the data with a schema (without talking about complex data structures).

Even that use of schemas is built “on top of” the Data Model, so it fits just fine in that layer. The way I try to look at it is that the purpose of each layer is just to enable the layer above it.

At some point, we may want a more detailed diagram of just the Schema Layer, but doing that now seems a little pre-mature since we’re still figuring a lot of this out.

rvagg:
> At some point, we may want a more detailed diagram of just the Schema Layer, but doing that now seems a little pre-mature since we’re still figuring a lot of this out.

My bet is on a different name emerging or it being split into two when we explore it deeper. But it's good as is so 👍 for merging.

README.md
warpfork: Intensely thumbs up to that text in bold.

(And I'm not particularly fond of 'adjunct' as a term either.)

warpfork (approved):

PR #122, opened 2019-05-13
[WIP] Initial work on contributing guidelines for specifications

mikeal:


CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
...
+  - Draft
+  - Final
+
+Prescriptive specifications are intended to describe future 
+implementations or, in some cases, changes to existing implementations.
+
+Descriptive specifications describe existing behavior. In many cases
+these specifications are not intended to drive new implementations and
+are only written to understand existing behaviors.
+
```
rvagg, line 19: I suggest we also say how to identify the stage of a spec, something like:

> Specification documents should include a descriptor in their heading that indicates their status in the staging process. e.g. "Prescriptive - Draft", or "Descriptive - Final".

Then we can go about amending what's in here.

rvagg (changes requested):

rvagg:
#123 introduces non-spec-but-related content, the README contains non-spec-but-related material, https://github.com/ipld/specs/blob/master/data-structures/multiblock-collections.md isn't strictly a "spec". We need to be more clear about what a "spec" is and how to identify one. Do we use directories to make it work? Do we identify them clearly in their title and/or filenames? How do the merge rules apply to things that aren't strictly specs?

mikeal:
If there are no objections I’ll merge this tomorrow.

mikeal:
@rvagg What should we do with this now? Was it all merged as part of the other PR or are there parts of it that weren’t included?

rvagg:
yes, all included along with some extra words, including notes about non-spec material (like "concepts" and README)

PR #123, opened 2019-05-26
Cardinality concept intro

warpfork:
Add 'concepts' folder; Introduce cardinality.

This is still pretty brief, but it's been mentioned elsewhere that since this is a recurring theme, we should introduce it somewhere on its own rather than just in passing in various places.

I'm guessing we'll end up with some other things to populate this 'concepts' folder later as well.

### DIFF
--- a/concepts/cardinality.md
+++ b/concepts/cardinality.md
@@ -1,0 +1,17 @@
+Cardinality
+-----------
+
+"Cardinality" means "[the number of elements in a set](https://en.wikipedia.org/wiki/Cardinality)".
+
+In IPLD, we use the phrase "the cardinality of a type" to define how many members a type can have.
+
+Cardinality is important to the way we define and think about our Data Model and Schema type systems,
+because if the cardinality of two parts of a model aren't the same, then that means one of them is less expressive,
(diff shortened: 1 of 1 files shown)


vmx (approved):
Nice. This reads good and clear to me.

rvagg:
The content is fine, but ... Other concepts are addressed in the README already, so maybe they should be pulled out into the "concepts" directory, or maybe this should go into the README? If it ends up in files other than the README then somewhere needs to spell out that what's in the "concepts" directory is not "spec" material. Also, how do the merge rules in #122 apply to concepts (and the README too I guess?).

warpfork:
I think I'll merge this so the content gets on master, and then agree we should do another pass on the readmes and such to see of other things might belong in this dir (or this content should settle elsewhere and we can drop this dir again).

I'd propose a rule of thumb that we put stuff that's underlying design reasoning and concepts rather than particulars of design; meta-rationale, if you will.

rvagg:
@warpfork we have a new directory layout now and "concepts" are alongside specs now but are also clearly labelled as such. See https://github.com/ipld/specs/pull/128 for a rearrangement of this to fit into the new layout.

PR #124, opened 2019-05-26
Schema overview page, and readme TOC links.

warpfork:
Riffing somewhat heavily from what's currently incubating in the other
repo at https://github.com/ipld/go-ipld-prime/blob/master/doc/schema.md

Further iteration on this will be super welcome; it's a bit abrupt.

### DIFF
--- a/schema/overview.md
+++ b/schema/overview.md
@@ -1,0 +1,68 @@
+IPLD Schemas Overview
+=====================
+
+
+Goals
+-----
+
+1. Provide a reasonable, easy to use, general type system for declaring
+  useful properties about data.
+2. Compose nicely over IPLD Data Model.  Schemas should be usable with any
+  data format you can build an IPLD Codec for.
+3. Be efficient to verify (e.g. predictable; roughly linear in the size of
+  the data and schema; and absolutely not turing complete).
+4. Be language-agnostic: many compatible implementations of the schema
+  checker tools should exist, as well as bindings for every language.
+5. Assist rather than obstruct migration.  (We expect data to exist from
+  *before* the current schemas; we need to work well on this case.)
+6. Be a home for collaboration and documentation!  Schemas should provide a
+  natural place for API documentation, and be a reasonable design literature.
+
+In total: **Make developing decentralized & distributed applications
+easier, clearer, and more robust!**
+
+
+Features
+--------
+
+Well-understood basics (sum types, product types, some specific recursive types):
+
+- typed maps (e.g. `{Foo:Bar}`)
+- typed arrays (e.g. `[Foo]`)
+- typedef'd primitives (e.g. `type Foo int`)
+- structs (e.g. `struct{ f Foo; b Bar }`)
+- discriminated unions (several styles)
+- enumerations (e.g. `["Foo" | "Bar" | "Baz"]`)
+
+Bonus features:
+
+- simple syntax for 'maybe' (e.g. `struct{ f nullable Foo }` or `[nullable Foo]`)
+- clear syntax for non-required fields (e.g. `struct { f optional Foo }`)
+
+All of this maps over the IPLD Data Model:
+
+- every type has a representation;
+- e.g. maps are represented as maps; structs are also represented as maps!
+- representations can specify aliases for how to serialize things; e.g. provide
+  shorter serial strings to use in place of longer struct names.
+- representations are customizable, including mapping a type onto a different
+  representation kind entirely!  e.g. a struct can be declared to have a
+  list representation (making it shorter to serialize, if less self-describing);
+  or, a struct can be represented by simple string patterns (which is
+  particularly neat since that string can then be used in map keys!).
+
+Some things can go even further than the basic representations:
+
+- "advanced layouts": types can be declared to act as already-familiar kinds
+  of data structures such as maps or lists or byte arrays, while using
+  a (fully pluggable) system to transform the content into other forms for
+  representation, *including* splitting the data into fully separate chunks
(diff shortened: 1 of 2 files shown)


vmx (approved):
Sweet!

schema/overview.md
```diff
@@ -0,0 +1,68 @@
+IPLD Schemas Overview
+=====================
+
+
+Goals
+-----
+
+1. Provide a reasonable, easy to use, general type system for declaring
```
rvagg, line 8: the relationship between "type system" and "schemas" seems to be in debate at the moment, we need to have a clear resolution of these two ideas before this lands

schema/overview.md
```diff
@@ -0,0 +1,68 @@
...
+easier, clearer, and more robust!**
+
+
+Features
+--------
+
+Well-understood basics (sum types, product types, some specific recursive types):
+
+- typed maps (e.g. `{Foo:Bar}`)
+- typed arrays (e.g. `[Foo]`)
```
rvagg, line 31: we've so far avoided using "array" and stuck to "list" I think, maybe it'd be safer to keep it that way, or at least put "list" primary, either "typed lists" or "typed lists / arrays"

schema/overview.md
```diff
@@ -0,0 +1,68 @@
...
+- typedef'd primitives (e.g. `type Foo int`)
```
rvagg, line 32: should we reinforce "kinds" here too? "primitives / kinds" or  "primitives / IPLD Data Model kinds"

schema/overview.md
```diff
@@ -0,0 +1,68 @@
...
+
+- simple syntax for 'maybe' (e.g. `struct{ f nullable Foo }` or `[nullable Foo]`)
+- clear syntax for non-required fields (e.g. `struct { f optional Foo }`)
+
+All of this maps over the IPLD Data Model:
+
+- every type has a representation;
+- e.g. maps are represented as maps; structs are also represented as maps!
+- representations can specify aliases for how to serialize things; e.g. provide
+  shorter serial strings to use in place of longer struct names.
```
rvagg, line 47: ahh, "to use in place", a question I've been meaning to raise about aliases is: does a serialiser use the primary or the alias or is it left up to the implementation because it doesn't matter? is the purpose of an alias to allow the _deserialising_ of data with an alternative form or to allow the _serialising_ of data in an alternate (maybe more efficient?) form, or some undefined combination of both? herein lies both the consistent-hash problem and the consistent-across-implementations problem.

doesn't need to be addressed here, just noting for future discussion

schema/overview.md
```diff
@@ -0,0 +1,68 @@
...
+- representations are customizable, including mapping a type onto a different
+  representation kind entirely!  e.g. a struct can be declared to have a
```
rvagg, line 49: you're using "kind" here so I think that suggests you should use it with or instead of "primitive" above (although I quite like using it _with_ "primitive" as it helps explain "kind" which is bound to be unfamiliar to some audience)

schema/overview.md
```diff
@@ -0,0 +1,68 @@
...
+  list representation (making it shorter to serialize, if less self-describing);
+  or, a struct can be represented by simple string patterns (which is
+  particularly neat since that string can then be used in map keys!).
+
+Some things can go even further than the basic representations:
+
+- "advanced layouts": types can be declared to act as already-familiar kinds
```
rvagg, line 56: in debate/flux, needs to be resolved before this can land

schema/overview.md
warpfork: **must** use alias.  consistency problem is right -- both for hashing, and also for cardinality correctness even aside from hashing.

schema/overview.md
warpfork: I've been using the two more or less interchangeably.  "IPLD Schemas" are the definite noun (we're *not* calling this the proper-noun "IPLD type system")... but they're a "type system" of some general sort.

I think we hit a "limits of language" thing somewhere in this area, tbh.  Lots of this system's designs are coming from "type theory" (for programming languages) -- so the term seems to apply -- even though we're explicitly bending and limiting it to working over data and *not* function signatures (which drastically simplifies and slightly pivots the domain) -- and that puts us in a space that's... not... super well explored or well defined, relatively speaking, in popular literature.

So I'm generally +1 to trying to clarify that, but with the caveat of I think we'll absolutely land in "what are words" territory at some point and have to call it good enough.

schema/overview.md
warpfork: ... so, 'alias' might not be an ideal name for this feature, since it gives the association of multiple validity :thinking:

schema/overview.md
warpfork: hrm on second thought i'm not sure about this one, it feels a little clunky to me.  This is the overview page; do we wanna start using terms like that so early?  In subsequent pages, yes, but here I'm unsure.

maps and ~arrays~ lists are also kinds.  Maybe modifying the sentence prefacing this block of bulletpoints to say "[...] named types can be defined as any of the essential kinds"?

rvagg: maybe, or maybe you could make the assumption that they've read the "data model" page(s) as this is now 2 layers up from there
your call, not a big deal

schema/overview.md
rvagg: well the name "alias" made me assume it was an alternate form valid during deserialisation and that serialisation would always use the canonical name that's in the parent `type` section. So if it's a "**must** use alias", then yes, I think it's the wrong term. I don't have a suggestion for an alternative just at the moment though.

PR #125, opened 2019-05-28
Major reorg and introduction of spec status & governance

rvagg:
WIP I suppose; I don't imagine this will be straightforward.

This is probably best assessed by viewing through the branch rather than trying to make sense of some of the diffs (especially README): https://github.com/ipld/specs/tree/rvagg/big-reorg

Inspired in part by the governance proposal at #122 (included here, but in the README) which focuses on specifications, but we have more than just specifications as #123 makes clear (as does a lot of the existing content).

- Divided into one directory per layer (sort of aligns with @warpfork's proposal from our Berlin team week but we've lost the specifics of that proposal, and I don't remember it, so it's not the same).
- Documents divided into "Specification" or "Concept"
- README trimmed of old concept/glossary content--moved to individual files or within existing files where there's overlap
- Added Specification status descriptors in README and each spec doc
- Added governance from #122
- Removed ancient REQUIREMENTS.md (I couldn't find anything in here that was useful and wasn't already restated in some new way, made redundant by new work or was included in WIP content)

Edits to README look bigger than they are because I went with a more consistent line-break strategy than we have now of break-on-sentence.


mikeal:
I’m planning on writing a bunch of specs today and I want this merged beforehand so I’m clicking the button :)

rvagg:
👌 brilliant, now I need to get more content in!

PR #126, opened 2019-06-05
[WIP] IPLD Composites (formerly IPLD Generics (formerly IPLD multi-block types))

mikeal:


schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,79 @@
+# Specification: IPLD Generics
```
mikeal: Once I started writing this up the name “multi-block types” didn’t make a lot of sense. A fair amount of what we’re doing is actually contained in a single block. Also, once I started to back out the `kind` inheritance hierarchy it was much more obviously a system for generics so I went with this naming, but the name is still up for debate so if people hate it let me know.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,79 @@
...
+
+While the `_type` property is reserved in every map regardless of where it appears in a block, IPLD Generics **MUST** include the value of `_type` in the root property of a discrete block. Note: you can still embed multiple IPLD Generics in a single block using [inline blocks]().
+
+### Version 1
+
+The `_type` property is a string identifier. This identifier is used to lookup the implementation and if it cannot be found by the host environment any operation is expected to throw an exception. 
+
+Implementations MUST NOT fallback to *Layer 1* operations of the contents of the node.
+
+### Version 2
```
mikeal, line 66: @vmx let me know if this makes sense for the WASM stuff.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,79 @@
...
+* Flexible multi-block binary types.
+
+## Requirements
+
+IPLD Generics cannot be implemented without:
+
+  * The IPLD Data-Model. There are codec agnostic but do require that the full data model be implemented by the codec.
+  * Reserved `_type` property.
+
+While the `_type` property is reserved in every map regardless of where it appears in a block, IPLD Generics **MUST** include the value of `_type` in the root property of a discrete block. Note: you can still embed multiple IPLD Generics in a single block using [inline blocks]().
```
mikeal: The motivation for this section is a bit complicated, but it comes down to making the implementation easier to reason about, consistent, and performant. After a block is decoded and used by the engine you completely lose track of whether or not it is at the root, so you need the property reserved at every layer. 

However, the individual generic implementations may traverse into the node data on their own and could easily skip over a `_type` property and there’s no good way to keep them from doing so unless we don’t send them the decoded node data and require them to call back for read operations for **everything** which would be prohibitively expensive. So, the only place we can **ensure** that decoded node data is wrapped in the generic implementation is at the root of the block because that’s something the generic implementations aren’t even allowed to do and have to call back into the engine to accomplish.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
+# Specification: IPLD Generics
+
+**Status: Prescriptive - Exploratory**
+
+Organizing IPLD data into usable, efficient, complex data structures spanning many blocks aimed for use by end-user applications.
+
+This document will re-use some terms found in the [IPLD data model](/IPLD-Data-Model-v1.md).
```
rvagg: `[IPLD data model](/data-model-layer/data-model.md)`

schema-layer/data-structures/generics.md
rvagg: Might be good to include some of this in that paragraph.

Although I'm not getting the nuance here that explains the contradiction between: "you lose track of whether or not it is at the root" and "the only place we can ensure that decoded node data is wrapped in the generic implementation is at the root of the block". It seems like the implementation difficulties are dictating that the root is treated as a special case anyway.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Version 2
+
+The `_type` property is a `Map`.
+
+The map must contain the following properties.
+
+  * `name` must be a string identifier.
+  * `engine` must be one of the following:
+	* `”IPLD/Engine/WASM/0”`
```
rvagg: fix indenting, is this meant to be a second-tier list with one item

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+Each additional property describes the implementation of every operation.
+
+*TODO: define structure of pointers to WASM functions*
```
rvagg, line 78: this isn't a blocker to merging is it?

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Operations
+
+This section describes only the operations for which there is currently an implementation. For a more exhaustive list of operations we may support in the future read [IPLD Multi-block Collections](https://github.com/ipld/specs/blob/master/schema-layer/data-structures/multiblock-collections.md).
```
rvagg: remove `https://github.com/ipld/specs/blob/master` from link

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+The term “leaf” is used often below. When an operation is performed on a node it may make any number of additional calls to operations on other data structures with many results being produced and passed into continuations. “leaf” responses are those produced by an engine that are from the original target operation and not from subsequent operations trigger by the original operation.
```
rvagg: s/trigger/triggered

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+### GET `{ path }`
+
+Used for property lookups. Enables functionality similar to what you would expect from `Map` and `List` kinds.
+
+Only a single leaf `{ result: Value }` must be received.
+
+### KEYS `{}`
+
+Use to produce an iterator of every top level property available in the data structure. Takes no options.
```
rvagg: s/Use/Used

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+This section describes only the operations for which there is currently an implementation. For a more exhaustive list of operations we may support in the future read [IPLD Multi-block Collections](https://github.com/ipld/specs/blob/master/schema-layer/data-structures/multiblock-collections.md).
+
+The term “leaf” is used often below. When an operation is performed on a node it may make any number of additional calls to operations on other data structures with many results being produced and passed into continuations. “leaf” responses are those produced by an engine that are from the original target operation and not from subsequent operations trigger by the original operation.
+
+### GET `{ path }`
+
+Used for property lookups. Enables functionality similar to what you would expect from `Map` and `List` kinds.
+
+Only a single leaf `{ result: Value }` must be received.
```
rvagg: is `{ result }` too implementation-specific? what would this mean in a Go context? and what about `GET { path }`, is the `{}` meant to indicate arguments here? named arguments? is this a `{ path: ... }`? Again, too implementation-specific?

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+While each implementation has a language specific API there is a fair amount in common between implementations that is specified here.
+
+The interface for every operation can do the following:
+
+  * Return a result.
+  * Perform multiple calls of **any** operation on multiple targets of `value` or `path`
+	  *  `value` is an IPLD Data Model value.
+	  *  `path` is a IPLD Path relative to the node performing the operation.
+		  *  This path should be interpreted as *Layer 1* within the node and *Layer 2* once the path leaves the current block.
```
rvagg: bad indenting

rvagg: I think you might need to expand on what this means, it's not even clear to me. It seems to me that "Layer 1" and "Layer 2" are the wrong way around here since "within the node", assuming a Typed node, it's got to be interpreted in some special way. Although for a `GET` is there another way to interpret a path than pull off the first segment of a `'/'` separated string and do something non-layer-1 with it? Does that need to be made clear?

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+      * The result of these operations should be passed back into the original operation as a continuation **or** “proxied” as the leaf result of the original operation.
```
rvagg: you need to expand in here somewhere what "proxied" means, something like: "A proxied operation instructs the implementation to re-perform the same operation on some other block as supplied by the type"?

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+	  *  `value` is an IPLD Data Model value.
+	  *  `path` is a IPLD Path relative to the node performing the operation.
+		  *  This path should be interpreted as *Layer 1* within the node and *Layer 2* once the path leaves the current block.
+      * The result of these operations should be passed back into the original operation as a continuation **or** “proxied” as the leaf result of the original operation. 
+
+### Languages
+
+  * JavaScript
+	  * [Spec]()
+		* [Code]()
```
rvagg: indenting, and link is good to go now isn't it? and what will a JS "spec" look like? we're doing language-specific specs now?

schema-layer/data-structures/generics.md
rvagg: `[inline blocks]()` needs a target or to not be a link (maybe add a TODO in brackets?)

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
+# Specification: IPLD Generics
+
+**Status: Prescriptive - Exploratory**
+
+Organizing IPLD data into usable, efficient, complex data structures spanning many blocks aimed for use by end-user applications.
+
+This document will re-use some terms found in the [IPLD data model](/IPLD-Data-Model-v1.md). 
+
+IPLD Generics offer codec agnostic programming interfaces for all common operations users can currently accomplish on Data-Model [“Kinds”]().
```
rvagg: `["Kinds"](/data-model-layer/data-model.md#kinds)`

rvagg (commented):
really needs @warpfork to help us find a meeting point between schemas and types, or at least to agree that there is some place where they intersect and our various visions for these things is mutually understood and sane

schema-layer/data-structures/generics.md
mikeal: To clarify, the *engine* loses track of whether the “node” is at the root of a block or not. This *would* enable generics in any Map except for the fact that all of the decoded node data is passed into the generic nodes, which means they are free to work through several maps and we can’t guarantee they are checking for `_type` unless we force them to call back into the engine for a read anywhere in the block data :/

schema-layer/data-structures/generics.md
mikeal: nope, it’s still an exploratory spec so we can merge without it being complete

schema-layer/data-structures/generics.md
mikeal: I’ll document this better, but you need to be able to pass in “named parameters” and return “named parameters.” This spec won’t define *how* you implement named parameters, or continuations for that matter.

schema-layer/data-structures/generics.md
mikeal: Ya, we don’t have a good definition of “path” in terms of layers, but I also didn’t have a better way to describe it.

Basically, you need to read into the node value (decoded block data) until you hit a link, then you need to run the rest of the path through the generics engine. You can’t use the generic implementation for the read because it could be a `get` method itself that asked for the read.

Note: this is another reason generic root nodes must be discrete blocks.

schema-layer/data-structures/generics.md
mikeal: maybe change to “docs?” agree that we don’t necessarily need a spec for each implementation.

rvagg: Maybe those docs just go into the repo where it lives so the repo would cover both. This may span multiple repos so this could just be a list of repos.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+  * They are lacking self-description. A consumer of a graph containing these structures would have to have logic on top of IPLD and vary the way it performs operations on there data structures.
+  * Implementations of these data structures cannot perform operations on each other. In other words, multi-block data structures have a hard time building on top of each other.
+
+Since there wasn’t a standardized way to describe these data structures we couldn’t build libraries for paths and selectors that seamlessly supported them.
+
+As we started designing this system several other requirements surfaced.
+
+* Transparent encryption envelopes on the read and write side.
+* Advanced `Link` types that can support some form of mutability and link to paths within other data structures.
```
warpfork, line 32: Not clear to me that this will end up being the same thing as the work we do to support mulitblock collections and transparent encryption.  I'd leave it out as a motivation here (although it surely is something that's been wished for).

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+* Flexible multi-block binary types.
+
+## Requirements
+
+IPLD Generics cannot be implemented without:
+
+  * The IPLD Data-Model. There are codec agnostic but do require that the full data model be implemented by the codec.
+  * Reserved `_type` property.
```
warpfork: We talked about this irl the other day, but for internet posterity: I don't actually think this is a strict requirement; and, it has huge drawbacks.

There are at least three different ways I can think of to solve the signalling problem:

1. when writing code against the libraries, simply Do Things as intended; signaling isn't necessary at all.  (This is defacto what everything previously has done, and we'll never *stop* it from being doable; though we may want to also offer *better* alternatives that are more communicable rather than implementation specific.)

2. Use schemas as a signalling mechanism -- some description at http://slides.com/warpfork/ipld-gpn-2019#/9/24 as well as our other various gists and drafts in recent weeks.  This has many boons:
  - it's not in-band signalling (not blue-box prone; requires no reserved words and causes no asciifucking or escaping requirements)...!
  - it provides documentation of what's going on (remember, one of the goals of the schema system as a whole is to make it possible to describe systems well from a high level; such descriptions would be woefully incomplete if they *didn't* include some definition of any sharding going on, so we **will** have this information there no matter what)
  - **it makes it possible to configure the libraries to produce *new data* without having do that configuration in an library/implementation-specific way.**  This is super important: if you're writing new data, where's the "_type" keyword going to come from?  It doesn't exists; there's no in-band signalling data at all when you're writing new data, so we need *something* here or we're just not feature complete! (We could try to separate out the concept of declarative formats for stating what a sharding or encrypting etc feature is going to do from schemas, and then just embed that in schemas again, if that makes this easier to think about.)
  - it's this concept of indicating the advanced layout strategies (or, not!) from the outside that enables features like choosing whether to path through things transparently at the high level or as easily address lower level content blocks: http://slides.com/warpfork/ipld-gpn-2019#/9/25 -- and remember, we need this to be serializable so that we can attach it to things like selectors sent over the wire.  (It's at this point that schemas become relevant again even if in the last bullet point, we separated out the declarative sharing config from schemas: to define *where* we're going to use a declarative sharding config, we need some recursive definition language, right?)

3. in-band signalling.

It's only the third of these options for signalling which requires a reserved property at all; therefore, I don't know if we should say we require a reserved property.

There's also *so many* upsides to the option 2 of signalling that I seriously question whether we need or want option 3 at all.  Especially if option 3 means we would lose the ability to use certain strings as map keys, which is... ouch.

warpfork: (Okay, this will be getting on a tangent, but I was thinking about the chooseable-transparency thing some more and schemas alone might not *always* do the thing we want.  I've mostly thought about that in the context of UnixFSv2 where we'll sometimes want to see "through" files to the raw blocks so we can get e.g. streaming the start of the file.  And in that case, it's clear how sending a different schema attached to a selector would allow that; files are a leaf.  But what if we wanted to do something like path normally through two directories, and then ask for only some shards of a third?  Since dirs are recursively defined in the hypothetical unixfsv2 schema, this is trickier: we might need something else: say, perhaps sending a replacement schema (which would switch the defn of dirs to the raw blocks rather than the advanced layout) that should only kick in beyond a certain path?  But this is complicated and not a thing we're trying to definitively solve today, so i'm gonna share but then park this train of thought.)

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+While the `_type` property is reserved in every map regardless of where it appears in a block, IPLD Generics **MUST** include the value of `_type` in the root property of a discrete block. Note: you can still embed multiple IPLD Generics in a single block using [inline blocks]().
+
+### Version 1
+
+The `_type` property is a string identifier. This identifier is used to lookup the implementation and if it cannot be found by the host environment any operation is expected to throw an exception.
```
warpfork: I thought one of the other original important ideas of what we'd put under our 'reserved' key in the root map included storing some opaque algorithm parameters (such as bitwidth params for the hamt example), so isn't this always going to be a map, and we should put the implementation identifier further in it?

mikeal: the bitwidth params for a hamt are just stored like any other normal property on the root block of that structure.

schema-layer/data-structures/generics.md
mikeal: Another update from our in-person chat, while the reservation of “/“ is a little problematic, it is *already* reserved for `dag-json` and is a problematic key for paths anyway. Rather than reserving yet-another-keyword I think it best to move this from `_type` to `{“/“: {_type: ... }}`.

mikeal: I’ll get into more details later, but I’m not at all convinced that we don’t need in-band signaling. I’ll write up the motivations in more detail (TLDR; you can’t effectively link to other people’s data freely if that data isn’t self-describing, you’d have to add application logic to layer on the out-of-band schemas for any data you ever link to), but for now I’m going to continue to detail this method of self-description.

I don’t think this is the **only** signaling method, but it’s one that we need to support. And, the other forms of signaling we’re considering with schemas are built quite nicely on top of it (you just use a schema to set the reserved property, you wouldn’t need any special syntax).

mikeal: > It's only the third of these options for signalling which requires a reserved property at all; therefore, I don't know if we should say we require a reserved property.

I should adjust the language here to make this more clear. By “reserved” I do not mean “this is the only way to signal the implementation of a generic.” What I actually mean is “do not use this property for anything *other* than this signaling” because if they do things will break when trying to read this data in APIs that support IPLD Generics.

warpfork: The problem here is that you're talking about creating a reserved key *at the data model layer*, and we can't do that and keep our shorts on.

>  already reserved for dag-json

1. First and foremost, we mostly regret that.  There's conversations about how to mitigate and hopefully move away from this; we don't want to start another one.
2. the operative wording is "in dag-json" -- It's a concern that's completely encapsulated **in *that instance of* a codec, at the codec layer** and ***not*** in the data model.  That makes it different from what you're proposing here, because it's got a significantly more limited splash radius and we actually *do* have ways to move away from it.
3. We mostly figure on marking dag-json as a debugging-only not-for-all-data extremely-caution-before-using format because of this asciifraking incompleteness.  We can do this -- effectively, flag dag-json as a not-quite-'native' (e.g. doesn't support all values) format -- and I think we should do so; it's honest.  Further, @Stebalien has seriously floated that we might correspondingly propose a new json-esque format that uses deeper nesting of objects simply to ensure it doesn't violate control-plane-vs-data-plane.  (I don't think getting someone to actually implement that is a high priority, but I do think it's a good idea and sets the goalposts in the right places.)

>  is a problematic key for paths anyway

We can *do* escaping in paths.  This is simple (if in the 'SMOP' sense): define paths as a series of segments which are opaque strings; define the path representation as a string itself to be slash-joined plus escaping.

I'm not super worried about this.

We've used the pattern of "here's the canonical 'AST' form; here's a string simplification that's usually usable and definitely well-defined" in several places already.  We can use it again here.

> you can’t effectively link to other people’s data freely if that data isn’t self-describing, you’d have to add application logic to layer on the out-of-band schemas for any data you ever link to

This is a real and valid concern for sure.  It's also one we can solve.  Reserved words at the data model layer which either asciifrak us or introduce escaping _at the data model layer, regardless of codecs_, is *not* something we can solve again once introducing it.

We're already doing this pattern of application logic handling it in e.g. how the gateways turn URL paths into unixfsv1 lookups.  If we can do this again, but with standard conventions and library features rather than one-offs of that particular application, then we're winning.

I think the way you phrased that actually hints at happier solution paths: if the data isn't self-describing, we can provide the description with the link.  This is as good a place as any to provide the description, isn't it?

I wonder if we can come up with some more ways to describe this without using the term "out-of-band"?  "Out-of-band" seems like a pretty sharp-edged thing, and istm we've got more of a gradient: since our links are 'in-band' themselves, if we're associating data with them, is that really out-of-band in any negative ways?

(Maybe a helpful analogy, maybe not: This draws up associations in my mind of when someone was asking how we might for example associate temperature units with numbers: what if "C" and "F" isn't expressive enough -- what if someone uses "C" to mean something *else* that's not even a temperature unit?  The answer to this *isn't* to try reach into the ether and try to conjure Gödel numbers unique to the universe they're embedded in; the answer is "context matters", we can use other data from earlier in the parse to feel more confident we know something about what we're looking at, and *it's okay* lean on that.  Things get worse when we try to deny context; things get more tractable when we acknowledge it and put it to work.)

---

tl;dr and with vigor: Escaping of keys in the data model is a no fly zone.  The compat story with other formats and existing data will be insane.  The performance will be insane (half the reason CBOR is fast is there's no escaping to check for wasting O(n) cycles on every string!).  Everything would become a fractal of bad.  It's pandora's box.  We cannot open it and still produce a viable standard.

Defining escaping for string-represented paths is viable in ways that escaping or reserving keys in the data model itself is not.  If there's a tug of war between those two: escape paths.

schema-layer/data-structures/generics.md
vmx: I thought of the name "generics" during our conversations on the call last week, but wasn't fully convinced, hence didn't mention it. Seems like a name at least two of us are OK with.

schema-layer/data-structures/generics.md
vmx: > This spec won’t define _how_ you implement named parameters

So it also won't specify the name of the parameter? I'm thinking again of the Go-style uppercase JSON field names.

vmx: I would also mention that `Value` can be any kind (also a `List` of `Map`).

schema-layer/data-structures/generics.md
vmx: So it's *only* top-level? You won't have a way to discover with paths for `GET` are available? I don't want to say we would need that, I just want to bring it up.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+Used for property lookups. Enables functionality similar to what you would expect from `Map` and `List` kinds.
+
+Only a single leaf `{ result: Value }` must be received.
+
+### KEYS `{}`
+
+Use to produce an iterator of every top level property available in the data structure. Takes no options.
+
+Multiple leaf `{ result: List(...keys) }` may be received.
```
vmx: We should come up with a "generally agreed" simple syntax for specs. I personally don't like Go syntax (e.g. []bytes) and I'd say it's the same for Go people reading JS syntax.

In this case I'd probably go for `{result: [keys]}` or `{result: List(keys)}`.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Implementations
+
+While each implementation has a language specific API there is a fair amount in common between implementations that is specified here.
+
+The interface for every operation can do the following:
+
+  * Return a result.
+  * Perform multiple calls of **any** operation on multiple targets of `value` or `path`
+	  *  `value` is an IPLD Data Model value.
```
vmx: Shouldn't the value be an IPLD Data Model "kind"?

schema-layer/data-structures/generics.md
vmx: Sounds OK so far.

schema-layer/data-structures/generics.md
vmx: This took me a some time to understand, due to the lengths of the sentences. What about splitting it into smaller ones. Here's a proposal, take this as an idea and make it read even nicer:

The term “leaf” is used often below. When an operation is performed on a node it may make any number of additional calls to operations on other data structures. Those will produce many results which are passed into continuations. “leaf” responses are those that are from the original target operation and not from subsequent operations trigger by the original operation.

schema-layer/data-structures/generics.md
mikeal: Working on a spec refactor that will:

* Split “Generic Definition” from the signal.
* Add an “open problem” section on how to best accomplish “in-data signalling”

schema-layer/data-structures/generics.md
mikeal: Ya, maybe the beginning of that can live at the top of this spec and we can iterate. Trying to find a generic way of describing “named parameter” has been tricky.

schema-layer/data-structures/generics.md
mikeal: yes, will fix.

schema-layer/data-structures/generics.md
Stebalien: We can't call these generics. A "generic" type is a type constructor or a type parametrized by another type. Calling these "generics" will confuse everyone.

Alternatives:

* Composite types
* Complex types
* Collection
* Composite Kinds

None of these are perfect but Generics is a no go (like, we-need-to-erase-every-mention-of-it no-go or we'll be answering questions on this till the end of time).

mikeal: I like “composite” because I don’t think I’ve ever seen it used as a name for anything before, so I’m +1 on composite types.

rvagg: "compound" is in a similar area.

Let's not totally throw out "advanced layouts" either, I think it has merit here.

mikeal: Most of the spec/system is a programmatic interface, so it seem odds to refer them as “layouts.”

vmx: I still like "advanced layout". To me that means data structures laid out in an advanced way. Think e.g. also about the use case were you want to store a big object and just want to split it internally into a maximum sized nodes (to get your stuff working with Bitswap).

rvagg: yeah, "advanced layout" is "this block isn't what it looks like"

Screen Shot 2019-06-14 at 8 54 25 pm

mikeal: > yeah, "advanced layout" is "this block isn't what it looks like"

Sure, but users of this system are moving past thinking in a block specific way, so it feels rather odd to describe that system in terms of the block layout since that’s exactly the mental model we’re trying to push them out of.

vmx: It's not necessarily about blocks. I think the user of IPLD still needs to be aware that their data is split into chunks (whether we call them "blocks" or not). So it's rather a "this piece of data isn't what it looks like, but has a different shape than you might think".

warpfork: I'm warmly neutral to "composite" and "complex" as part of the wording we settle on (which is about how I'd describe my overall warmth to 'advanced layouts' as well).

Bit dislike putting "types" or "kinds" immediately after the other word though, as it is in several of steb's examples above.

e.g. "composite kinds" is a bit odd -- it's not the kinds that are composite; the overall thing still acts like it a regular unsurprising not-novel kind from the enum of kinds we already know and love.

Moar brainstorming ftw?  I've got a piece of paper on which I also jotted "prisms" for example.  (Terrible?  Maybe.  Brainstorming?  Does go through phases of such...)

warpfork:
I'm not sure this is a particularly direct response to any of the diff content, but it's on the topic space as a whole, so as a top level comment:

Perhaps one way we could consider tersely describing the contract of these things is:

`function (LinkLoader, Node) -> (Node)`

Such a definition is:

- terse;
- can lean nicely on the properties and enumerations we already established in the Data Model docs;
- clearly points out this may be multi-block;
- gives us a lot of nice structure to communicate what this feature is and is not meant to encompass.

It might not be literally what the code is doing in our implementations, and as documentation it also still needs surrounding explanation, but structurally, does this seem about right?

schema-layer/data-structures/generics.md
rvagg: Yeah, "prisms" is like "lens" which is a concept I like here, "the lens through which you can turn these raw crazy blocks into useful, traveresable, queryable kinds". Probably a bit too esoteric for naming purposes though.

rvagg:
`function (LinkLoader, Node) -> (Node)` is conceptually accurate insofar as reads are concerned. Does that break down when you bring writes into it though? And therefore, is it even useful here as anything more than just a description for how traversal extended with these advanced types/generics?

schema-layer/data-structures/generics.md
rvagg: Remember that we justified the use of "kinds" with the idea that "types" in the schema layer was where most of the interesting action was going to be. Our current trajectory has us splitting off from schemas into this new place, entirely detached. So we no longer have a clear hierarchy and may end up having schema "types" being a more obscure thing that's only used when we invoke schemas for specifying a certain subsets of data layouts. We're currently heading to a place where we will mostly be talking about data layer kinds and these advanced layout / generic things most of the time, and "types" only occasionally.
We need to find the synthesis with schemas here. And if not, maybe "types" is back on the table and schemas can get something less general because they're in the corner doing a specific, limited job.

vmx: > Yeah, "prisms" is like "lens" which is a concept I like here

Before someone starts to like the idea of calling it "lens", I'm strongly oppose it. It might be niche, but it's a term from bi-directional transformations. I even wrote a JS lib implementing those: https://github.com/vmx/jslens

warpfork:
The write side is a bit different, but if you dream of it being based around a `NodeBuilder`, it's not *so* different: `function (Storer) -> (NodeBuilder)`: where `Storer` is a function that serialize nodes, writes them to disk, and returns their links, and it's used to take care of all the internal nodes.

(In go-ipld-prime, it'll be even more similar: `function (Storer, NodeBuilder) -> (NodeBuilder)`, because the first `NodeBuilder` is used to define the memory representations used for the internal node.  Although I should tag this as a "probably"; haven't written it yet.)

(Aside: I wrote a whole first draft of this comment based on `function (Storer, NodeBuilder) -> (Node)` before realizing I was doing too much in one step... that approach came out much worse.  Such a signature, in addition to doing too much in one step, would've been fairly terrible on the eyes for documentation/explanation purposes  -- `f(a, b) -> (b)` is nice because it makes clear implications about how an abstraction is staying the same on both sides; `f(a, b) -> (c)` says roughly nothing.)

But as to whether or not this is good as a description: well, maybe.  I think we'll get most of the descriptive benefits from putting `function (LinkLoader, Node) -> (Node)` pretty high in the documentation, right under it's own "H1" tag... and maybe if we do document `function (Storer, NodeBuilder) -> (NodeBuilder)`, it would be much further on down the page, under an "H3" tag of additional clarifying information, or possibly tucked in some notes and suggestions for implementers of new IPLD libraries in new languages.

schema-layer/data-structures/generics.md
mikeal: renamed to composites.

schema-layer/data-structures/generics.md
mikeal: this section has been removed and replaced with a section about definitions and the open problem of fat pointers.

schema-layer/data-structures/generics.md
mikeal: i want to capture it, even if we don’t have a great solution yet. it’s something we need to keep in mind as we desing the write interfaces.

mikeal:
Big changes were pushed:

* Removal of reserved property
* Addition of “definition” section
* Added “open problem” regarding fat pointers
* Remove the entire operations section and moved, temporarily, to the README in the experimental implementation. It’s not yet ready to be spec’d in a language agnostic way.

mikeal:
I’m going to merge this in 24h if there are no objections.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,73 @@
...
+What we need in order to move forward is some extension/modification to `CID` in order to signal that “the data being linked to is a composite definition” at which point we can safely ad semantics to `_type` or other properties without reserving any property universally.
+
+### Version 1
+
+The `_type` property is a string identifier. This identifier is used to lookup the implementation and if it cannot be found by the host environment any operation is expected to throw an exception. Implementations MUST NOT fallback to *Layer 1* operations of the contents of the node if they do not have an implementation.
+
+Example:
+
+```json
+{ “_type”: “IPLD/Experimental/HAMT/0” }
```
rvagg: needs ascii `"`

schema-layer/data-structures/generics.md
warpfork: That's well and good, but capturing a list of things we want to think about, for which we don't have clear solutions we're about to write up yet... sounds like something that belongs in a design history notes document moreso than something we commit to in a specs-track document.

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Composite Definition
+
+Composite definitions describe how to find an implementation of the data structure. When encoded into the data these also serve as a the self-description mechanism.
+
+***Open Issue: Fat Pointers***
+
+Early experiments simply reserved the `_type` property for composites to describe themselves. Reserving this property by default across any data in any block is highly problematic and makes it impossible to express certain data in IPLD.
+
+What we need in order to move forward is some extension/modification to `CID` in order to signal that “the data being linked to is a composite definition” at which point we can safely ad semantics to `_type` or other properties without reserving any property universally.
```
warpfork: The preceding paragraph is good, but the conclusion in this paragraph is not the only one possible from that setup, and yet phrases itself as if it is ('need').  I believe there are other solutions to this issue that aren't explored here -- and may be better -- and therefore this isn't very suitable to merge in a specs-track document.  (But, again, it would be great design history notes material.)

schema-layer/data-structures/generics.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Version 1
+
+The `_type` property is a string identifier. This identifier is used to lookup the implementation and if it cannot be found by the host environment any operation is expected to throw an exception. Implementations MUST NOT fallback to *Layer 1* operations of the contents of the node if they do not have an implementation.
```
warpfork: @mikeal I think your own comment at https://github.com/ipld/specs/pull/131#discussion_r301324246 represents a more up-to-date understanding than this, and conflicts with (or at least should update) the "expected to throw an exception" statement.

Additionally, we've on several occasions discussed the idea that one *might* explicitly do operations such as: send a selector paired with instructions *to intentionally* use Layer 1 operations on the content of a node (e.g., request the left-leaning tree of nodes in a file).  While that idea also remains flagged as "exploratory" until we build a PoC or such, it's an idea that certainly hasn't been conclusively dismissed, and conflicts with this "MUST NOT".  Perhaps some other phrasing using phrases about "must not sliently" or "must not implicitly" would address this?

warpfork:
I would prefer to move this to a design history document, write up some other learnings afterwards, and later make a fresher, more limited, and clarified doc for spec drafts.

I like all the changes in the [latest set](https://github.com/ipld/specs/pull/126#issuecomment-509715141), but there are several things in this file that make me uncomfortable, particularly with merging it as a spec-track document:

- There are [several matters discussed](https://github.com/ipld/specs/pull/126/files#r291765173) in the doc that are *interesting* to our design process, but we haven't concretely proposed solutions to yet, nor necessarily are we even sure they're part of this topic.  This is fine for design exploration notes, but not good for a spec document.
- We have not resolved that this new name is by any consensus superior to the old name.  It may be, but we haven't polled for agreement on that; and, since the existing name exists in several other widely circulated documents already, we should be engaging this as a *rename* and take it with suitable aplomb.
- We have a lot of very similar concepts, approached with different granularity of definition, in issue #130.  Could we write a better document by starting again and selecting a hybrid of the best parts from that issue and this document?
- ... and several other comments raised inline.

Proposal: let's move this file to the `./design-history/exploration-reports` folder, date it, and call it lessons learned -- I like to see things land too, and I'd be more than happy to see it merge *there*, as it more accurately represents the state of play.

rvagg:
If we're going with the design-history thing then I agree that this might be better there (I'm still not convinced about design-history but it's a thing for now). The main hint in here is the digression into `_type` which we seem to be agreeing is just a placeholder for some superior mechanism yet defined. So it's more of a just-make-it-work-for-now description than a spec.

schema-layer/data-structures/generics.md
mikeal: Good point, I think I’ll break this into a section called “considerations” rather than “requirements.”

schema-layer/data-structures/generics.md
mikeal: I think “need” is appropriate in the context of “fat pointers.” Unless there’s an alternative I’m not aware of, something in the CID needs to signal the use of a fat pointer, whether that’s a difference in codec, codec prefix, or CID version. If there’s another path to fat pointers that can be encoded into the data I’d love to hear it and incorporate it, but if this is just another branch of the “out of band signal” thread then I’m inclined to leave the statement as-is.

schema-layer/data-structures/generics.md
mikeal: I think the problem is that the nesting of the document isn’t providing sufficient context.

This section is now a sub-section of the “Defintion” section. Meaning that, this is a reserved property “of a Composite Definition” and not a reserved property generally. Same goes for the “Version 2” section. I think I need to add a statement to this effect in the section since the nesting itself isn’t sufficient context.

schema-layer/data-structures/generics.md
whyrusleeping: does merging this document imply that were all in agreement that we will be using _type to convey type information?

mikeal:
> Proposal: let's move this file to the ./design-history/exploration-reports folder, date it, and call it lessons learned -- I like to see things land too, and I'd be more than happy to see it merge there, as it more accurately represents the state of play.

The entire purpose of having an “Exploratory” stage is to avoid having a bunch of exploratory documents in notes or “design history.”

The goal of this document is to write out and align effort going into Composites, as it will soon include at last 3 people on the team. Having people aligned on terminology and direction is the primary purpose of the document, as is the purpose of an “Exploratory” stage. 

A document being in an exploratory stages does not solidify any of the approach, it’s all still subject to change or even to be discarded entirely. But, while people are actively trying to make forward progress, we need a way to write down and align that work, which is what the document is attempting to do. You may not agree with where all of the discussions ended up, and we may still return to them later, but in the interest of making forward progress we need to “disagree and commit” for the purposes of continuing to be able to do more exploratory development.

> So it's more of a just-make-it-work-for-now description than a spec.

This is accurate, and should be what we assume the state of any Exploratory specification is.

We need to stop having PR’s open indefinitely because we lack 100% alignment on topics we have yet to even fully implement. We can continue to have discussions and make changes to Exploratory specifications, in fact, that’s one of the reason why we have a staging system at all.

warpfork:
Regarding the describing this is "spec-track": even if labeled 'exploratory', it *is* taking up a slot in the file tree that makes it hard to propose anything else in parallel.  I think whether or not something takes up a 'slot' in such a way is the crux of the matter in communication and development flow in practice, and that's why I describe this as 'spec-track'.

I think we may actually want to update the governance doc to suggest that exploratory docs simply go in a different directory (as we've started doing already), and we don't use the main directories until things are at least at the 'draft' stage.

schema-layer/data-structures/generics.md
mikeal: Now that I’m thinking about it, I need to add a paragraph in the “Definition” to talk about applying definitions, and start to map out how in-band and out-of-band signaling might work.

mikeal:
> I think we may actually want to update the governance doc to suggest that exploratory docs simply go in a different directory (as we've started doing already), and we don't use the main directories until things are at least at the 'draft' stage.

+1000, I’ll work this into my next spec change.

schema-layer/data-structures/generics.md
mikeal: > does merging this document imply that were all in agreement that we will be using _type to convey type information?

The document doesn’t actually say this. `_type` is defined as part of the “Composite Definition” which is being set aside so that we don’t reserve the property in a general way, it’s only defined as part of a “definition.” We’ve already conceded that there needs to be an additional signal to say “this is a definition” at which point we add these semantics to the property as to avoid reserving the property name.

Merging the document conveys that, in the experiments we’re doing, we’ll use a string `_type` (we already do) **in** definitions.

I updated the paragraph in the `_type` section to be a little more explicit about this only being “when a definition is applied.”

schema-layer/data-structures/generics.md
mikeal: I updated this paragraph to be more explicit about the context for the property. Let me know if you think it’s sufficient.

schema-layer/data-structures/generics.md
warpfork: Something about this sentence still bothers me, and it's difficult to put a finger on it.  It's not the "_type" string; not super concerned with the string itself, now that it's clear we're out of the blue-boxing trap.

Something seems fishy because we have examples of systems that work today without "extension or modification of CID".  The current gateways use HAMTs and other shard systems for dirs and files; and they didn't need an extension or modification of CID.  Filecoin is planning to use several fancy layouts; and they don't seem to be discussing an extension or modification of CID.  If we can't describe and reason about whatever these systems are doing, we've got a fairly serious blind spot.

Maybe some extension or modification of CID will be wise.  If fat pointers count as that (... do they?  not clear to me if all variations of fat pointers are necessarily modifications of CID), maybe they'll be wise.  But stating it as "need" and without alternatives seems wrong to me, given that we can observe systems evolving *right now* which very much don't seem to "need" this.  And I'm very wary of making a committal statement like "CIDs need to change".  (Maybe that's not how this is supposed to taken, but that's the idea I keep getting when I re-read the current wording.)

---

ANYWAY, I'm still thumbs up on merging this file overall under the [exploration-reports heading](https://github.com/ipld/specs/pull/126#issuecomment-510013092) in its current state, mind.  I'm just really very conscious of this cluster of neurons firing for me saying we're still missing something, even if I can't articulate it particularly well yet.  (I think I have some ideas, but I had to descend *way* deep into the Turing Tarpit to get them, and it might take a bit to write up readably.)

schema-layer/data-structures/generics.md
warpfork: I like this new phrasing :+1: Indicates that silent fallbacks are wrong without excluding the possibility of explicitly different modes of operation :ok_hand: super good

schema-layer/data-structures/generics.md
mikeal: So, my current thinking is that we probably will need an extension to CID, but given how much my thinking has evolved on this recently I’m going to back out definitive sounding language from the statement.

rvagg:
This was merged with the file named 'generics.md', the title "Composites" and an ongoing discussion about whether even that's the right name. That's strong proof to me that this belongs in design/ history/ or whatever that directory is going to be.

mikeal:
Ya, that was an oversight, it’s fixed in the “rename PR” that moves it to ‘design/composites-A.md’

-Mikeal

________________________________

PR #128, opened 2019-06-11
move cardinality into schema-layer

rvagg:
Moving concepts/cardinality.md into schema-layer and link it from README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ The Schema Layer will also serve as an enabling layer for complex multi-block da
 |     |      |
 |-----|------|
 | [Specification: IPLD Schemas](schema-layer/schemas/schemas.md) | [schema-layer/schemas/schemas.md](schema-layer/schemas/schemas.md) |
+| [Concept: Cardinality](schema-layer/cardinality.md) | [schema-layer/cardinality.md](schema-layer/cardinality.md) |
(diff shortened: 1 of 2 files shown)


vmx (approved):

warpfork:
I'm kinda opposed.  Cardinality is a concept that also applies to the data model layer (there's just a *lot* of things that are "inf"), as well as something we'll perhaps want to reflect more on for building good documentation for what is and is not a 'native' codec and why.

I'd also like to add more things to a "concepts" dir that include such things as...

- a brief primer on Gödel numbers and in particular https://en.wikipedia.org/wiki/G%C3%B6del_numbering#Lack_of_uniqueness
- a link to https://en.wikipedia.org/wiki/Lambda_calculus#Undecidability_of_equivalence and "there is no divine interpreter"
- and generally a bunch of TANSTAAFL and other dream-breakers stuff that's useful to be able to reference when stating why we've chosen to make practical choices that we have.

But then, acknowledging I haven't written any of these other things, if we want to remove a single-content directory until more *does* exist, I guess I shan't protest too much.

rvagg:
Well I don't want a "concepts" directory unless we move all concepts into that directory, which might be fine but we lose a bit of context for some of them. Or, you could come up with a new name for a "concepts" directory that doesn't make it look like all our "concepts" should be in that place. Or you could keep "concepts" for a directory name but change the use of "Concept" in all of the current concepts docs to some other term.

PR #129, opened 2019-06-11
Proposal: new version of dag-json

mikeal:
This spec change would resolve the outstanding issue of `dag-json` not being able to represent the “/“ key.

Since “/“ is already reserved, we use that key for the version definition, which makes this new version backwards compatible with all of the existing `dag-json` data.

This change would also set up `dag-json` to store data outside of the actual node data, which could be used similar to how tags are used in cbor or for something like https://github.com/ipld/js-generics/issues/3

### DIFF
--- a/block-layer/codecs/DAG-JSON.md
+++ b/block-layer/codecs/DAG-JSON.md
@@ -14,14 +14,56 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
 JavaScript itself that has trouble with them. This means JS implementations
 of `dag-json` can't use the native JSON parser and serializer.
 
+### Version 0
+
+This is an old version of `dag-json` that reserved the `"/"` key in order to
+represent binary and link data types.
+
 #### Binary Type
 
 ```javascript
 {"/": { "base64": String }}
 ```
 
-### Link Type
+#### Link Type
 
 ```javascript
 {"/": String /* base encoded CID */}
+```
+
+### Version 1
+
+#### Format
+
+The internal data format is valid JSON but is **NOT** identical to the decoded
+node data codecs produce.
+
+Example internal format:
+
+```javascript
+{ "/": { "_": 1 },
+  "data": { "hello": "world": { "obj": { "array": [0, 0] } } },
+  "meta": {
+    base64: [
+      [[ "key" ], "dmFsdWU="],
+      [[ "obj", "key"], "dmFsdWU="],
+      [[ "obj", "array", 0], "dmFsdWU="]
+    ],
+    links: [
+      [["obj", "array", 1], "zb2rhe5P4gXftAwvA4eXQ5HJwsER2owDyS9sKaQRRVQPn93bA"]
+    ]
+  }
+}
+```
+
+Decodes to:
+
+```javascript
+{ hello: 'world',
+  key: Buffer.from('value'),
+  obj: {
+    array: [ Buffer.from('value'), new CID('zb2rhe5P4gXftAwvA4eXQ5HJwsER2owDyS9sKaQRRVQPn93bA')],
+    key: Buffer.from('value')
+  }
+}
 ```
--- a/block-layer/codecs/DAG-JSON.md
+++ b/block-layer/codecs/DAG-JSON.md
@@ -42,7 +42,7 @@ Example internal format:
 
 ```javascript
 { "/": { "_": 1 },
-  "data": { "hello": "world": { "obj": { "array": [0, 0] } } },
+  "data": { "hello": "world", { "obj": { "array": [0, 0] } } },
   "meta": {
     base64: [
       [[ "key" ], "dmFsdWU="],


block-layer/codecs/DAG-JSON.md
```diff
@@ -14,14 +14,56 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
...
+#### Format
+
+The internal data format is valid JSON but is **NOT** identical to the decoded
+node data codecs produce.
+
+Example internal format:
+
+```javascript
+{ "/": { "_": 1 },
+  "data": { "hello": "world": { "obj": { "array": [0, 0] } } },
```
rvagg: `s/"world":/"world",/`

block-layer/codecs/DAG-JSON.md
```diff
@@ -14,14 +14,56 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
...
-```
\ No newline at end of file
+```
+
+### Version 1
+
+#### Format
+
+The internal data format is valid JSON but is **NOT** identical to the decoded
+node data codecs produce.
```
daviddias, line 39: I do not understand why this is necessary. Is there a place with the rationale to make JSON not JSON?

block-layer/codecs/DAG-JSON.md
```diff
@@ -14,14 +14,56 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
...
+
+Example internal format:
+
+```javascript
+{ "/": { "_": 1 },
+  "data": { "hello": "world", { "obj": { "array": [0, 0] } } },
```
vmx, line 45: So those `[0, 0]` are placeholders? So if you want to encode `[CID, "some string"]` it would be `[0, "some string"]`?

block-layer/codecs/DAG-JSON.md
```diff
@@ -14,14 +14,56 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
...
+
+#### Format
+
+The internal data format is valid JSON but is **NOT** identical to the decoded
+node data codecs produce.
+
+Example internal format:
+
+```javascript
+{ "/": { "_": 1 },
```
vmx, line 44: Wouldn't it be simpler to just add a field called `version` and move `meta` and `data` to the top-level?

block-layer/codecs/DAG-JSON.md
```diff
@@ -14,14 +14,56 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
...
+  "data": { "hello": "world", { "obj": { "array": [0, 0] } } },
+  "meta": {
+    base64: [
+      [[ "key" ], "dmFsdWU="],
```
vmx, line 48: I key see where those tuples come from. Why not using an "overlay object" with proper nesting as you did in your other dag-json experiment?

Is it because you could get the binary data/links without traversing a nested structure?

block-layer/codecs/DAG-JSON.md
```diff
@@ -14,14 +14,56 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
...
+    ]
+  }
+}
+```
+
+Decodes to:
+
+```javascript
+{ hello: 'world',
+  key: Buffer.from('value'),
```
vmx, line 63: I would really prefer if we'd move to an Browser first approach and not using Node.js primitives:

```JS
    key: new TextEncoder().encode('value'),
```

block-layer/codecs/DAG-JSON.md
mikeal: Right now, we’re using a reserved key, `”/“`, for encoding links and binary. Several people think that key reservation is highly problematic because it means there are certain data sets you just can encode.

This change would fix that changing the internal block format. The format would **still* be valid JSON, it just wouldn’t have as close of a 1-to-1 match with the decoded form of the data.

However, this change may not be worth it. This PR is to show that we have a solution to the reserved key issue and that is what it would take to fix it.

block-layer/codecs/DAG-JSON.md
mikeal: They are just placeholders to preserve the ordering of the array.

block-layer/codecs/DAG-JSON.md
mikeal: “meta” and “data” *are* top level. using “_” was just to save space but we can do whatever we want.

block-layer/codecs/DAG-JSON.md
mikeal: If you recall, we considered that in a `dag-json` alternative proposal and the problem, which I think you ended up alerting me to, was how to represent sparse values in an array.

block-layer/codecs/DAG-JSON.md
mikeal: it’s just not entirely clear to everyone that this does binary encoding. perhaps we should just move to psuedo-code, something like `Bytes(‘value’)`

block-layer/codecs/DAG-JSON.md
vmx: I'd use `null` for that, but that's a minor detail. Perhaps the example could be expended to `[null, "some string", null]` to make it clearer that there are place holders which are mixed with actual values.

block-layer/codecs/DAG-JSON.md
vmx: Oh, sorry I missed that they are top-level.

block-layer/codecs/DAG-JSON.md
vmx: Right, that would be a lot of placeholders then.

block-layer/codecs/DAG-JSON.md
vmx: That would work for me. I just don't want to give the impression it's a Node.js Buffer.

Stebalien:
Is the motivation _just_ to support `/` keys? IMO, dag-json exists so users could read and write IPLD using JSON. I'm concerned this format looses that.

If we _just_ want to support `/` in keys, we could use some form of escaping syntax. For full backwards compatibility, we could do something like `{"/": {"literal": foobar}}`. Alternatively, we can _probably_ allow escaping forward slash without breaking too much: `//` -> `/`.

block-layer/codecs/DAG-JSON.md
mikeal: originally it was null but i went with 0 because null encoding is extra bytes :/

mikeal:
> IMO, dag-json exists so users could read and write IPLD using JSON. I'm concerned this format looses that.

I’m a little confused by this use case, because the `dag-json` codec already isn’t `json`. If you want to take a block of existing `json` data you should just use a `json` codec (which we should implement since it would be trivial).

If you want to take existing JSON data and add links to it, you’re going to have to use an encoding interface anyway and all of these details are invisible.

Is there a case I’m missing where you want `dag-json` instead of `json` **and** you’re interacting directly with the block format?

> Is the motivation just to support / keys?

Yes, that’s the only thing we get out of this change. Using “/“ as a key will *still* be highly problematic for anyone using paths and selectors, but this would fix the issue at the codec level.

I’m actually not too worried about “/“ being reserved in `dag-json` but other people seem to be. I opened this PR to show what it would take to fix the issue, given some people are concerned.

Stebalien:
> I’m a little confused by this use case, because the dag-json codec already isn’t json. If you want to take a block of existing json data you should just use a json codec (which we should implement since it would be trivial).

dag-json wasn't designed to be 100% compatible but still mostly compatible. I agree we should have a plain JSON format for users who just want JSON but the point of _this_ format was to provide a familiar, JSON-like format that mostly "just works" (i.e., you can link up existing JSON objects).

vmx:
Another thing that might be worth keeping in mind. Having a format you can easily use as exchange format for tests: https://github.com/multiformats/rust-cid/pull/17#issuecomment-502395133

vmx:
@mikeal We decided to put this into design history, but after having a closer look  I think the discussion can be easily be summed up. Instead of having it archived automatically I suggest that we create a proper exploration report like https://github.com/ipld/specs/blob/035683c97d0280de5e2d490822d63ad618a8acab/design/history/exploration-reports/2019.06-unixfsv2-spike-01.md. @mikeal if you think it's worth having it there, could you please create a new PR with that report?

mikeal:
Closing as this was mostly informational. @vmx will use his script to create an exploration report for this.

warpfork:
This is already commented as closed but not closed, heh.  Closing.

Issue #130, opened 2019-06-11
Identifying the critical decisions in IPLD advanced data structure design


warpfork:
(Forward: this is a collection of a *lot* of stuff, and I'm trying to be comprehensive, but will inevitably fail.  If you want to read this and visualize semi-continuous injections of "TODO: expand" and "TODO: more alternatives?", that will probably be appropriate!)

---

There are approximately three distinct and critical decisions we need to regard:

1. Signaling
2. Referencing
3. Having the implementation

### Signaling

"Signaling" refers to deciding / detecting that we need any advanced layout
engagement at all, rather than simply having another regular map, list, or
byte sequence.

Note that "signaling" refers *only* to the conditional branch that decides
we're going to do *something* -- it doesn't define what the something *is;
that's the "referencing" decision.

(It's useful to regard signaling as a distinct issue because we should be
able to point on a flow control map to the exact position where signaling
occurs -- it's _one point_.)

### Referencing

(alternate titles: "rendezvous", "implementation selection")

"Referencing" refers to the need to pick a particular implementation of code
we're about to use to understand the data layout.

Roughly, "referencing" is about how something says "I'm a hamt!" or
"i'm a rabin chunked byte sequence!".

Note that referencing is about *which* of those things a data structure is;
the "signaling" decision already handled whether or not *something* special
is about to happen.  (Referencing can involve more complex decision trees,
whereas signaling is a pure boolean.)

### Having the implementation

"Having the implementation" refers to the need to have actual executable code
which understands and manipulates the data layout.  This might refer to a
plugin registration system, or a bytecode interpreting system, or... etc.

### Other

Some advanced layout algorithms have additional parameters than merely their reference;
for example, HAMTs can have a 'bitwidth' parameter.

These additional parameters may be worthy of special consideration, because
they have the interesting property of being variable without needing *new code*
in the client libraries acting on the data.  (This also means that parameters
as recorded in serial data can vary in ways that don't affect whether or not
we "have the implementation", which may be useful; however, by the other side
of the same coin, it poses further interesting questions.)

We could also choose to remove these from the field of play entirely and
simply state that the must be part of the Referencing data overall and
opaque to the rest of the system.

Known (so far) solution vectors
-------------------------------

(These are selected vectors in the solution space; they might not be exhaustive.)

#### defacto signaling

We can implement signaling by leaving it completely unspecified in the data
and completely unstandardized in general, instead having applications "know"
what advanced layouts they expect where.

[...]

PR #131, opened 2019-06-12
HashMap (HAMT) spec

rvagg:
An iteration of #109

### DIFF
--- a/schema-layer/data-structures/hashmap.md
+++ b/schema-layer/data-structures/hashmap.md
@@ -109,13 +109,9 @@ type HashMapNode struct {
 }
 
 type Element union {
-  | Child "child"
-  | Bucket "bucket"
-} representation keyed
-
-type Child union {
-  | Link link
   | HashMapNode map
+  | &HashMapNode link
+  | Bucket list
 } representation kinded
 
 type Bucket list [ BucketEntry ]
--- a/schema-layer/data-structures/hashmap.md
+++ b/schema-layer/data-structures/hashmap.md
@@ -1,0 +1,300 @@
+# Specification: HashMap
+
+**Status: Prescriptive - Draft**
+
+* [Introduction](#Introduction)
+* [Useful references](#Useful-references)
+* [Summary](#Summary)
+* [Structure](#Structure)
+  * [Parameters](#Parameters)
+  * [Node properties](#Node-properties)
+  * [Schema](#Schema)
+* [Algorithm in detail](#Algorithm-in-detail)
+  * [`Get(key)`](#Getkey)
+  * [`Set(key, value)`](#Setkey-value)
+  * [`Delete(key)`](#Deletekey)
+  * [`Keys()`, `Values()` and `Entries()`](#Keys-Values-and-Entries)
+  * [Differences to CHAMP](#Differences-to-CHAMP)
+  * [Canonical form](#Canonical-form)
+* [Use as a "Set"](#Use-as-a-%22Set%22)
+* [Implementation defaults](#Implementation-defaults)
+  * [`hashAlg`](#hashAlg)
+  * [`bitWidth`](#bitWidth)
+  * [`bucketSize`](#bucketSize)
+  * [Maximum key size](#Maximum-key-size)
+  * [Inline values](#Inline-values)
+* [Possible future improvements and areas for research](#Possible-future-improvements-and-areas-for-research)
+  * [Maximum depth limitations](#Maximum-depth-limitations)
+  * [Hash algorithm](#Hash-algorithm)
+  * [Buckets](#Buckets)
+  * [Security](#Security)
+
+## Introduction
+
+The IPLD HashMap provides multi-block key/value storage and implements the Map [kind](/data-model-layer/data-model.md#kinds) as a [generic](generics.md) ***(TODO: "generic" or "advanced data structure" or "advanced layout"?)*** in the IPLD type system.
+
+The IPLD HashMap is constructed as a [hash array mapped trie (HAMT)](https://en.wikipedia.org/wiki/Hash_array_mapped_trie) with buckets for value storage and [CHAMP](https://michael.steindorfer.name/publications/oopsla15.pdf) mutation semantics. The CHAMP invariant and mutation rules provide us with the ability to maintain canonical forms given any set of keys and their values, regardless of insertion order and intermediate data insertion and deletion. Therefore, for any given set of keys and their values, a consistent IPLD HashMap configuration and block encoding, the root node should always produce the same content identifier (CID).
+
+## Useful references
+
+* [Fast And Space Efficient Trie Searches](https://infoscience.epfl.ch/record/64394/files/triesearches.pdf) by Phil Bagwell, 2000, and [Ideal Hash Trees](http://lampwww.epfl.ch/papers/idealhashtrees.pdf) by Phil Bagwell, 2001, introduce the AMT and HAMT concepts.
+* [CHAMP paper](https://michael.steindorfer.name/publications/oopsla15.pdf) presented at Oopsla 2015 by Steinforder & Vinju
+* [Java implementation](https://github.com/msteindorfer/oopsla15-artifact/) accompanying the original CHAMP paper (see https://github.com/msteindorfer/oopsla15-artifact/blob/master/pdb.values/src/org/eclipse/imp/pdb/facts/util/TrieMap_5Bits.java and other TrieMap files in the same directory).
+* [Optimizing Hash-Array Mapped Tries for Fast and Lean Immutable JVM Collections](https://blog.acolyer.org/2015/11/27/hamt/) a high-level description of HAMT data structures in general and the specifics of CHAMP.
+* Peergos [CHAMP](https://github.com/Peergos/Peergos/blob/master/src/peergos/shared/hamt/Champ.java) implementation
+* [IAMap](https://github.com/rvagg/iamap) JavaScript implementation
+* [go-hamt-ipld](https://github.com/ipfs/go-hamt-ipld) Go implementation
+
+## Summary
+
+The HAMT algorithm is used to build the IPLD HashMap. This algorithm is common across many language standard libraries, particularly on the JVM (Clojure, Scala, Java), to power very efficient in-memory unordered key/value storage data structures. We extend the basic algorithm with buckets for block elasticity and strict mutation rules to ensure canonical form.
+
+The HAMT algorithm hashes incoming keys and uses incrementing subsections of that hash at each level of its tree structure to determine the placement of either the entry or a link to a child node of the tree. A `bitWidth` determines the number of bits of the hash to use for index calculation at each level of the tree such that the root node takes the first `bitWidth` bits of the hash to calculate an index and as we move lower in the tree, we move along the hash by `depth x bitWidth` bits. In this way, a sufficiently randomizing hash function will generate a hash that provides a new index at each level of the data structure. An index comprising `bitWidth` bits will generate index values of  `[ 0, 2`<sup>`bitWidth`</sup>` )`. So a `bitWidth` of `8` will generate indexes of `0` to `255` inclusive.
+
+Each node in the tree can therefore hold up to `2`<sup>`bitWidth`</sup> elements of data, which we store in an array. In the IPLD HashMap we store entries in buckets. A `Set(key, value)` mutation where the index generated at the root node for the hash of `key` denotes an array index that does not yet contain an entry, we create a new bucket and insert the `key` / `value` pair entry. In this way, a single node can theoretically hold up to `2`<sup>`bitWidth`</sup>` x bucketSize` entries, where `bucketSize` is the maximum number of elements a bucket is allowed to contain ("collisions"). In practice, indexes do not distribute with perfect randomness so this maximum is theoretical. Entries stored in the node's buckets are stored in `key`-sorted order.
+
+If a `Set(key, value)` mutation places a new entry in a bucket that already contains `bucketSize` entries, we overflow to a new child node. A new empty node is created and all existing entries in the bucket, in addition to the new `key` / `value` pair entry are inserted into this new node. We increment the `depth` for calculation of the `index` from each `key`'s hash value to calculate the position in the new node's data array. By incrementing `depth` we move along by `bitWidth` bits in each `key`'s hash. With a sufficiently random hash function each `key` that generated the same `index` at a previous level should be distributed roughly evenly in the new node's data array, resulting in a node that contains up to `bucketSize` new buckets.
+
+The process of generating `index` values from `bitWidth` subsections of the hash values provides us with a depth of up to `(digestLength x 8) / bitWidth` levels in our tree data structure where `digestLength` is the number of output bytes generated by the hash function. With each node able to store up to `2`<sup>`bitWidth`</sup> child node references and up to `bucketSize` elements able to be stored in colliding leaf positions we are able to store a very large number of entries. A hash function's randomness will dictate the  even distribution of elements and a hash function's output `digestLength` will dictate the maximum depth of the tree.
+
+A further optimization is applied to reduce the storage requirements of HAMT nodes. The data elements array is only allocated to be long enough to store actual entries: non-empty buckets or links to actual child nodes. An empty or `Null` array index is not used as a signal that a `key` does not exist in that node. Instead, the data elements array is compacted by use of a `map` bitfield where each bit of `map` corresponds to an `index` in the node. When an `index` is generated, the `index` bit of the `map` bitfield is checked. If the bit is not set (`0`), that index does not exist. If the bit is set (`1`), the value exists in the data elements array. To determine the index of the data elements array, we perform a bit-count (`popcount()`) on the `map` bitfield _up to_ the `index` bit to generate a `dataIndex`. In this way, the data elements array's total length is equal to `popcount(map)` (the number of bits set in all of `map`). If `map`'s bits are all set then the data elements array will be `2`<sup>`bitWidth`</sup> in length, i.e. every position will contain either a bucket or a link to a child node.
+
+Insertion of new buckets with `Set(key, value)` involves splicing in a new element to the data array at the `dataIndex` position and setting the `index` bit of the `map` bitmap. Converting a bucket to a child node leaves the `map` bit map alone as the `index` bit still indicates there is an element at that position.
+
+A `Get(key)` operation performs the same hash, `index` and `dataIndex` calculation at the root node, traversing into a bucket to find an entry matching `key` or traversing into child nodes and performing the same `index` and `dataIndex` calculation but at an offset of an additional `bitWidth` bits in the `key`'s hash.
+
+A `Delete(key)` mutation first locates the element in the same way as `Get(key)` and if that entry exists, it is removed from the bucket containing it. If the bucket is empty after deletion of the entry, we remove the bucket element completely from the data element array and unsets the `index` bit of `map`. If the node containing the deleted element has no links to child nodes and contains `bucketSize` elements after the deletion, those elements are compacted into a single bucket and placed in the parent node in place of the link to that node. We perform this check on the parent (and recursively if required), thereby transforming the tree into its most compact form, with only buckets in place of nodes that have up to `bucketSize` entries at all edges. This compaction process combined with the `key` ordering of entries in buckets produces canonical forms of the data structure for any given set of `key` / `value` pairs regardless of their insertion order or whether any intermediate entries have been added and deleted.
+
+By default, each node in an IPLD HashMap is stored in a distinct IPLD block and CIDs are used for child node links. The schema and algorithm presented here also allows for inline child nodes rather than links, with read operations able to traverse multiple nodes within a single block where they are inlined. The production of inlined IPLD HashMaps is left unspecified and users should be aware that inlining breaks canonical form guarantees.
+
+## Structure
+
+### Parameters
+
+Configurable parameters for any given IPLD HashMap:
+
+* `hashAlg`: The hash algorithm applied to keys in order to evenly distribute entries throughout the data structure. The algorithm is chosen based on speed, `digestLength` and randomness properties (but it must be available to the reader, hence the need for shared defaults, see below).
+* `bitWidth`: The number of bits to use at each level of the data structure for determining the index of the entry or a link to the next level of the data structure to continue searching. The equation `2`<sup>`bitWidth`</sup> yields the arity of the HashMap nodes, i.e. the number of storage locations for buckets and/or links to child nodes.
+* `bucketSize`: The maximum array size of entry storage buckets such that exceeding `bucketSize` causes the creation of a new child node to replace entry storage.
+
+### Node properties
+
+Each node in a HashMap data structure contains:
+
+* `data`: An Array, with a length of one to `2`<sup>`bitWidth`</sup>.
+* `map`: A bitfield, stored as Bytes, where the first `2`<sup>`bitWidth`</sup> bits are used to indicate whether a bucket or child node link is present at each possible index of the node.
+
+An important property of a HAMT is that the `data` array only contains active elements. Indexes in a node that do not contain any values (in buckets or links to child nodes) are not stored and the `map` bitfield is used to determine the `data` whether values are present and the array index of present values using a [`popcount()`](https://en.wikipedia.org/wiki/Hamming_weight). This allows us to store a maximally compacted `data` array for each node.
+
+### Schema
+
+The **root block** of an IPLD HashMap contains the same properties as all other blocks, in addition to configuration data that dictates how the algorithm below traverses and mutates the data structure.
+
+See [IPLD Schemas](/schema-layer/schemas) for a definition of this format.
+
+```ipldsch
+# Root node layout
+type HashMapRoot struct {
+  hashAlg String
+  bitWidth Int
+  bucketSize Int
+  map Bytes
+  data [ Element ]
+}
+
+# Non-root node layout
+type HashMapNode struct {
+  map Bytes
+  data [ Element ]
+}
+
+type Element union {
+  | Child "child"
+  | Bucket "bucket"
+} representation keyed
+
+type Child union {
+  | Link link
+  | HashMapNode map
+} representation kinded
+
+type Bucket list [ BucketEntry ]
+
+type BucketEntry struct {
+  key Bytes
+  value Value (implicit "null")
+} representation tuple
+
+type Value union {
+  | Bool bool
+  | String string
+  | Bytes bytes
+  | Int int
+  | Float float
+  | Map map
+  | List list
+  | Link link
+} representation kinded
+```
+
+Notes:
+
+* `hashAlg` in the root block is a string identifier for a hash algorithm. The identifier should correspond to a [multihash](https://github.com/multiformats/multihash) identifier as found in the [multiformats table](https://github.com/multiformats/multicodec/blob/master/table.csv).
+* `bitWidth` in the root block should be at least `3`.
+* `bucketSize` in the root block must be at least `1`.
+* Keys are stored in `Byte` form.
+* The size of `map` is determined by `bitWidth` since it holds one bit per possible data element. It must be `1` or `2`<sup>`bitWidth`</sup>` / 8` bytes long, whichever is largest.
+
+## Algorithm in detail
+
+### `Get(key)`
+
+1. Set a `depth` value to `0`, indicating the root block
+2. The `key` is hashed, using `hashAlg`.
+3. Take the left-most `bitWidth` bits, offset by `depth x bitWidth`, from the hash to form an `index`. At each level of the data structure, we increment the section of bits we take from the hash so that the `index` comprises a different set of bits as we move down.
+4. If the `index` bit in the node's `map` is `0`, we can be certain that the `key` does not exist in this data structure, so return an empty value (as appropriate for the implementation platform).
+5. If the `index` bit in the node's `map` is `1`, the value may exist. Perform a `popcount()` on the `map` up to `index` such that we count the number of `1` bits up to the `index` bit-position. This gives us `dataIndex`, an index in the `data` array to look up the value or insert a new bucket.
+6. If the `dataIndex` element of `data` contains a link (CID) to a child block or an inline child block, increment `depth` and repeat with the child node identified by the link from step **3**.
+7. If the `dataIndex` element of `data` contains a bucket (array), iterate through entries in the bucket:
+   1. If an entry has the `key` we are looking for, return the `value`.
+   2. If no entries contain the `key` we are looking for, return an empty value (as appropriate for the implementation platform). Note that the bucket will be sorted by `key` so a scan can stop when a scan yields keys greater than `key`.
+
+### `Set(key, value)`
+
+1. Set a `depth` value to `0`, indicating the root block
+2. The `key` is hashed, using `hashAlg`.
+3. Take the left-most `bitWidth` bits, offset by `depth x bitWidth`, from the hash to form an `index`. At each level of the data structure, we increment the section of bits we take from the hash so that the `index` comprises a different set of bits as we move down.
+4. If the `index` bit in the node's `map` is `0`, a new bucket needs to be created at the current node. If the `index` bit in the node's `map` is `1`, a value exists for this `index` in the node's `data` which may be a bucket (which may be full) or may be a link to a child node or an inline child node.
+5. Perform a `popcount()` on the `map` up to `index` such that we count the number of `1` bits up to the `index` bit-position. This gives us `dataIndex`, an index in the `data` array to look up the value or insert a new bucket.
+6. If the `index` bit in the node's `map` is `0`:
+   1. Mutate the current node (create a copy).
+   2. Insert a new element in `data` at `dataIndex` containing an new bucket (array) with a single entry for the `key` / `value` pair.
+   3. Create a CID for the mutated node.
+   4. If `depth` is `0`, the CID represents the new root block of the HashMap.
+   5. If `depth` is greater than `0`:
+      1. Mutate the node's parent
+      2. Record the new CID of the mutated child in the appropriate position of the mutated parent's `data` array. 
+      3. Recursively proceed, by recording the new CIDs of each node in a mutated copy of its parent node until `depth` of `0` where we produce the the new root block and its CID.
+7. If the `index` bit in the node's `map` is `1`:
+   1. If the `dataIndex` element of `data` contains a link (CID) to a child node or an inline child node, increment `depth`:
+      1. If `(depth x bitWidth) / 8` is now greater than the `digestLength`, a "max collisions" failure has occurred and an error state should be returned to the user.
+      2. If `(depth x bitWidth) / 8` is less than the number of bytes in the hash, repeat with the child node identified in step **3**.
+   2. If the `dataIndex` element of `data` contains a bucket (array) and the bucket's size is less than `bucketSize`:
+      1. Mutate the current node (create a copy).
+      2. Insert the `key` / `value` pair into the new bucket at a position sorted by `key` such that all entries in the bucket are ordered respective to their `key`s. This helps ensure canonical form.
+      3. Proceed to create new CIDs for the current block and each parent as per step **6.c**. until we have a new root block and its CID.
+   3. If the `dataIndex` element of `data` contains a bucket (array) and the bucket's size is `bucketSize`:
+      1. Create a new empty node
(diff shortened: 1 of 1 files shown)


mikeal:
Wow! I’ll need to set aside some time to read fully but just from the cursory read I’ve done this looks great! :)

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+* `hashAlg`: The hash algorithm applied to keys in order to evenly distribute entries throughout the data structure. The algorithm is chosen based on speed, byte-size and randomness properties (but it must be available to the reader, hence the need for shared defaults, see below).
+* `bitWidth`: The number of bits to use at each level of the data structure for determining the index of the entry or a link to the next level of the data structure to continue searching. The equation `2`<sup>`bitWidth`</sup> yields the arity of the HashMap nodes, i.e. the number of storage locations for buckets and/or links to child nodes.
+* `bucketSize`: The maximum array size of entry storage buckets such that exceeding `bucketSize` causes the creation of a new child node to replace entry storage.
+
+### Node properties
+
+Each node in a HashMap data structure contains:
+
+* `data`: An array, with a length of one to `2`<sup>`bitWidth`</sup>.
+* `map`: A bitfield where the first `2`<sup>`bitWidth`</sup> bits are used to indicate whether a bucket or child node link is present at each possible index of the node.
```
Stebalien: Specifically, stored as a byte array.

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+
+Each node in an IPLD HashMap is stored in a distinct IPLD block and CIDs are used for child node links.
+
+## Structure
+
+### Parameters
+
+Configurable parameters for any given IPLD HashMap:
+
+* `hashAlg`: The hash algorithm applied to keys in order to evenly distribute entries throughout the data structure. The algorithm is chosen based on speed, byte-size and randomness properties (but it must be available to the reader, hence the need for shared defaults, see below).
```
Stebalien: I'd say "digest length" instead of "byte size".

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+### Schema
+
+The **root block** of an IPLD HashMap contains the same properties as all other blocks, in addition to configuration data that dictates how the algorithm below traverses and mutates the data structure.
+
+See [IPLD Schemas](/schema-layer/schemas) for a definition of this format.
+
+```ipldsch
+# Root node layout
+type HashMapRoot struct {
+  hashAlg String
```
Stebalien, line 98: Any reason to use the string name over the codec number? I guess we already have a bunch of string overhead from the object keys so it _shouldn't_ be an issue.

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+  | Map map
+  | List list
+  | Link link
+} representation kinded
+```
+
+Notes:
+
+* `hashAlg` in the root block is a string name that corresponds to a [multihash](https://github.com/multiformats/multihash) as found in the [multiformats table](https://github.com/multiformats/multicodec/blob/master/table.csv).
+* `bitWidth` in the root block must be at least `2` and no more than `12`. ***(TODO: this is arbitrary and currently IAMap only supports a maximum of `8` and go-hamt-ipld _only_ supports `8`. `12` yields a maximum `data` array length of `4096`, is it reasonable to go higher? A two-byte maximum would be neat, but `65,536` elements is maybe too much.)***
```
Stebalien: WRT the cap, we quickly exceed the maximum block size of 1MiB. However, we may want to specify this as a should instead of a must.

We may also want to specify a _minimum_ of 3 instead of 2. That would make the definition of `map` a bit simpler (we'd define the length of the bitmap to be exactly `(2^bitWidth)/8` bytes.

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+* `bucketSize` in the root block must be at least `1`. ***(TODO: does it need a maximum?)***
```
Stebalien: Meh? I wouldn't overspecify. Every constraint we add to the spec is something someone will get wrong in an implementation.

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+* Values can be any kind but cannot be `Null`
```
Stebalien: Any reason _not_ to allow Null? Most map implementations allow users to distinguish between "null" and "not set".

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+}
+
+# Non-root node layout
+type HashMapNode struct {
+  map Bytes
+  data [ Element ]
+}
+
+type Element union {
+  | Link "link"
```
Stebalien: * We may want to call this a "shard" (this is a keyed union).
* I'm not sure if we should _require_ that this be a link. Like the UnixFS-2 proposal, it would be nice if we could abstract over block boundaries" (but this SHOULD be a link).
* If we abstract over block boundaries, I wonder if we even need buckets? We can literally say: If there are fewer than 3 values in a subdag, erase block boundaries. To weird?

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+## Algorithm in detail
+
+### `Get(key)`
+
+1. Set a `depth` value to `0`, indicating the root block
+2. The `key` is hashed, using `hashAlg`.
+3. Take the left-most `bitWidth` bits, offset by `depth x bitWidth`, from the hash to form an `index`. At each level of the data structure, we increment the section of bits we take from the hash so that the `index` comprises a different set of bits as we move down.
+4. If the `index` bit in the node's `map` is `0`, we can be certain that the `key` does not exist in this data structure, so return a `Null` value.
+5. If the `index` bit in the node's `map` is `1`, the value may exist. Perform a `popcount()` on the `map` up to `index` such that we count the number of `1` bits up to the `index` bit-position. This gives us `dataIndex`, an index in the `data` array to look up the value or insert a new bucket.
+6. If the `dataIndex` element of `data` contains a link (CID) to a child block, increment `depth` and repeat with the child node identified by the link from step **3**.
```
Stebalien: (keyed union)

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,292 @@
...
+7. If the `dataIndex` element of `data` contains a bucket (array), iterate through entries in the bucket:
+   1. If an entry has the `key` we are looking for, return the `value`.
+   2. If no entries contain the `key` we are looking for, return a `Null` value. Note that the bucket will be sorted by `key` so a scan can stop when a scan yields keys greater than `key`.
+
+### `Set(key, value)`
+
```
Stebalien, line 159: May want to bring up what happens when we run out of bits in the hash digest. IMO, we should just fail (HAMTs with untrusted writers should use a secure hash algorithm).

schema-layer/data-structures/hashmap.md
warpfork: +1 not supporting null will be problematic; the usual understanding in the data model is that a map can contain null values.

Schemas flip the default (nullability must be opt-in), but even there it would be a little startling to have some advanced layouts which aren't allowed to be used if the nullable modifier is on the value type.

schema-layer/data-structures/hashmap.md
warpfork: +1 to 'should' over 'must' if there's any discussion of message/block size in bytes.  IPLD doesn't have a strict depends-on relationship to libp2p, and it wouldn't be correct to take any current constants about their message size limits as gospel for IPLD protocol specifications.  (But making a practical note about it seems fine and constructive to me.)

schema-layer/data-structures/hashmap.md
rvagg: ok, I've gone with the [kinds](http://github.com/ipld/specs//data-model-layer/data-model.md#kinds) naming:

* `data`: an Array, with ...
* `map`: A bitfield, stored as Bytes, where ...

schema-layer/data-structures/hashmap.md
rvagg: I originally did use the multihash varint for this but I got sick of the terseness and indirection for very little gain.
I could switch back if someone has strong feelings about this but my default preference is for these data structures to be _somewhat_ readable on their own—at for the debugging case, to avoid indirection as much as possible (both code indirection "what is this varint again?" and cognitive indirection, "what's `p` doing again?").

schema-layer/data-structures/hashmap.md
rvagg: Great point on the minimum of `3`, any problem with that being a "must"?

Regarding a maximum here, how do we handle the interop problem of one implemetation producing data that may be operating outside of the bounds of another? If the JS version can do a bitWidth of 12 but the Go only a maximum of 8? That's where my thoughts on "must" are coming from.

The alternative is to say that "should be at least" and "behavior beyond is this is not defined in this specification" which I've done elsewhere in here. So we have a well defined interop space, beyond which, caveat emptor. If we do that here, what upper bound of `bitWidth` for defined behavior (i.e. "commonly supported") do we specify? `8` is pretty small and limits your ability to pack data into blocks, but `16` is massive—keeping in mind in common use we will multiply "data elements" by at least `size(CID)` to get block size.

| `bitWidth` | data elements | `map` size |
|--|--|--|
| 3 | 8 | 1 byte |
| 4 | 16 | 2 bytes |
| 5 | 32 | 4 bytes |
| 6 | 64 | 8 bytes |
| 7 | 128 | 16 bytes |
| 8 | 256 | 32 bytes |
| 9 | 512 | 64 bytes |
| 10 | 1024 | 128 bytes |
| 11 | 2048 | 256 bytes |
| 12 | 4096 | 512 bytes |
| 13 | 8192 | 1024 bytes |
| 14 | 16384 | 2048 bytes |
| 15 | 32768 | 4096 bytes |
| 16 | 65536 | 8192 bytes |

schema-layer/data-structures/hashmap.md
rvagg: 👍 but the question of well-defined interop space may be important here too, see above about `bitWidth`. Do you want people showing up in the Go repo saying "I made this HashMap in my browser but the Go implementation won't read it? Being able to fall back to saying "you pushed `bucketSize` beyond `<well defined point>` and you hit caveat emptor space, sorry" might be nice. Or do we just let those points evolve and each implementation document their limits?

schema-layer/data-structures/hashmap.md
rvagg: OK, I can live with that. I've built `null` as a signal into a lot of JS implementation but that should probably have been `undefined`.

The schema can probably become:

```
type BucketEntry struct {
  key Bytes
  value Value (implicit Null)
} representation tuple
```

schema-layer/data-structures/hashmap.md
rvagg: I'm not a fan of the term "shard" but am not strongly tied to "link" either.

@mikeal you've talked a lot about inlining entire UnixFSv2 structures into a single block, does that requirement extend to the case where you're needing a HAMT to store directories, or is there some upper size bound where that's not allowed? If you need this data structure to be inlinable then yes, we'll need to do something about this "link" piece.

In terms of having a heuristic for automatic inlining of child nodes, (a) that would make the case for buckets harder to sustain so they'd probably need to go and (b) we need to figure out how to retain canonical form. Is it as easy as applying the same rules as applied here to buckets? "if child node has <= `X`, inline it"?

There's a broader issue here of how to pack tree structures into blocks that I keep bumping into with collections, particularly sorted collections, but that might be a discussion for elsewhere.

schema-layer/data-structures/hashmap.md
rvagg: I've gone with:

> If `(depth x bitWidth) / 8` is now greater than the `digestLength`, a "max collisions" failure has occurred and an error state should be returned to the user.

schema-layer/data-structures/hashmap.md
rvagg: I have two variants scattered in this section: "contains a link (CID)" and "contains a bucket (array)", assuming that the reader will identify that the discriminator is in the keyed union. Should it be something like "contains a 'link' property" and "contains a 'bucket' property"?

rvagg:
latest update inserts "`digestLength`" in place of "`hashBytes`" but it still uses the term "hash", I'd appreciate sanity checking that this is reasonable or if we should just standardise on one of them: "digest" or "hash" or "hash output". I'm not a fan of "digest" but it's not an unreasonable choice.

schema-layer/data-structures/hashmap.md
Stebalien: It should be fine, we just need to make sure we don't change the names of these codecs. I wonder if we should come up with a namespaced naming scheme and do a mass-rename first.

schema-layer/data-structures/hashmap.md
mikeal: I’ve taken a note to add an explicit “not found” message from `get` functions in IPLD Generics.

schema-layer/data-structures/hashmap.md
mikeal: ### For Reads

If you avoid defining where links are and build the read side entirely path based you end up with the flexibility to inline wherever you want. With IPLD Generics this concept ends up extending to all of the types, so even something like binary data could be a single inlined binary object or a big multi-block binary object. 

### For Writes

The heuristics here are a bit tricky. You don’t have *great* knowledge about the eventual size of the block before you encode it in order to decide when to inline and when to not inline. 

For files in the experimental `unixfsv2` implementation we use the `size` of the underlying binary data, and target about half the max-block size we’re looking for, and this `size` property accumulates and bubbles up through the directories as well (but we don’t currently shard directories in a HAMT). The obvious problem with this method is that if you have a lot of small files with large file names the encoding overhead is much larger than the size.

But, since the read size is completely agnostic of block boundaries we have room to improve all of this logic in the future.

schema-layer/data-structures/hashmap.md
Stebalien: > Regarding a maximum here, how do we handle the interop problem of one implemetation producing data that may be operating outside of the bounds of another? If the JS version can do a bitWidth of 12 but the Go only a maximum of 8? That's where my thoughts on "must" are coming from.

IMO, that only gets worse when specifying "MUST". If we say SHOULD, implementations won't depend on it. If we say MUST, some implementations will depend on it, and some simply won't check.

At the end of the day, the real issue is orthogonal: If a block size gets above 1MiB, many IPFS implementations will reject it. Even if we say that bitWidth MUST be less than 12, I can set the bucket size to 1000 and run into the same problem.

schema-layer/data-structures/hashmap.md
Stebalien: You're right, we should definitely recommend a maximum. I just don't want implementations _rejecting_ or failing to handle HAMTs with large buckets (even if the block size is reasonable). (i.e., SHOULD versus MUST)

schema-layer/data-structures/hashmap.md
Stebalien: We should avoid making decisions in IPLD because $lang.

schema-layer/data-structures/hashmap.md
rvagg: OK so for the purpose of this spec we still have to define `Link` in the schema even if it's a union of some kind (I know you don't like that @mikeal but we need to be able to say clearly where these kinds are in a block and a Link is an important one) but we should allow for fully inline data structures too.

Currently we have:

```ipldsch
type Element union {
  | Link "link"
  | Bucket "bucket"
} representation keyed
```

I'm going to have to involve @warpfork on this one because we lack the ability to define that one named property could be either a kind (Link) or a complex map representation of some other type (HashMapNode), so I can't just rename "link" to "child" and stick either a Link or HashMapNode in there, I have to do something like this:

```ipldsch
type Element union {
  | Child "child"
  | Bucket "bucket"
} representation keyed

type Child union {
  | Link "link"
  | HashMapNode "inline"
} representation keyed
```

So you end up with either:

```
{
  ...
  "data": [
    ... buckets and other child nodes here
    "child": {
      "link": CID
    },
    ... buckets and other child nodes here
}
```

or

```
{
  ...
  "data": [
    ... buckets and other child nodes here
    "child": {
      "node": {
        ... actual inline child node here
      }
    },
    ... buckets and other child nodes here
}
```

The other options we have are kinded (HashMapNode isn't a kind, it's a complex map representation), inline (won't work where one of these things is a kind, i.e. Link, it can only inline two complex map representations with a discriminator key), and envelope (would work but requires a discriminator key in addition to the child key). See docs from [here](https://github.com/ipld/specs/blob/master/schema-layer/schemas/representations.md#union-keyed-representation-example) down.

So we can't do a union of: `"child": CID` and `"child": { ... HashMapNode here ... }`. Which I know will bother at least @mikeal because it doesn't make block boundaries "transparent". But that's @warpfork's argument to have.

rvagg: @warpfork could this particular case could be solved by treating `HashMapNode` as a `map` for a kinded union?

```
type Child union {
  | Link link
  | HashMapNode map
} representation kinded
```

making

```
    "child": CID
```

or

```
    "child": { ... HashMapNode here }
```

is that approach a general solution to the "everything could be a link" desire?

warpfork: you beat me to the punch.  yes, we should be able to do exactly such an expression with a kinded union.

schema-layer/data-structures/hashmap.md
rvagg: cool, I'll just leave it out then, so it's just:

> `bitWidth` in the root block should be at least `3`

schema-layer/data-structures/hashmap.md
rvagg: > We should avoid making decisions in IPLD because $lang.

A mindset I'm still adapting to tbh. Happy to be kept in check on it of course.

I've removed this limitation so values can now be `Null` so the schema for an entry is:

```ipldsch
type BucketEntry struct {
  key Bytes
  value Value (implicit "null")
} representation tuple
```

which looks like `[ "akey": "avalue" ]` for a standard key/value pair, and `[ "akeywithnull" ]` for a key/null pair.

The docs also had 2 other places that mentioned returning `Null` if the value wasn't found, I've replaced those with: "return an empty value (as appropriate for the implementation platform)".

schema-layer/data-structures/hashmap.md
rvagg: Here's what I've done for both `bitWidth` and `bucketSize`. In this section I've just specified the minimums ("should" for `bitWidth`, "must" for `bucketSize` because `<=0` doesn't make sense). Down below in the "Implementation Defaults" section I've added this type of thing to them both: "No maximum is specified, however implementers should be aware that interoperability problems may arise with very large `bucketSize` values."

schema-layer/data-structures/hashmap.md
rvagg: OK, I've gone with that kinded form above, it's neat. I've also removed or altered explicit references to CIDs for read operations such that they could be inline. But I've also added this to the bottom of the summary in the top section of this doc:

> By default, each node in an IPLD HashMap is stored in a distinct IPLD block and CIDs are used for child node links. The schema and algorithm presented here also allows for inline child nodes rather than links, with read operations able to traverse multiple nodes within a single block where they are inlined. The production of inlined IPLD HashMaps is left unspecified and users should be aware that inlining breaks canonical form guarantees.

i.e. Inline is possible and reads should assume it's possible. How you write them and under what circumstances are up to you and they shouldn't be default mode. See 73ab06b for the precise changes I made on this point. Reviews appreciated.

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,298 @@
...
+These defaults are descriptive rather than prescriptive. New implementations may opt for different defaults, while acknowledging that they will produce different graphs (and therefore CIDs) for the same data as with the defaults listed below. Users may also be provided with facilities to override these defaults to suit their use cases where these defaults do not produce optimal outcomes.
+
+### `hashAlg`
+
+* The default supported hash algorithm for writing IPLD HashMaps is the x64 form of the 128-bit [MurmurHash3](https://github.com/aappleby/smhasher), identified by the multihash name ['murmur3-128'](https://github.com/multiformats/multicodec/blob/master/table.csv). Note the x86 form will produce different output so should not be confused with the x64 form. Additionally, [some JavaScript implementations](https://cimi.io/murmurhash3js-revisited/) do not correctly decompose UTF-8 strings into their constituent bytes for hashing so will not produce portable results.
```
dignifiedquire: The current go implementation relies on the serialization of the two resulting `u64`s, in a specific endianess. This needs to be specced, as it is _not_ part of the original reference implementation. In addition, given that there is no spec it should be clear how edge cases like  https://github.com/spaolacci/murmur3/issues/11  should be handled.

Ref for serialization: https://github.com/spaolacci/murmur3/blob/master/murmur128.go#L52-L61

dignifiedquire: Also the go-hamt-ipld implementation does not use a multihash, but the raw murmur3 hash, this sholud be probably clarified, as the reference to multihash is confusing in this case.

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,298 @@
...
+
+By default, each node in an IPLD HashMap is stored in a distinct IPLD block and CIDs are used for child node links. The schema and algorithm presented here also allows for inline child nodes rather than links, with read operations able to traverse multiple nodes within a single block where they are inlined. The production of inlined IPLD HashMaps is left unspecified and users should be aware that inlining breaks canonical form guarantees.
+
+## Structure
+
+### Parameters
+
+Configurable parameters for any given IPLD HashMap:
```
dignifiedquire, line 74: I would really like to see these being optional, such that we don't have to add the overhead of storing these in every hamt that gets serialized (especially for smaller structures)

dignifiedquire: bonus: if these become optional a regular node and a root node could be the same data structure, allowing for some interesting nesting

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,298 @@
...
+
+See [IPLD Schemas](/schema-layer/schemas) for a definition of this format.
+
+```ipldsch
+# Root node layout
+type HashMapRoot struct {
+  hashAlg String
+  bitWidth Int
+  bucketSize Int
+  map Bytes
```
dignifiedquire, line 101: Shouldn't this be a fixed size?

schema-layer/data-structures/hashmap.md
rvagg: @dignifiedquire: re the reference to "multihash" here is simply as a string identifier that goes into the root block under the property "hashAlg", with the intention that any alternative hash algorithm found at "hashAlg" will also be a valid identifier in the multihash table, that's all. e.g. https://github.com/rvagg/iamap/blob/ec2259f5f6108954d511c688aafb91273dc50828/iamap.js#L92-L94 is the only place I use multihash at all in my implementaiton, just a check that it's valid. What I've done is clarified it a bit more at the top of the doc where it first references "multihash" and then put this reference in brackets: _(identified by the multihash name ...)_.

Regarding the specifics of how Go implements multihash and the edge cases, can you help me out with some wording for this? I've had a bunch of trouble with murmur3 across platforms already and it's making me a bit weary of this area.

My go-to JavaScript implementation does some [awkward concat-byte-to-hex work](https://github.com/cimi/murmurhash3js-revisited/blob/de8d6862b5b6507a96c0114547be7944dedd8d86/lib/murmurHash3js.js#L543) which I then have to unpack into bytes again. I have no confidence that the values I'm using will end up being cross platform suitable! I even gave up trying to generate compatible murmur3 hashes in JavaScript as produced by the Scala-to-JavaScript version in [@magik6k's distributed-wiki-search](https://github.com/magik6k/distributed-wiki-search/blob/de58727c3a03476627a9c6eb012d45bf5b4e0e07/client/src/main/scala/WikiSearch.scala#L66) when I was playing with that. Perhaps it's the wrong hash to be defaulting to?

schema-layer/data-structures/hashmap.md
rvagg: @dignifiedquire: optional means we're going to need to be more confident of the default values and their future suitability. I don't think we have that, we don't know how these data structures will be used by real-world applications or what the implications of the default shapes we're defining here are. For example, I have a suspicion that buckets might not be the right approach and that the elasticity in block sizes they will yield might be too messy. But I can't make assertions on that without real world applications and a lot of research work.

My preference would be to stick with explicit definitions for now, and maybe over time it'll be clear what the right parameters are an define implicit defaults.

> allowing for some interesting nesting

Not really though, depth changes everything, a depth=0 node (root) is useless at depth=>0. Do you have something concrete in mind?

schema-layer/data-structures/hashmap.md
rvagg: @dignifiedquire: it's dependent on `bitWidth` but @warpfork has been very firm on his desire to _not_ introduce dependent types into IPLD schemas so we can't represent that here. 

I've added this to the notes below the schema:

> * The size of `map` is determined by `bitWidth` since it holds one bit per possible data element. It must be `1` or `2`<sup>`bitWidth`</sup>` / 8` bytes long, whichever is largest.

schema-layer/data-structures/hashmap.md
dignifiedquire: > Do you have something concrete in mind?

no, just a random idea

> My preference would be to stick with explicit definitions for now

fair enough

schema-layer/data-structures/hashmap.md
dignifiedquire: I am confused, is this a different indexing scheme than used in go-hamt-ipld? 
Because looking at https://github.com/ipfs/go-hamt-ipld/blob/master/hamt.go#L73-L74 it needs to always be of size 256 bits.

schema-layer/data-structures/hashmap.md
dignifiedquire: Soooo murmur3 is complicated at best, under specified at worst I would argue. But it does work nicely overall and is fast. 

- In regards to compatible versions: I have this rust version https://github.com/dignifiedquire/murmur3/blob/nicer-hashing/src/murmur3_x64_128.rs that has fuzz tests to match against the reference implementation + now matches the bytes of the go implementation for all cases I tried so far. This might be useful to use in JS land as well. 
- If we want to really rely on murmur3 here, I think we should write a spec, for it based on the reference implementation + generate a set of test vectors
  - the minimum here I would say is:
    - define the (byte) order of how the two `u64`s produced by the reference implementation are supposed to be written to bytes
    - say sth like "in case of differences, the reference implementation is to be taken as the correct version, unless a serious bug is discovered" 
- Do we have concrete alternatives in the pipeline we could use, that provide similar speeds?

schema-layer/data-structures/hashmap.md
rvagg: Yes, this spec is not intended to reflect go-hamt-ipld. It has a flexible `bitWidth`, the go-hamt-ipld data structure is fixed to 8-bits for simplicity. Fixing to `8` gives a fixed `map` size of `32` bytes and a maximum number of data elements of `256`. The reference implementation for the purposes of writing this spec is https://github.com/rvagg/iamap (it's currently missing the ability to deal with inlined child nodes).

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,299 @@
...
+} representation kinded
+
+type Bucket list [ BucketEntry ]
+
+type BucketEntry struct {
+  key Bytes
+  value Value (implicit "null")
+} representation tuple
+
+type Value union {
```
warpfork, line 124: This whole thing might be possible to replace with the `Any` type.

(Which... needs specification.  But it's come up more than a few times that we need a way to go back to free-for-all untyped data land, and when that's been happening, I've been calling it `Any`.  So I'd say feel free to do so here, too -- and we should go sort out a full spec for `Any` soon.)

schema-layer/data-structures/hashmap.md
warpfork: I'm perfectly happy with that phrasing.

Using a kinded union, as you've done, makes it super clear that we've got the possibility for "inline" vs separate linked blocks -- A+ good.

I also think -- if you want to -- you could get even more prescriptive than that text does:

(With the caveat that refining our exact language and understandings of this has been tricky...)  My best thinking at present is that having "implicit links" in *general* was problematic (as it made canonicality more or less undefined)... but, we're not "in general", here; we're in the scope of an advanced layout.

Inside an advanced layout, we can set the definition of canonicality to whatever we want.  Because it's already scoped by the defn of the advanced layout, and since an advanced layout defn already includes descriptions of algorithms/decisions... it therefore follows that adding another decision description for whether or not to inline-vs-link isn't out of character at all; it fits right along with the rest.

In other words -- as long as the decision to inline-vs-link happens *inside* the scope of the advanced layout, rather than being unspecified and left to some higher level (aka, made into a problem for the user), we're good!

schema-layer/data-structures/hashmap.md
rvagg: 👍 my original draft used `any` as a placeholder but since we don't have that defined yet ...

schema-layer/data-structures/hashmap.md
whyrusleeping: I don't even think we need to put them in there. We already have to have context to know that this is a HAMT and how to traverse it. might as well say that we need context to know the width.

Additionally, we can infer the width from the bitfield.

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,300 @@
...
+}
+
+# Non-root node layout
+type HashMapNode struct {
+  map Bytes
+  data [ Element ]
+}
+
+type Element union {
+  | Child "child"
```
whyrusleeping: would prefer this to be:
```
type Element union {
  | Child "c"
  | Bucket "b"
} representation keyed
```

For brevity

schema-layer/data-structures/hashmap.md
rvagg: Well, that goes back to the discussion @ https://github.com/ipld/specs/issues/130 I think so I'd rather that all be resolved before I remove these. Right now they're useful since we have no formal way of defining "context" except by inlining it. Fat pointers? CIDv2? We have neither.

The current experiment that I'm working with, the only thing that currently leverages these advanced layouts / generics / composites, is @mikeal's js-types which embeds a `_type` property that points to the implementation. Search for `_type` in https://github.com/ipld/js-generics/pull/1/files to see it embedded in blocks as they are encoded for a HashMap. The js-types implementation then knows when it encounters a `_type` at the root of a block that it can go and find the implementation that matches that string and traverse the advanced layout / generic / composite with the associated logic.

I understand this may not be ideal, but it works at least and in lieu of an alternative I'm not sure how to make this useful without hardwiring all of these parameters into code that creates and reads these blocks.

Good point re `width`, I'll try removing that from my implementation and update here.

schema-layer/data-structures/hashmap.md
rvagg: 👍

schema-layer/data-structures/hashmap.md
mikeal: So, we shouldn’t need an identifier like `_type` in *this* spec. For the (advanced layouts, / generics / composites) I’m separating out the “definition,” which describes the implementation of the data structure, from the actual data. Everything in this spec, including these parameters, are part of the data.

It’s important to keep these separate because we’re exploring in-band **and** out-of-band signaling mechanisms. Since the “definition” can be applied in multiple ways we need to make sure that definitions stay out of specs like this. As @whyrusleeping notes, there are cases where you already know what data structure you’re working with and don’t even need an in-band description.

The experimental implementation we have will change once we figure out how the “fat pointer” works and will only read the `_type` field when we encounter the right kind of link *or* when you explicitly tell it the data you’re passing is a definition.

rvagg: I'll admit that I don't fully understand this desire to remove this kind of data from blocks. I've been imagining that making blocks as self describing as possible was a worthy goal. There are obvious limitations to that but CIDs seem to get us over most of those. It seems to me that once you have a CID for _something_ that's all you ideally should need to understand what's in it—minus all of the programatic pieces to decode and navigate through it (codecs and these advanced layout pieces). There seems to be a desire to push a lot of that self describing stuff into the "how we got to here" place. That's the bit I don't get.

mikeal: This got a little off-track, let me try and float a proposal that I think can resolve it.

In our experimental composite system, we could have something like this:

```javascript
{ _type: ‘IPLD/Experimental/HAMT/0’,
  hamt: { 
    hashAlg: ‘sha2-256’,
    bitWidth: 8,
    bucketSize: 3,
    root: {
      data: [],
      map: Bytes(),
    }
  }
}
```  

The top level node containing `_type` and `hamt` are part of the “definition” which isn’t part of this spec and is only there for illustrative purposes.

The root block of this spec is then modified to put `bytes` and `map` under a `root` key. In the schema we can define this as a union of Map or Link. In the Filecoin use case, they’d just stick to encoding the root node all the way down. For everyone else, including people linking to Filecoin’s hamt, they can create a block containing the parameters and then link to this root block.

This gives us the flexibility to avoid encoding parameters which are constant in the application while still allowing people to create pointers to the data containing the parameter context they need in order to read the structure properly. The only thing we need to make sure to note is that if you inline the Map vs Link to a Map you’re going to end up with a different hash.

rvagg:
re the outstanding point about the `Element` union

```
type Element union {
  | Child "child"
  | Bucket "bucket"
} representation keyed
```

becoming "c" and "b" for brevity: I'm moving my code to a `kinded` union between `link` and `array` such that the `Child` will be just a CID and the `bucket` will be an array of key/value tuples, you switch depending on which one of these you find. So no need for keys or extra nesting.

But then we get some confusion when we handle the inline case where `Child` could be inlined:

```
type Element union {
  | Child what?
  | Bucket list
} representation kinded

type Child union {
  | Link link
  | HashMapNode map
} representation kinded
```

I might have to introduce `Child` and `ChildInlined` into the first kinded set to make this work with current schemas.

(Edit: for clarity, "kinded" would work for `ChildInlined` as a `map` and `Child` as a `link` with `Bucket` being a `list`, but separating out the inlined vs linked `Child` seems .. inelegant?)

warpfork:
I don't know if we've hit this syntactical corner before (!) but I think the following should be a sensible expression that might do what you want:

```
type Element union {
  | Child map
  | &Child link
  | Bucket list
} representation kinded
```

Does that seem to resonate at all rightly?

(I don't know if I've sufficiently fully wrapped my head around the problem space, but I wonder if this might make it possible to drop one of the types entirely?  I don't *quite* fully grok BucketEntry/Bucket/Child/Element/Value/Root and I've got this vague idea that it seems like maybe one more type than is necessary.  But I'm not at all prepared to back that feeling up and could easily be wrong.)

rvagg:
@warpfork thanks, you're right that it does collapse one type from the list, `Child` can go away, leaving:

```ipldsch
type Element union {
  | HashMapNode map
  | &HashMapNode link
  | Bucket list
} representation kinded
```

Full schema with this change now at https://github.com/ipld/specs/blob/b4814bba65df1fcf1a067a5da7e27e4a222653e3/schema-layer/data-structures/hashmap.md#schema, there's an outstanding TODO of removing `bitWidth` and using `map` length to determine that and also `Value` is basically an `Any` union which we may want in schemas.

warpfork:
Awesome! :tada: 

I also re-read the other types and now they all make sense to me.  (I ground my gears for a second figuring out how different leafward types were and why, though.  Maybe a line or two of very terse comment in the schemas about each type's role would help?  Just enough to back out to the bigger document.)

schema-layer/data-structures/hashmap.md
```diff
@@ -0,0 +1,296 @@
...
+
+An implementation should only emit any given `key`, `value` or `key` / `value` entry pair once per iteration.
+
+### Differences to CHAMP
+
+This algorithm differs from CHAMP in the following ways:
+
+1. CHAMP separates `map` into `datamap` and `nodemap` for referencing local data elements and local references to child nodes. The `data` array is then split in half such that data elements are stored from the left and the child node links are stored from the right with a reverse index. This allows important speed and cache-locality optimizations for fully in-memory data structures but those optimizations are not present, or make negligible impact in a distributed data structure.
+2. CHAMP does not make use, of buckets, nor do common implementations of HAMTs on the JVM (e.g. Clojure, Scala, Java). Storing only entries and links removes the need for an iterative search and compare within buckets and allows direct traversal to the entries required. This is effective for in-memory data structures but is less useful when performing block-by-block traversal with distributed data structures where packing data to reduce traversals may be more important.
```
warpfork, line 233: Stray comma after "use".

PR #132, opened 2019-06-20
An exploration of unixfsv2, schemas, and the intersection

warpfork:
Like the top of the document inside says: This document is an exploration of what a hypothetical "unixfs v2" might look like, using schemas as a design language for describing the data structures.

This is not a final proposal, but rather a lens to explore both unixfsv2
concepts and schemas themselves at the same time, and how we expect them
to contribute to each other.  Several concepts in schemas, advanced layouts,
their interactions, and their syntaxes will be trialed here; they are not final.

(This is opening up a whole new directory tree, the idea of which is
roughly that "it would be nice to collect some experience reports and
user stories, *before* writing final specs"; and we might still want to
be able to easily refer and link to them and accumulate more of them
while we work towards being ready to make final specs proposals.)

(There's correspondingly a series of "in the opinion of this author"
text in this file.  The idea is that other folks might write very
different explorations, and we might merge and retain both of them.)

(I'd actually like to re-issue https://github.com/ipld/specs/issues/130
as a file in this sort of a format.  One can't link to headings in
a github issue; one *can* link to headings in a markdown file, like
the one this diff is adding.  Maybe other organizational formats could
make even better trades, but this deficiency of github issues is
worth noting.)

### DIFF
--- a/design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
+++ b/design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
@@ -1,0 +1,296 @@
+A unixfsv2 spike, using schemas
+===============================
+
+This document is an exploration of what a hypothetical "unixfs v2" might look like,
+using schemas as a design language for describing the data structures.
+
+This is not a final proposal, but rather a lens to explore both unixfsv2
+concepts and schemas themselves at the same time, and how we expect them
+to contribute to each other.  Several concepts in schemas, advanced layouts,
+their interactions, and their syntaxes will be trialled here; they are not final.
+
+Without further ado...
+
+Schema
+------
+
+```
+type AnyFile union {
+	| File "f"
+	| Dir "d"
+	| Symlink "s"
+} representation keyed
+
+type Filename string
+
+type Dir struct {
+	members {Filename:DirEnt}<HAMT>
+}
+
+type DirEnt struct {
+	content &AnyFile
+	attribs Attribs
+}
+
+type File struct {
+	content Bytes<Rabin>
+}
+
+type Symlink struct {
+	target String
+}
+
+advanced Rabin {
+	kind bytes
+	implementation "IPLD/experimental/rabin/v1"
+}
+
+advanced HAMT {
+	kind map
+	implementation "IPLD/experimental/HAMT/v1"
+}
+
+type Attribs struct {
+	mtime Int # Danger danger will robinson; int sizes will rear their ugly head.
+	posix Int # Can we just... do the standard 0777 mask packing here?  I don't think anything else is *less* work.
+	sticky Bool (implicit: false)
+	setuid Bool (implicit: false)
+	setgid Bool (implicit: false)
+	uid Int # make another schema which deliberately excludes this.  try to match this one (the one with more info) first.
+	gid Int # make another schema which deliberately excludes this.  try to match this one (the one with more info) first.
+}
+
+# You should be able to imagine sending a fancy graphsync request with some structure like:
+#
+#  {"schema": {{schemaCIDofUnixfs}}, "selector": {{selectorSpec}},
+#   "schemaPatch": {"./dir/transparently/through/to/file": {{schemaCIDofRabinLayout}},
+#   "startAt": {{rootBlockCID+path}} }
+#
+# This kind of fancy graphsync request can be used to make a query which does "the right thing" with
+# all advanced layouts transparently; then when reaching the 'patch' area, perhaps the selector spec
+# wants to use the seeded-psuedorandomness feature, so it can issue queries like this across many peers.
+#
+# Perhaps we won't actually need the 'patch' feature, because in practice if doing queries with lots
+# of fancy sharded requests, prooobably you'll do the path *up to* that first, in a separate query,
+# But either way, you can see how using schemas for signaling is useful here.
+```
+
+Interesting Highlights
+----------------------
+
+### Using various schemas as a feature!
+
+Observe the comments on the `Attribs.uid` and `Attribs.gid` fields.
+
+By using several different (but significantly overlapping) schemas to describe variations
+in which attributes we're storing, we can make unixfsv2 into an extensible yet still
+very well defined system.
+
+Concretely: If the `uid` and `gid` elements are found present in the first `Attribs` map
+processed in a tree of data, then we match the first schema (the one with more info),
+and we can now expect *all future* `Attribs` maps to consistently include the same data.
+If the `uid` or `gid` elements are found to be missing, we would flunk out of matching
+the first schema, and try a second schema which does not include them... and again,
+now expect *all future* `Attribs` maps to consistently *not* include these data.
+
+This kind of deep consistency makes it easier to build applications,
+especially applications which may want to change their behavior
+or apply distinct transformations depending on which of the schemas matches.
+
+### Advanced layout usage sites are marked with angle brackets
+
+E.g., on the line `members {Filename:DirEnt}<HAMT>`, it's the `<` and `>`
+characters which denote that we'll be using an advanced layout here.
+
+We might also consider using different characters for this denotation.
+(Angle brackets are reminiscent of generics in Java syntax;
+this may or may not be a helpful reminiscence.)
+
+We might also consider a different approach entirely: ditching the ability to
+declare advanced layouts on inline type terms at all, and instead using
+something more like the `representation` keyword.
+Such an approach would syntactically require a fully named type for any
+structure that wants to use an advanced layout.
+Given that the most common applications of advanced layouts are for map, list,
+and byte kinds, all of which tend to be used via anonymous inline types,
+this could result in a substantial increase in syntax volume
+(e.g., in this example, it would've lead to two new named types).
+
+(Note that the previous paragraph has only considered schema *DSL syntax*;
+perhaps the issue looks differently if considered via the IPLD-native "AST".)
+
+### Advanced layout usage sites are marked
+
+See https://github.com/ipld/specs/issues/130 -- specifically,
+the "signaling" problem -- for more discussion of this.
+
+We're exploring the "explicit signaling" approach, and doing so
+using schemas as the vehicle, here.
+
+### Advanced layouts promising a kind?
+
+The `kind map` and `kind bytes` markings in the `advanced` blocks are a hypothesis.
+
+In the cases shown in this example, we definitely know that these are the kinds
+we expect the advanced layout algorithms in question to behave as.
+This may not always be true (e.g. in the encryption user story, not explored here).
+
+The kind is also indicated at the usage site -- e.g. the `{` character
+at the start of the usage site `members {Filename:DirEnt}<HAMT>` already
+makes it quite clear that we're dealing with a map kind -- so perhaps
+kind indicators in the `advanced` block is redundant.  Or, perhaps by the
+other side of the same coin, they're a good sanity-check landmark.
+
+### Advanced layout implementation strings
+
+See https://github.com/ipld/specs/issues/130 -- specifically,
+the "referencing" (or "rendezvous") problem -- for more discussion of this.
+
+We're exploring with the "behavioral specification" approach here.
+
+### What are executables, anyway?
+
+In this schema, the executablity bit is in the `Attribs.posix` bitfield...
+as three distinct bools, once per user/group/everyone (the `0111` mask).
+
+Another way to regard executablity which would often (often!) be valid is
+to have a single bool per file.
+
+These are two different opinions application design might take in regarding
+unix*ish* filesystems.  Logically consistent arguments can be made for both.
+Perhaps this is another situation where two different schemas would be useful;
+we can support both, and do so clearly, if we think it's useful to do so.
+
+### Ugh, int sizes
+
+> mtime Int
+
+In designing the IPLD Data Model itself, we still have some deferred decisions
+about how we want to regard integer precision (as well as what we expect,
+in practice, when client implementations in various languages hit their
+host language's natural limitations).
+
+We should probably give particular attention to timestamps in such decisions.
+They're a very common user story.
+
+Additionally, regardless of what precisely we decide to do about integer sizes,
+we should aim to have documentation with recommendations of best practices
+for the timestamp user story.
+
+### Links
+
+The `&` syntax seen in `content &AnyFile` denotes that the `AnyFile` is
+expected on the far side of a Link in that position.
+
+Denotation of the position of a link is important because It Matters To The Hash,
+as well as mattering to the overall understanding of topology of the larger DAG;
+yet it's terse because it doesn't matter to the semantics of the design (much).
+
+The single character sigil would seem to imply that there's no such thing as an `&&Foo` or `&&&Foo`.
+Which indeed I hope there isn't: this author can't think of any situations
+where that expresses a sane thing.  (It's rare enough in programming with pointers;
+it's an entirely separate thing to propose in a merkle-DAG, which has no mutablity
+semantics in the first place, which removes even the 'rare' excuses.)
+We haven't officially explored or documented this anywhere yet.  Probably we should.
+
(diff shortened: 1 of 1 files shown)


design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Do advanced layouts deserve parameters?
+
+We could imagine having blocks with details such as this in the schema:
+
+```
+advanced Rabin {
+	kind bytes
+	implementation "IPLD/experimental/rabin/v1"
+	bitmatch 14
```
vmx, line 219: I would put the actual values in the root block and having the schema define only the types.

```ipldsch
type RabinParameters struct  {
    bitmatch Int (implicit: 14)
    minchunksize Int (implicit: 256)
    maxchunksize Int (implicit: 262144)
}

advanced Rabin {
    kind bytes
    implementation "IPLD/experimental/rabin/v1"
    parameters RabinParameters
}
```

vmx (commented):
That reads really good.

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
warpfork: Oh, that's...  wow, that's almost "duh" now that you've said it.  I like that.  We have a system for assigning implicit/default values: it makes perfect sense to use it.

The one tweak I'd make to that is... the `type RabinParameters struct {...}` declaration would make more sense to put *inside the ADL's* schema -- a.k.a. [hereish, in the example of the spec Rod's working on](https://github.com/ipld/specs/pull/131/files#r295781918), I think?  But... hrm, no, that doesn't make it any clearer where another schema that just wants to use an existing thing would configure it from.

warpfork: I like this idea, but also don't intend to try to reach final consensus and resolution on this topic today.  This doc is one example thought.  We should write another doc to explore this new thought more fully.

(And as exemplified by how I had to link to a PR comment above, rather than the actual line of interest... I'd still really like to merge this "today"(ish) so it's nicely linkable for those future discussions.)

vmx (approved):
Feel free to merge. If there's any better place to post me comment, let me know.

warpfork:
> If there's any better place to post me comment, let me know.

I think it'd be cool to write up another doc and toss it in the same folder this one established -- do some exploring of a couple different variations on where we might stuff those parameters, and what impacts that has, and maybe try to sift a pros/cons list out of the end of it.

(I suspect there's `$N$` different ways of tackling the parameter association problem, for pretty large value of `$N$`, so it will get a much better treatment by getting its own full exploration doc that's got fewer other issues to cover.)

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+}
+
+advanced HAMT {
+	kind map
+	implementation "IPLD/experimental/HAMT/v1"
+}
+
+type Attribs struct {
+	mtime Int # Danger danger will robinson; int sizes will rear their ugly head.
+	posix Int # Can we just... do the standard 0777 mask packing here?  I don't think anything else is *less* work.
```
mib-kd743naq, line 55: A note on this particular point. Thre **is** a way to go with less work, at least on the golang side.
This is how I re-package all of `filemode` into a little endian 32 bit valie on my side for UFS1.5:
```
      # Mode bit-positions are protocol-safe:
      # https://golang.org/pkg/os/#FileMode
      # https://github.com/golang/go/issues/25422#issuecomment-389497051
      ( $_->{mode} & 0777 )
        |
      ( ( S_IFDIR == $_->{ifmt} ) && 1<<31 )
        |
      ( ( S_IFLNK == $_->{ifmt} ) && 1<<27 )
        |
      ( ( $_->{mode} & S_ISUID ) && 1<<23 )
        |
      ( ( $_->{mode} & S_ISGID ) && 1<<22 )
        |
      ( ( $_->{mode} & S_ISVTX ) && 1<<20 )
```

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+	implementation "IPLD/experimental/rabin/v1"
+}
+
+advanced HAMT {
+	kind map
+	implementation "IPLD/experimental/HAMT/v1"
+}
+
+type Attribs struct {
+	mtime Int # Danger danger will robinson; int sizes will rear their ugly head.
```
mib-kd743naq, line 54: Optionality of schema's aside, on the subject of `mtime` these days I would say subseconds are mandatory. This is how I extended the protobuf in UFS1.5:
```
// A generic high precision time specification
message Timespec {
       // Use of int64 is deliberate - negative epoch is super unlikely
       // Using sint64 would introduce zig-zag encoding ( harder to eyeball )
       // The varint representing the time of writing this is 5 bytes long
       // It will remain so until October 26, 3058 ( 34,359,738,367 )
       required int64 UnixEpoch = 1;

       // fixed32 ( always 4 bytes, no varint ), as nanosecs are often > 2^28
       // https://developers.google.com/protocol-buffers/docs/proto#scalar
       // NOTE: on the wire this value will be *little* endian
       optional fixed32 UnixEpochNsec = 2;
}
```

Similar sizing considerations will apply to CBOR. I would strongly recommend reconsidering use of 32bit ints...

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+type Filename string
+
+type Dir struct {
+	members {Filename:DirEnt}<HAMT>
+}
+
+type DirEnt struct {
+	content &AnyFile
+	attribs Attribs
+}
```
mib-kd743naq, line 33: This is the most repugnant comment I am going to leave here, but consider it in terms of pretty significant space-savings. Especially interesting if we agree to **not** use HAMT for a predefined raw-block size, as part of the spec. But even with HAMT the approach can still be applied and drastically reduce directory block sizes ( often by half )

Originally I took a very similar approach to yours - a decorator per link, but later @olcbean pointed out that within a typical distribution tarfile the value of `attribs` will be usually identical for all files ( `root/root/0644/identical-mtime` ) and for all directories ( `root/root/0755/identical-mtime` ).

Thus ( again, in my little fork-world of legacy protobuf ) I went with:

```
message Data {
	enum DataType {
		Raw = 0;
		Directory = 1;
		File = 2;
		Metadata = 3;
		Symlink = 4;
		HAMTShard = 5;
	}

	required DataType Type = 1;
	optional bytes Data = 2;
	optional uint64 filesize = 3;
	repeated uint64 blocksizes = 4;

	optional uint64 hashType = 5;
	optional uint64 fanout = 6;

	// contains interesting parts of the POSIX stat struct
	repeated DecoratorPOSIX DecoratorPOSIX = 7;
}

message DecoratorPOSIX {

        // 0-indexed array indicating which PBLink's does this decorator apply to
        // similar in spirit to blocksizes = 4
        // Mad props to @olcbean for pointing out that within an e.g. tar archive
        // almost all the modes/mtimes will converge to 2 or 3 unique decorators
        repeated uint32 DecoratedPBLinkIndices = 1 [packed=true];

        // Carries precise layout of https://golang.org/pkg/os/#FileMode
        // NOTE: on the wire this value will be *little* endian
        // Mode bit-positions are protocol-safe:
        // https://github.com/golang/go/issues/25422#issuecomment-389497051
        optional fixed32 FileMode = 2;

        // Each timespec weighs at 11 + 2 bytes of wrapping
        optional Timespec MTime = 3;
}


```

rvagg:
I'm not seeing in here how a root block is defined..

```
 type Dir struct {
	members {Filename:DirEnt}<HAMT>
}
```

```
 advanced HAMT {
	kind map
	implementation "IPLD/experimental/HAMT/v1"
}
```

What exactly is `HAMT` in this context? what's the in-built assumption here about what we'll find when we hit `members`? It's going to have the magic properties that a HAMT needs/uses?
If so, how would you integrated this into my schema for #131. Where does `advanced` fit there? I want to be able to _describe_ the underlying properties inside a root block (or maybe it's not a root block, maybe it's an inline-advanced-root) so I can validate that the layout is correct for my HAMT, but I also want to include this `advanced` thing that says how to break out and interpret it through the `kind map` lens.

This also goes back to @vmx' suggestions about default properties. I want a `bitWidth Int` in the root block of my HAMT, maybe we default that to `8`, but where exactly does that go? `type HashMapRoot struct { bitWidth Int (implicit 8) ...}`, then what does `advanced` hold and how does it connect to `HashMapRoot`. Or flip it and go with @vmx's example, `advanced HashMap { parameters HashMapParameters }` `type HashMapParameters struct { bitWidth Int (implicit 8) }`, but I don't want a separate "parameters", I just want to store a simple `"bitWidth"` data model integer in the root block.

Resolve these for me.

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+------
+
+```
+type AnyFile union {
+	| File "f"
+	| Dir "d"
+	| Symlink "s"
+} representation keyed
+
+type Filename string
```
mikeal, line 24: In the latest round of iterations I’m holding all file and directory metadata in a single property called “meta.” This includes the filename, since it’s also just metadata.

As far as we’re concerned, attributes from a given filesystem (uid, guid, permissions) are no different than the filename or any other custom metadata people might decide to add.

The trick will be to write a strict schema for all the metadata attributes we want to have some semantic meaning around and then make sure IPFS:

1. **Only** ads attributes we’ve defined in the schema.
2. Consistently and non-configurably adds them every time.

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+	members {Filename:DirEnt}<HAMT>
+}
+
+type DirEnt struct {
+	content &AnyFile
+	attribs Attribs
+}
+
+type File struct {
+	content Bytes<Rabin>
```
mikeal, line 36: We need a lot more flexibility here. We need to be able to inline raw binary for small files and to support other Bytesish advanced layouts. Basically, this is `Bytes` **OR** “CID that resolves to a Bytesish advanced layout”.  We also need to decide if a `raw` CID would count as “Bytesish advanced layout” or if we have to carve out another clause for it.

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+	gid Int # make another schema which deliberately excludes this.  try to match this one (the one with more info) first.
+}
+
+# You should be able to imagine sending a fancy graphsync request with some structure like:
+#
+#  {"schema": {{schemaCIDofUnixfs}}, "selector": {{selectorSpec}},
+#   "schemaPatch": {"./dir/transparently/through/to/file": {{schemaCIDofRabinLayout}},
+#   "startAt": {{rootBlockCID+path}} }
+#
+# This kind of fancy graphsync request can be used to make a query which does "the right thing" with
```
mikeal, line 69: I started this “File and Dir *as* composites” experiment and it’s quite nice. It literally **forces** us to store the configuration that generated the file which should make it a lot easier to ensure different implementations produce consistent hashes.

* [File](https://github.com/ipld/js-unixfsv2/blob/types/src/file.js)
* [Dir](https://github.com/ipld/js-unixfsv2/blob/types/src/dir.js)

One really cool thing we get here is that the metadata is easily exposed on File objects as `/Filename.txt/guid`. However, I can’t think of the ideal way to:

* Expose the same metadata on directories since the namespace will conflict with files in the directory.
* Expose a property that returns **all** metadata.

Any property you pick will end up being reserved so it can’t be used for arbitrary metadata or as a filename. The one property we *could* use safely since it’s already reserved in filesystems is “.” but then our paths start to differ wildly from actual filesystem paths: `directoryName/./guid`

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+
+Observe the comments on the `Attribs.uid` and `Attribs.gid` fields.
+
+By using several different (but significantly overlapping) schemas to describe variations
+in which attributes we're storing, we can make unixfsv2 into an extensible yet still
+very well defined system.
+
+Concretely: If the `uid` and `gid` elements are found present in the first `Attribs` map
+processed in a tree of data, then we match the first schema (the one with more info),
+and we can now expect *all future* `Attribs` maps to consistently include the same data.
```
mikeal, line 91: I’m not sure I understand what “all future” means in this context. Do you mean all future sub-files and directories in the tree?

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+We might also consider using different characters for this denotation.
+(Angle brackets are reminiscent of generics in Java syntax;
+this may or may not be a helpful reminiscence.)
+
+We might also consider a different approach entirely: ditching the ability to
+declare advanced layouts on inline type terms at all, and instead using
+something more like the `representation` keyword.
+Such an approach would syntactically require a fully named type for any
+structure that wants to use an advanced layout.
+Given that the most common applications of advanced layouts are for map, list,
```
mikeal, line 114: If we have an “in band” mechanism for attaching schemas to composites, do we actually *need* to do all of this definition up front?

For example, if we have a definition of “Bytesish” that is something along the lines of “implements `read()` and `length()`” we would use that in this base schema. The schema for `RabinBytes` could then live in the composite and we don’t need to increase the volume of syntax and definitions here. This is sort of nice as it lets you provide general assurances over a broad graph and then much more granular assertions as you work your way down the graph.

If someone wants to be very explicit and say “this must be RabinBytes” we obviously have the ability to do that, but it would be much nicer to just say “I need a thing that I can treat as binary” and leave the door open to future Bytes implementations. As long as we attach the schema to the composite you’ll still have codegen and stricter validation once you have one of these nodes.

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
```diff
@@ -0,0 +1,296 @@
...
+
+advanced HAMT {
+	kind map
+	implementation "IPLD/experimental/HAMT/v1"
+}
+
+type Attribs struct {
+	mtime Int # Danger danger will robinson; int sizes will rear their ugly head.
+	posix Int # Can we just... do the standard 0777 mask packing here?  I don't think anything else is *less* work.
+	sticky Bool (implicit: false)
```
Artoria2e5, line 56: A very late comment here: on most Unix systems these three boolean bits are part of a 16-bit mode. These systems typically reserve the top 4 bits for seven types of files, the mid 3 bits for the 3 special bits, and the final 9 bits for permissions.

design-history/exploration-reports/2019.06-unixfsv2-spike-01.md
Artoria2e5: Typical Unix systems do with a 16-bit mode value. The top four bits (S_IFMT) are used for 7+ types of files, the middle 3 for the special bits, and the last 9 for the traditional permissions. The only good thing about this stuff (save for the sizes) is that it matches the "100644" notation found in git, but there ain't much else.

Either way, the three bool fields definitely should be stored with the mode.

Artoria2e5:
A nice-to-have would be extended attributes or forks. This could simply be a layer over unixfs v2 by using a separate file with a name that is illegal on Unix, like `"foo.txt\x00user.comment"`. The only issue is to make sure everyone is doing it the same way. (It's also possible to [go pax on this](https://www.freebsd.org/cgi/man.cgi?query=tar&apropos=0&sektion=5), but we are only supposed to use these sparingly.)

PR #133, opened 2019-06-28
Update Selectors doc; new schema proposal.

warpfork:
This is significantly more fully-worked than the previous schema;
as of this commit, however, it also doesn't update all examples.
More work needed; hoisting this as it is, and the example updates
shall come in subsequent commits as the time is available.

### DIFF
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -296,7 +296,7 @@ type Selector union {
 ## ExploreAll is similar to a `*` -- it traverses all elements of an array,
 ## or all entries in a map, and applies a next selector to the reached nodes.
 type ExploreAll struct {
-	next Selector (alias ">")
+	next Selector (rename ">")
 }
 
 ## ExploreFields traverses named fields in a map (or equivalently, struct, if
@@ -307,22 +307,22 @@ type ExploreAll struct {
 ## as a set of three nexted ExploreFields selectors, each specifying one field.
 ## (For this reason, we don't have a special "ExplorePath" feature; use this.)
 type ExploreFields struct {
-	fields {String:Selector} (alias "f>")
+	fields {String:Selector} (rename "f>")
 }
 
 ## ExploreIndex traverses a specific index in a list, and applies a next
 ## selector to the reached node.
 type ExploreIndex struct {
-	index Int (alias "i")
-	next Selector (alias ">")
+	index Int (rename "i")
+	next Selector (rename ">")
 }
 
 ## ExploreIndex traverses a list, and for each element in the range specified,
 ## will apply a next selector to those reached nodes.
 type ExploreRange struct {
-	start Int (alias "^")
-	end Int (alias "$")
-	next Selector (alias ">")
+	start Int (rename "^")
+	end Int (rename "$")
+	next Selector (rename ">")
 }
 
 ## ExploreRecursive traverses some structure using fragments of another
@@ -335,9 +335,9 @@ type ExploreRange struct {
 ## Be careful when using ExploreRecursive with a large maxDepth parameter;
 ## it can easily cause very large traversals.
 type ExploreRecursive struct {
-	sequence Selector (alias ":>")
-	maxDepth Int (alias "d")
-	stopAt Condition (alias "!") # if a node matches, we won't match it nor explore its children.
+	sequence Selector (rename ":>")
+	maxDepth Int (rename "d")
+	stopAt Condition (rename "!") # if a node matches, we won't match it nor explore its children.
 }
 
 ## ExploreRecursiveEdge is a special sentinel value which is used to mark
@@ -364,8 +364,8 @@ type ExploreUnion list [Selector]
 ## whereas a Matcher with a Condition may look deeper to make its decision,
 ## but returns a match for the node it's on rather any of the deeper values.
 type ExploreConditional struct {
-	condition Condition (alias "&")
-	next Selector (alias ">")
+	condition Condition (rename "&")
+	next Selector (rename ">")
 }
 
 ## Matcher marks a node to be included in the "result" set.
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -325,15 +325,31 @@ type ExploreRange struct {
 	next Selector (rename ">")
 }
 
-## ExploreRecursive traverses some structure using fragments of another
-## Selector tree; whenever it reaches an ExploreRecursiveEdge marker in the
-## Selector tree, it will instead produce another new Selector which is a
-## copy of itself, but with a decremented maxDepth parameter.
+## ExploreRecursive traverses some structure recursively.
+## To guide this exploration, it uses a "sequence", which is another Selector
+## tree; some leaf node in this sequence should contain an ExploreRecursiveEdge
+## selector, which denotes the place recursion should occur.
 ##
-## ExploreRecursive is useful when walking large tree structures.
+## In implementation, whenever evaluation reaches an ExploreRecursiveEdge marker
+## in the recursion sequence's Selector tree, the implementation logically
+## produces another new Selector which is a copy of the original
+## ExploreRecursive selector, but with a decremented maxDepth parameter, and
+## continues evaluation thusly.
+##
+## It is not valid for an ExploreRecursive selector's sequence to contain
+## no instances of ExploreRecursiveEdge; it *is* valid for it to contain
+## more than one ExploreRecursiveEdge.
+##
+## ExploreRecursive can contain a nested ExploreRecursive!
+## This is comparable to a nested for-loop.
+## In these cases, any ExploreRecursiveEdge instance always refers to the
+## nearest parent ExploreRecursive (in other words, ExploreRecursiveEdge can
+## be thought of like the 'continue' statement, or end of a for-loop body;
+## it is *not* a 'goto' statement).
 ##
 ## Be careful when using ExploreRecursive with a large maxDepth parameter;
-## it can easily cause very large traversals.
+## it can easily cause very large traversals (especially if used in combination
+## with selectors like ExploreAll inside the sequence).
 type ExploreRecursive struct {
 	sequence Selector (rename ":>")
 	maxDepth Int (rename "d")
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -6,16 +6,11 @@ This document is a designdoc for IPLD Selectors.
 
 ## [Meta: Status of this doc]
 
-- This was written around 2018-10-16 ([video presentation](https://drive.google.com/file/d/1NbbVxZQFKXwW6mdodxgTaftsI8eID-c1/view))
-- It narrows down the decision space enough to make significant progress.
-- good enough for trying out an implementation to learn more and make choices.
-- But it is not complete.
-- some choices that need to be made:
-  - [ ] select general binary and string format structure for selectors. (options given here)
-  - [ ] binary and string formats for each selector. doesn't have to be here.
-  - [ ] whether to dump all selector codes into multicodec table, or one code.
-  - [ ] which S expression selector variant to use (may be out of scope for this doc)
-- more prose here may help implementors.
+- Much of the problem definition was was written around 2018-10-16 ([video presentation](https://drive.google.com/file/d/1NbbVxZQFKXwW6mdodxgTaftsI8eID-c1/view))
+  - It narrows down the decision space enough to make significant progress.
+  - But it is not complete.
+- More concrete proposals for selector implementations (circa 2019-06) can be seen later in the doc.
+  - Particularly, in the [Schema](#schema) section.
 
 ## Motivation - what are Selectors?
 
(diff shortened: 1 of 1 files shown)


selectors/selectors.md
```diff
@@ -115,47 +110,15 @@ These components imply or expand into the following things:
...
-Uses another Selector to select things recursively. It also takes a limit on how often this recursion should happen (the limit can be set to "infinity" to keep traversing until there are no matching paths left.
-
-The recursion may also be stopped by a given CID.
-
-
-### CID Selector
-
-There is no CID selector.  Asking for a specific CID doesn't involve any traversal; it's just asking for a node itself.  No selectors are necessary to ask for data by CID.
-
+See [Schema](#schema) section below.
```
rvagg, line 113: why not just entirely remove the "Selector types" section?

selectors/selectors.md
```diff
@@ -302,67 +265,152 @@ If it's a file in some directory, you can also start at a deeper level:
...
+	| ExploreRecursiveEdge "@" # sentinel value; only valid in some positions.
+} representation keyed
 
-# SelectAll considers all members of the data; think of it as similar to `./*`.
-type SelectAll struct {
-	next Selector (alias ":")
+## ExploreAll is similar to a `*` -- it traverses all elements of an array,
+## or all entries in a map, and applies a next selector to the reached nodes.
+type ExploreAll struct {
+	next Selector (alias ">")
```
rvagg: Off topic, regarding schemas, I think we have to change "alias" to something else. It seems to be a translation, not an alternative:

> alias (n): an assumed or **additional** name that a person (such as a criminal) **sometimes uses**

In this case, is it possible that a block with "next" will resolve here and be what's used if data were serialised according to this schema? Or is it intended that ">" would be the most common (or only) form at the data model layer?

warpfork: Agree.  Someone suggested to me last night that in the words used by Rust's Serde system, this would be called "rename".  That would probably do a better job describing what's going on here than "alias".

---

Separately: I think it might be neat to have a tool that does transforms *on schemas*, which we could use to (for example) drop all the 'rename's.  Dropping all the 'rename's in a schema would produce a topologically/cardinalically equivalent schema -- meaning we could map data described by one schema into the transformed schema guaranteed losslessly -- and then we could do things like easily have compact representations on the wire, using one schema, and flip it into more "debug-friendly" representations at will.

(This is all something we could also build special features for, but I think it might just "fall out naturally" from the rest of the foundational work, which would be neat.)

That's an idea for "later", though.

selectors/selectors.md
hannahhoward: +1

selectors/selectors.md
```diff
@@ -115,47 +110,15 @@ These components imply or expand into the following things:
...
-
-There is no CID selector.  Asking for a specific CID doesn't involve any traversal; it's just asking for a node itself.  No selectors are necessary to ask for data by CID.
-
+See [Schema](#schema) section below.
 
 
 Use cases
 ---------
 
+WARN: not all of these examples have been updated recently.
```
hannahhoward, line 119: Similar to above -- IMHO it's better to have no example at all than one that is out of date. If we need to update and want to build of what was there we can always look at git history.

selectors/selectors.md
```diff
@@ -302,67 +265,152 @@ If it's a file in some directory, you can also start at a deeper level:
...
+##   - in ExploreRecursive, to halt exploration.
+##   - in ExploreConditional,
+##
+##
+## TODO -- Condition is very skeletal and incomplete.
+## The place where Condition appears in other structs is correct;
+## the rest of the details inside it are not final nor even completely drafted.
+type Condition union {
+	# We can come back to this and expand it later...
+	# TODO: figure out how to make this recurse correctly, so I can say "hasField{hasField{or{hasValue{1}, hasValue{2}}}}".
```
hannahhoward, line 414: It seems like we're heading towards matching / explore being equal in capability, and I wonder if that's neccesary. Maybe there is only matching at the end -- i.e. exploreFields{exploreFields{or{hasValue{1}, hasValue{2}} -- because that's the equivalent yes? -- I would argue against having matching be able to recurse

hannahhoward:
@warpfork @mikeal unless anyone objects can I suggest we merge this and mark selectors exploratory? It's easier to work off a not-PR :)

selectors/selectors.md
```diff
@@ -302,67 +265,152 @@ If it's a file in some directory, you can also start at a deeper level:
...
-type SelectFields map {String:Selector}
+## ExploreFields traverses named fields in a map (or equivalently, struct, if
+## traversing on typed/schema nodes) and applies a next selector to the
+## reached nodes.
+##
+## Note that a concept of "ExplorePath" (e.g. "foo/bar/baz") can be represented
+## as a set of three nexted ExploreFields selectors, each specifying one field.
+## (For this reason, we don't have a special "ExplorePath" feature; use this.)
+type ExploreFields struct {
+	fields {String:Selector} (alias "f>")
```
hannahhoward: Question: The current go-ipld-prime code treats the boolean true as valid in the place of Selector here. Does that mean:
a) ExploreFields should be something like:
```
type ExploreFieldsValue union { boolean | Selector }
type ExploreFields struct {
   fields {String:ExploreFieldsValue} (alias "f>")
}
```
or
b) is the value "true" a valid abbrievation for a selector that is a position only matcher with no label?
(I kind of like that)

warpfork: That code is old and we should replace it with something matching this schema.

(I kinda liked the "true" as an abbreviation too, tbh, but it made a bunch of other things harder and weirder.  (I have an open TODO to make schemas have a "unit" kind (!) plus representation options so it can declare things such as `type Dummy unit representation true`.  And then make kinded unions know how to support that.  Yeeeee...ahh.))

selectors/selectors.md
warpfork: Yeah, it's possible that it would turn out wiser to limit this feature.

The one major case I can think of that's super super common and definitely needs to be supported is: given some nodes in an array, each is a map, and I want to decide whether or not to explore or match that node based on a property in it.

I don't have any example use cases of deeper recursion.

Another possibly sensible limit might be to say "this can recurse, but not across another link load boundary"?

Looking at what `jq` does might also be interesting.  That system is subject to different constraints than we are (namely, link loading doesn't exist for them, so of course it can't be a performance cost issue), but if there *are* examples we can find of `jq` usage in the wild that do any deeper conditions, that might be interesting to consider.

selectors/selectors.md
```diff
@@ -302,67 +265,152 @@ If it's a file in some directory, you can also start at a deeper level:
...
+	# TODO: figure out how to make this recurse correctly, so I can say "hasField{hasField{or{hasValue{1}, hasValue{2}}}}".
+	| Condition_HasField "hasField"
+	| Condition_HasValue "=" # will need to contain a kinded union, lol.  these conditions are gonna get deep.)
+	| Condition_HasKind "%" # will ideally want to refer to the DataModel ReprKind enum...!  will we replicate that here?  don't want to block on cross-schema references, but it's interesting that we've finally found a good example wanting it.
+	| Condition_IsLink "/" # will need this so we can use it in recursions to say "stop at CID QmFoo".
+	| Condition_GreaterThan "greaterThan"
+	| Condition_LessThan "lessThan"
+	| Condition_And "and"
+	| Condition_Or "or"
+	# REVIEW: since we introduced "and" and "or" here, we're getting into dangertown again.  we'll need a "max conditionals limit" (a la 'gas' of some kind) near here.
```
hannahhoward, line 423: Condition_Not? Just saying you need And/Not/Or to derive all boolean logic.

selectors/selectors.md
hannahhoward: is it safe to say a `glob-star` selector i.e. ** (for now, 'just gimme the whole dag') could be constructed as:
is there a simpler version of this?

```
exploreUnion:
   list:
     matcher:
     exploreRecursive:
       maxDepth: ...
       stopAt:
          conditionAnd:
              conditionNot
                 conditionHasKind:
                   list
              conditionNot
                 conditionHasKind:
                   map
       sequence:
          exploreAll:
            next:
               exploreRecursiveEdge:
```

hannahhoward: also do we have a canonical notation for selectors? Json? I realize now they're actually written in abbreviated keyed union form.

selectors/selectors.md
```diff
@@ -302,67 +265,152 @@ If it's a file in some directory, you can also start at a deeper level:
...
+## ExploreIndex traverses a list, and for each element in the range specified,
+## will apply a next selector to those reached nodes.
+type ExploreRange struct {
 	start Int (alias "^")
 	end Int (alias "$")
-	next Selector (alias ":")
+	next Selector (alias ">")
+}
+
+## ExploreRecursive traverses some structure using fragments of another
```
hannahhoward: is it considered value to have an ExploreRecursive inside another ExploreRecursive, and if so, is it reasonable to assume any ExploreRecursiveEdges reference their immediate parent?

selectors/selectors.md
warpfork: The idea of "exploreRecursive" is it'll actually do most of that faff for you.  The following should be a way to say "match everything":

```yaml
exploreRecursive:
  maxDepth: 999
  sequence:
    exploreUnion:
      - exploreAll:
        next:
          exploreRecursiveEdge: {}
      - matcher: {}
```

One could also skip the union and the `Matcher` and still have a selector that explores the whole graph, but doesn't actively 'match' any nodes of it... which is perhaps a subtle distinction, but think of it like a regexp that doesn't have any parens in it.

(Internally in a library implementation: there's almost certainly going to be code which uses unions to carry around the partially evaluated exploreRecursive, if I guess rightly.  But I dunno, maybe there will be other clearer ways to implement that.  And in any event, that's an internal detail.)

...ah.  Also `stopAt` should be optional.  That's a bug, I think.

warpfork: re: notation:

I was writing them as JSON for a while as a path of least resistance, but I've started to notice the amount of closing curly-braces I accumulate gets pretty wicked, to the point that I started writing it in a very lispesque way.  JSON certainly works; I'd also consider [Indentation-sensitive S-exprs](https://srfi.schemers.org/srfi-49/srfi-49.html)!

*cough* or YAML seems particularly handy and fine.  Even if we don't have an official IPLD codec for yaml, it's pretty trivial to mentally flip it into JSON and go from there, and it saves a lot of `}`'s.  (Though one moderate downer with yaml is that a terminal `Matcher` with no value for both of the optional fields serializes as `matcher: {}`, which is perhaps a bit odd on the eyes.  Eh.  C'est la vie.)

Honestly, I don't care.  As long as it's obvious enough how it maps onto the IPLD Data Model (or JSON, which is then obvious enough to carry the rest of the way), it's fine by me.

We've also been sketching a lot of stuff which refers to the full names of fields (and full names of types, when it comes to unions) rather than honoring the short alias/renames in this schema.  I think that's fine for documentation and example purposes (and making schema tooling explicitly support things like "actually, ignore the renames" is something I'd like to do someday, incidentally)... but we should make a footnote stating that we're doing that, since it's otherwise potentially a bit confusing.

selectors/selectors.md
warpfork: Yes and yes, I think.

I had some earlier draft where nested recursion was forbidden, but then a use case was proposed which actually needed it, so I dropped that text :(  So, the first yes.

I couldn't think of any cases where it's sane for ExploreRecursiveEdge to refer to anything other than their immediate parent, though.  We're not trying to build gotos here; that'd be... bad.  So, the second yes!

warpfork: One other moderately terrifying thing I think should probably be supported here: there can be two (or more) different ExploreRecursiveEdge both referring back to the same ExploreRecursive.  Roughly analogous to the use of a 'continue' statement in a for-loop.

selectors/selectors.md
warpfork: Probably :)  I kinda just spitballed this section, in order to say "something roughly like this belongs here".  It almost certainly needs another round of draft -- would definitely give the `Condition` and its members another hard design look first before trying to implement these.

(Also figure they're probably roughly one of the last things to do in the overall order of implementing the selector system.)

hannahhoward: ok. Imma hope i get go aways with filecoin before I need conditionals.

selectors/selectors.md
warpfork: Ok.  I'm going to get increasingly aggressive about the diffs in the next couple commits and/or next PR.

I've been trying to be pretty conservative about editing this file, but you're right, that's silly -- and per https://github.com/ipld/specs/pull/141#issuecomment-510777921, istm a fair amount of the older sections of this file belong lovingly cordoned off to some nicely dated, referenceable, but no-longer-here design history notes.

selectors/selectors.md
warpfork: Pushed some additional docs here to clarify and expand upon this. :+1:

selectors/selectors.md
warpfork: Marking this thread as resolved because it's a known todo and it's staying that way until future iterations -- not gonna handle in this PR.

warpfork:
Ok, I'm now working on an updated (and let's be honest, at this point, it's more "new" than "updated") set of examples -- but that, as well as [factoring out any design history](https://github.com/ipld/specs/pull/141#issuecomment-510777921) from what's modern and succinctly relevant, are both going to be large diffs and I'd like to do them in new PRs.

So... mergin'!

PR #134, opened 2019-07-01
Schema update: use implicits; also, more docs.

warpfork:
Schema update: use implicits; also, more docs.

- the "default" feature of representations has been renamed:
it is now known as "implicit".  We believe this better clarifies
what the feature does: the old name suggests that it might create
a surjective function; it is in fact a bijective function.
(Thanks to @merovius for the first suggestion of this new name!)

- several representation blocks removed: we've introduced the
ability to annotate fields in struct types with per-field
representation details inline, which shortens some things
significantly.  (The IPLD-native representation still keeps
representations clearly separated even at the cost of redundant keys.)

- new documentation file describing the above design.

### DIFF
--- a/schema-layer/schemas/schema-dsl.md
+++ b/schema-layer/schemas/schema-dsl.md
@@ -1,0 +1,40 @@
+Schema DSL
+==========
+
+IPLD Schemas can be represented in a compact, human-friendly
+[DSL](https://en.wikipedia.org/wiki/Domain_specific_language).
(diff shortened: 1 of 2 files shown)


warpfork:
(I think this one is basically uncontroversial, but I'm going through the motions of a PR anyway so the diff is highlighted and linkable via the web.)

whyrusleeping (approved):

PR #135, opened 2019-07-01
Schema update: no more anon params to representation strategy.

warpfork:
Previous syntax which allowed anonymous parameters to representation strategies dropped.  (This had appeared in the schema-schema in the uses of "inline" unions.)  Those parameters are now always named, and thus require the use of a full block to complete the representation description.  This is much more future-proof, and gives more indications about what is going on, as well as more words we can ascribe documentation to.

This change, like most of the others I'm making today, is prompted by multiple accumulated incidents of user feedback indicating that the earlier syntax was non-obvious and forced them to ask questions.

### DIFF
--- a/schema-layer/schemas/schema-schema.ipldsch
+++ b/schema-layer/schemas/schema-schema.ipldsch
@@ -87,7 +87,9 @@ type Type union {
 	| TypeUnion "union"
 	| TypeStruct "struct"
 	| TypeEnum "enum"
-} representation inline "kind"
+} representation inline {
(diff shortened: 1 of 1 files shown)


warpfork:
(I think this one is basically uncontroversial, but I'm going through the motions of a PR anyway so the diff is highlighted and linkable via the web.)

whyrusleeping (approved):

rvagg:
+1 but the increase in explicitness conflicts with the increase in implicitness in #136, more explicit is usually better

PR #136, opened 2019-07-01
Schema update: remove obvious/redundant kind keywords in type declarations.

warpfork:
Elide 'map' and 'list' keywords in top-level type declarations.

In all other positions (e.g. inline type declarations for struct
fields, nested maps, and nested lists), there's no kind keyword used:
the '{' and '[' characters in the typeterm are already enough to be
explicit about what's going on.  Historically, we've also added
a kind keyword when making top-level type declarations, even though
it's technically redundant.

I'm a bit tepid about this change.  However, many people have
commented that the 'map' and 'list' keywords in this position seem odd
and redundant.  I'm outvoted.

The original idea was that the form is always roughly
"`type $typeName $kind [$block...]`" -- and the "block" would be
present only in types that need more info (like structs).
But it's not much of a pivot to just switch the model to
"`type $typeName $typeDefn`" -- where there's not necessarily an
appearance of a kind keyword, and instead when we see words like
"struct" appear here, it's because it's part of the "typeDefn".

So.  Here's that take.

### DIFF
--- a/schema-layer/schemas/schema-schema.ipldsch
+++ b/schema-layer/schemas/schema-schema.ipldsch
@@ -29,7 +29,7 @@ type TypeName string
 ## }
 ## ```
 ##
-type SchemaMap map {TypeName:Type}
+type SchemaMap {TypeName:Type}
 
 ## Schema is a single-member union, which can be used in serialization
 ## to make a form of "nominative type declaration".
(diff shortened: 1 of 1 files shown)


dignifiedquire (approved):
thank you

whyrusleeping (approved):

rvagg:
this makes parsing much less straightforward

I liked the previous form, were the extra characters really so difficult to type?

warpfork:
The main issue was the consistency with how in structs and other places where inline/nested defns are allowed...

```
type Foo struct {
    fieldname {String:String} # no map keyword here
    nested {String:[nullable String]} # definitely no list keyword here
}
```

This seems to have confused a lot of people: if there's no keyword in situations like that, why is there a keyword in the type name assignment clause?

Having a kind keyword always required in the type naming clause made more sense to me, too, when thinking about it as a grammar.  But I lost count of how many people at the IPFScamp week either had questions triggered by this, or else directly identified it as confusing.  Of those who had looked at the selector schema, the schema schema, or any other document where a named map shows up, I think nearly 100% had questioned this.  I regret not keeping a proper tally in a notebook, but in short, I feel very out-voted.

PR #138, opened 2019-07-03
Add Vector spec

rvagg:
I've not yet included algorithm details for operations I don't yet support in <https://github.com/rvagg/iavector>, but I don't want those to be a blocker for making progress on getting this merged, they could come later. The current operations are the most useful for putting this thing to work at the moment.

### DIFF
--- a/schema-layer/data-structures/vector.md
+++ b/schema-layer/data-structures/vector.md
@@ -1,0 +1,362 @@
+# Specification: Vector
+
+**Status: Prescriptive - Draft**
+
+* [Introduction](#Introduction)
+* [Useful references](#Useful-references)
+* [Summary](#Summary)
+  * [Example](#Example)
+  * [Algorithmic properties](#Algorithmic-properties)
+* [Structure](#Structure)
+  * [Parameters](#Parameters)
+  * [Node properties](#Node-properties)
+  * [Schema](#Schema)
+* [Algorithm in detail](#Algorithm-in-detail)
+  * [`Get(index)`](#Getindex)
+  * [`PushTail(value)`](#PushTailvalue)
+  * [`Size()`](#Size)
+  * [`CreateFrom(array, width)`](#CreateFromarray-width)
+  * [`Set(index, value)`](#Setindex-value)
+  * [`PopTail()`](#PopTail)
+  * [Full copy mutations: `Delete(index)`, `PushHead(value)`, `PopHead()`, `Slice(start, end)`](#Full-copy-mutations-Deleteindex-PushHeadvalue-PopHead-Slicestart-end)
+* [Implementation defaults](#Implementation-defaults)
+* [Performance profile](#Performance-profile)
+* [Possible future improvements and areas for research](#Possible-future-improvements-and-areas-for-research)
+  * [Compression for sparse arrays](#Compression-for-sparse-arrays)
+  * [Flexible head for efficient operations a the head and for sub-trees](#Flexible-head-for-efficient-operations-a-the-head-and-for-sub-trees)
+
+## Introduction
+
+The IPLD Vector distributes ordered-by-construction values across a tree structure with a pre-definable branching factor (`width`) that is consistent across all nodes in the tree. The tree defines node "height" rather than "depth", with the values stored in leaf nodes with a `height` of `0`. All these leaf nodes are organized into a collection of `height` `1` nodes which contain links to the leaf nodes rather than values. We continue to increment `height` until we have a single node as the root of the tree.
+
+An IPLD Vector containing fewer values than the `width` of the Vector can be represented by a single root node, with a `height` of `0` and a `data` array containing only the values. Once an IPLD Vector expands beyond `width`, we add additional sibling `height` `0` nodes and reference them in a parent `height` `1` node. Once the `height` `1` node expands beyond `width` child nodes, we perform the same operation by adding a new `height` `2` node to organize `height` `1` nodes.
+
+## Useful references
+
+* [Understanding Clojure's Persistent Vectors, pt. 1](https://hypirion.com/musings/understanding-persistent-vector-pt-1) and [pt. 2](https://hypirion.com/musings/understanding-persistent-vector-pt-2) are excellent introductions to some of the concepts covered here, including graphics that help describe the tree structures.
+* [Fast And Space Efficient Trie Searches](https://infoscience.epfl.ch/record/64394/files/triesearches.pdf) by Phil Bagwell, 2000, and [Ideal Hash Trees](http://lampwww.epfl.ch/papers/idealhashtrees.pdf) by Phil Bagwell, 2001, introduce the AMT and HAMT concepts, which has some similarity to the data structure described here.
+* Daniel Spiewak's Strange Loop '01 presentation titled [Extreme Cleverness: Functional Data Structures in Scala](https://www.infoq.com/presentations/Functional-Data-Structures-in-Scala/) covers the concept of Bitmapped Vector Tries, roughly equivilant to Clojure's Persistent Vector which is roughly the same as the data structure outlined here.
+* [sharray](https://github.com/whyrusleeping/sharray) a Go implementation of this data structure directly implemented against CBOR with Create and Iteration functionality.
+* [iavector](https://github.com/rvagg/iavector) the JavaScript reference implementation of this data structure.
+
+## Summary
+
+IPLD Vector's algorithm is roughly based on the data structure used by [PersistentVector in Clojure](https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/PersistentVector.java) and referred to as a Bitmapped Vector Trie in [Scala's Vector](https://github.com/scala/scala/blob/v2.13.0/src/library/scala/collection/immutable/Vector.scala). It has roots in the concept of Array Mapped Tries (AMT), as outlined in Phil Bagwell's [papers](https://infoscience.epfl.ch/record/64394/files/triesearches.pdf) on the [subject](http://lampwww.epfl.ch/papers/idealhashtrees.pdf). In these data structures, the indexing at each level of the trie comprises portions of the requested index. By taking advantage of efficient bitwise operations, we can slice an index into segments which point us through each level as we descend to the final value. The concept is roughly similar to slicing a hash as outlined in the [IPLD HashMap](hashmap.md) specification, except that we are slicing an index.
+
+One major difference with these data structures come from IPLD's minimal capacity to make use of the efficiencies afforded by bitwise operations. Without requiring bitwise operations, we don't have a strong need to align to byte or word boundaries and can use non-bitwise operations to perform our indexing function. Hence, the `width` of the IPLD Vector is variable (not a power of `2` as for the width of nodes in the [HashMap](hashmap.md)), from a lower bound of `2`, for very tall trees that yield very small blocks, up to very large values that yield shallow trees but very large blocks. We leave the option of storing leaf values as CIDs or inline data up to the user, thereby affording the possibility of tuning `width` to the desired block size with a traversal cost trade-off.
+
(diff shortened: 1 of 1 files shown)


rvagg:
One thing I wanted to draw attention to is the section at the bottom _"Flexible head for efficient operations at the head and for sub-trees"_.
I haven't fully implemented that idea yet, made a start but it got complicated really quickly. But if we successfully managed to spec that out then we'd have a pretty performant data structure that can be mutated and sliced in all sorts of ways without the need to do bulk copies (deletes in the middle still require bulk copies).
Losing canonical form guarantees is a bit of a cost but that could be mitigated by maintaining strict rules where these things are used _if_ there is a need for such guarantees. Or we could have a mode that turns off head-modification operations (and opts for bulk copying).
The tricky bit is that it needs time to investigate and I don't think I can justify that right now as I have some other things on my plate.

schema-layer/data-structures/vector.md
```diff
@@ -0,0 +1,361 @@
...
+2:                                       [A  B]
+                 ┌────────────────────────┘  │
+1:            [a b c]                    [d  e]
+         ┌─────┘ │ └─────┐        ┌───────┘  │
+0:    [1 2 3] [4 5 6] [7 8 9] [10 11 12] [13 14 15]
+
+      Size: 15
+      Max size before root overflow: 27
+      Head chain: A → a → 1
+      Tail chain: B → d → 15(full)
```
vmx: That should be: B → **e** → 15(full)

schema-layer/data-structures/vector.md
```diff
@@ -0,0 +1,361 @@
...
+      Tail chain: ii → D → j → 30(full)
+```
+
+### Algorithmic properties
+
+We can derive the **maximum number of elements containable under a node and its children** with a `height` of `X` with `width`<sup>`X + 1`</sup>. At a `height` of `0`, we can only fit `width` elements into a single node. Add a `height` `1` parent and we can fit in `width`<sup>`2`</sup> elements across all child `height` `0` nodes underneath it. This calculation is useful for at least `Size` and append (which we will call `PushTail`) operations.
+
+**Overflows** during `PushTail` operations do not always cascade to upper levels of the tree. Cascades only occur where an upper level is maximally full. So a cascade causing the addition of a new root level at `height + 1` only happens when all levels are full.
+
+**All nodes are valid IPLD Vectors**, and as such may be extracted to represent slices of the overall parent data structure. This becomes less useful as we increase `width` as slices must align with node boundaries, particularly the head of the slice (as the tail may be adjusted with minimal mutations).
```
vmx: Would it make sense to say "particularly the head **node** of the slice". Coming from Erlang "head" means to me only the very first element and not the full node. You also previously in this doc say "head node". We could also add a definition of "head" and say that "head" is the "head node" and if you mean the first element of the head node, it's the "head element" or so.

vmx: One more note: it took me a second thought to understand why an increase in `width` makes things less useful. If I understand correctly the problem is if you take a slice that is within a node (and not at the boundary of one) at position `n`, you have two partitions. One with `n` elements and another with `x = width - n` elements. In order to have a proper vector again, you would need to shuffle `x` elements into the head node from the next node (and so on).

schema-layer/data-structures/vector.md
```diff
@@ -0,0 +1,361 @@
...
+  | List list
+} representation kinded
+```
+
+**Constraints:**
+
+* Non-leaf (`height` greater than `0`) nodes only contain `Link`s to other `Vector` nodes in their `data` array.
+* `width` must be consistent across all nodes in a Vector.
+* `height` must be at least `0`.
+* `data` can contain between `0` and `width` elements.
```
vmx: Is there a case where there's a node without any data? Wouldn't you remove that node completely? If not, then `data` can contain between **`1`** and `width` elements.

schema-layer/data-structures/vector.md
```diff
@@ -0,0 +1,361 @@
...
+4. If `dataIndex` is greater than or equal to the length of this node's `data` array, return undefined or an OOB indication.
+5. If `height` is `0`, return the element at the `dataIndex` position of this node's `data` array.
+6. If `height` is greater than `0`, retrieve the CID of the next node from the `dataIndex` position of this node's `data` array.
+7. Recurse or iterate from Step 1 with the next node with `index` set to `index % width`<sup>`height`</sup> to identify its place in the sub-tree.
+
+### `PushTail(value)`
+
+`PushTail` appends `value` to the tail of the data structure. This increases the Vector's `Size` by `1` and may cause zero or more overflows throughout the tree but will result in a single mutation at each `height` of the Vector, including the root and possibly the creation of a new root node at `height + 1` if an overflow cascades to the root.
+
+This algorithm can be implemented as a recursive or iterative process. The additional state needing to be transferred when mutating back up the tree from the mutated root may make recursive implementations awkward.
```
vmx: My gut feeling is that a recursive implementation is even simpler than a iterative one. You don't need to keep additional state, you already have the state you need at each recursion step anyway.

schema-layer/data-structures/vector.md
```diff
@@ -0,0 +1,361 @@
...
+   1. Pop the tail node of the tail chain as the "current node"
+      1. If the `mutated` state variable is `true` we only need to replace the link to a child node.
+         1. Create a copy of the current node, replacing the last element of the `data` array of the copy with `value`. This is guaranteed to be at `height` of at least `1` since `mutated` begins as `false`, so we are only replacing links to mutated nodes.
+         2. Repeat from step 3 with `value` set to the CID of the mutated node.
+      2. If the length of the `data` array at the current node is less than `width`, no overflow is necessary.
+         1. Create a copy of the current node, appending `value` to the end of the `data` array of the copy.
+         2. Set `mutated` to `true`
+         3. Repeat from step 3 with `value` set to the CID of the mutated node.
+      3. If the length of the `data` array at the current node is `width`, an overflow is necessary.
+         1. Create a new node with the same `height` as the current node and a `data` array that only contains the single `value`.
```
vmx, line 273: I think you need a "2. Set `mutated` to `false`." here as else you would lose the data of the overflowing node.

schema-layer/data-structures/vector.md
```diff
@@ -0,0 +1,361 @@
...
+      2. Return the CID of the new node, which becomes the new root node.
+   2. If `mutated` is `true`
+      1. Return `value`, which should be the CID of the last mutated node, which becomes the new root node.
+
+### `Size()`
+
+The `Size` algorithm uses a subtractive process, first calculating the maximum potential size and then iterating through the tail chain to subtract known empty portions given varying `data` array sizes at each `height`. If `data` arrays are maximally full at each `height` then the original maximum potential size calculation is used as the correct value.
+
+1. Set `size` to be the maximum potential size of the data structure with `width`<sup>`height + 1`</sup>.
+2. Calculate the number of empty potential value slots at this height with `width`<sup>`height * (width - dataLength)`</sup>, where `dataLength` is the number of elements in the current node's `data` array and `height` is the current node's `height`. Subtract this value from `size` and set the result as the new `size`.
```
vmx: Could it be that the formula should be:

`width`<sup>`height`</sup>` * (width - dataLength)`

schema-layer/data-structures/vector.md
```diff
@@ -0,0 +1,361 @@
...
+   1. Locate the next tail node identified by the CID at the last position of the current node's `data` array.
+   2. Repeat from step 2 with the next tail node as the "current node".
+
+### `CreateFrom(array, width)`
+
+1. Set `height` to `0`, indicating that each node created at this `height` will have leaf values in it rather than links to child nodes.
+2. Calculate `nodeCount` for this `height` with `ceiling(arrayLength / width)`, where `arrayLength` is the number of elements in `array`.
+3. Create a temporary `nextArray` array which starts empty but will fill up to `nodeCount`.
+4. Iterate with an incrementing integer `i` from `0` up to, but not including `nodeCount`.
+   1. Take a slice of `array` beginning at element `i x width` and ending at the element before `(i + 1) x width` or `arrayLength`, whichever is least.
```
vmx, line 301: In other parts you've used `*` as symbol for multiplications.

schema-layer/data-structures/vector.md
```diff
@@ -0,0 +1,361 @@
...
+
+`Slice` operations taking an arbitrary subset of the tree's values would only need to mutate nodes in the head chain and tail chain, leaving internal nodes intact.
+
+Two primary trade-offs are an increse in algorithmic complexity and some loss of efficiency as the head chain needs to be traversed to understand both the `Size` and the zero-reference point for `Get` indexing. The efficiency losses could be offset by storing additional metadata in each node regarding its own head chain or its head chain size (the total number of values stored below it on nodes unaligned to `width`). By pre-loading this information we could assume both the head-leaning size for `Size` operations and the head offset for `Get` indexing.
```
vmx: Typo: increse -> increase

vmx (commented):
Wow, great write-up!

schema-layer/data-structures/vector.md
rvagg: OK, so I've tried to address this as:

> This becomes less useful as we increase `width`, as the start index of a slice must align with a node boundary such that the new head node is a copy of an existing node. Otherwise all elements need to be shuffled toward the head, therefore mutating the entire data structure. Lack of alignment a the tail of a slice only requires mutations on the tail chain.

schema-layer/data-structures/vector.md
rvagg: wow, good catch, nice to see someone thinking through the algorithm
confirmed @ https://github.com/rvagg/iavector/blob/01d58752a1f0fc29555680db4d19c917730340e3/iavector.js#L131

schema-layer/data-structures/vector.md
rvagg: great catch again! confirmed @ https://github.com/rvagg/iavector/blob/01d58752a1f0fc29555680db4d19c917730340e3/iavector.js#L651

corrected as:
> `width`<sup>`height`</sup>` x (width - dataLength)`</sup>

schema-layer/data-structures/vector.md
rvagg: I meant to use `x`, I corrected the previous use of `*`

schema-layer/data-structures/vector.md
rvagg: you might be correct, I've soften this language to just:

> A recursive process will need to propagate additional state when mutating back up the tree from the mutated root.

I'll say more about why I'm leaning toward iterative processes below because I think you'll find it relevant.

rvagg:
On the topic of iterative vs recursive processes: I've started off implementing operations both here and in HashMap as recursive because it _feels_ more appropriate. However I'm migrating toward iterative approaches in order to make these algorithms more atomic, where each step is a discrete thing that operates on a block that has only just the right amount of state.

In js-composites it's only possible to do it this way, the logic gets invoked per block, so https://github.com/ipld/js-composites/pull/5/files#diff-1a52ff098f2bcf415524a4220e8d61feR20 is a single block/node traversal of a Vector. That one's nice because we can traverse a single block as if it's the only traversal, just adjusting the `index` for each step. For the HashMap we need more state so we have to use a continuation: https://github.com/ipld/js-composites/pull/1/files#diff-7a3b725525de535be6fe08715127453eR31-R46

Forcing this design style has a benefit of getting us theoretically closer to WASM implementations of these algorithms. Execute individual steps, maintain a certain amount of state. In iavector, I've tried to do most of it this way, even calculating `size()` is a step-by-step traversal https://github.com/rvagg/iavector/blob/01d58752a1f0fc29555680db4d19c917730340e3/iavector.js#L635

Mutations are more awkward and I've started down the road of doing them this way but the state management quickly gets out of hand and I haven't come up with a nice API for it yet. I'm sure there is one but I haven't had time to _iterate_ on the design enough.

schema-layer/data-structures/vector.md
rvagg: > * `data` can contain between `1` and `width` elements. For the special case of the empty Vector, a single root node may have `0` elements in `data`.

rvagg:
This is now merged, but it's: **Status: Prescriptive - Draft**
So amendments are anticipated. PRs invited if you can see ways to improve it.

PR #139, opened 2019-07-08
Exploration notes on int ranges

warpfork:
Numeric precision is a tricky issue.

As with other files added to this directory, this is an "exploration report"
and it contains a series of "in the opinion of this author" sorts of comments.

I'm not sure this one came to a great conclusion, but it contains at least
one interesting probably-good idea, and a variety of useful references for
future iterations.

Since I can't easily proceed much further with this document without mentally
"crumpling it up" and starting a fresh one, it seems like at this point I
ought to checkpoint and share.

### DIFF
--- a/design-history/exploration-reports/2019.07-int-ranges-as-adjuncts.md
+++ b/design-history/exploration-reports/2019.07-int-ranges-as-adjuncts.md
@@ -1,0 +1,367 @@
+What to do about integer range?
+===============================
+
+Concepts
+--------
+
+### cardinality
+
+See [concepts/cardinality](../../concepts/cardinality.md).
+
+### totality
+
+A "total function" is one that has defined outputs for all inputs.
+
+A "partial function" is one that has undefined outputs for some inputs.
+
+### bijection
+
+[Wikipedia has a particularly nice diagram](https://en.wikipedia.org/wiki/Injective_function).
+
+### overflow
+
+Integer overflow can be defined as when one fails to have a bijection.
+
+Overflow without explicit error is a surjection (it silently returns another number).
+
+Overflow with explicit error is a partial function.
+
+(Overflow with explicit error can also be visualized as having a codomain
+which is tuples of numbers and errors, but it's unclear if this produces
+any more useful intuitions than simply considering it a partial function.)
+
+
+Concerns
+--------
+
+### overflow is annoying
+
+This is basic, but let's say it.
+
+Overflow is a source of bugs (especially when it occurs at runtime and
+without explicit error).
+
+Overflow is also a potential source of *design* bugs: if two communicating
+systems have different numeric support ranges, that almost certainly indicates
+a potential compatibility issue.
+(Whether or not such a compatibility issue is reachable in practical usage may
+be another matter, but the possibility itself is identifiable.)
+
+### we need to support "large enough" numbers
+
+We might not be able to remove the concept of overflow from the world.
+
+Knowing that overflow is a source of problems, however, we should at least
+be able to specify when and where it's going to affect us.
+
+We should be able to be clear about any limits to the precision of numbers
+we can handle.  If an application needs to serialize and deserialize numbers
+between 1 and 10 billion, it might be important to declare this: such an
+application will not work with libraries limited to 32 bit integer precision!
+
+### let's not have dependent types
+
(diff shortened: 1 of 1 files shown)


design-history/exploration-reports/2019.07-int-ranges-as-adjuncts.md
```diff
@@ -0,0 +1,367 @@
...
+
+Knowing that overflow is a source of problems, however, we should at least
+be able to specify when and where it's going to affect us.
+
+We should be able to be clear about any limits to the precision of numbers
+we can handle.  If an application needs to serialize and deserialize numbers
+between 1 and 10 billion, it might be important to declare this: such an
+application will not work with libraries limited to 32 bit integer precision!
+
+### let's not have dependent types
```
rvagg, line 62: I foresee this coming up again and again so I think you should consider writing a concepts doc to address this. You're drawing a line in the sand that people are going to consistently bump up against when they want more power out of schemas and we need to be able to point to something to say exactly why that line is there and why making schemas more "powerful" or "expressive" doesn't jive with what they are trying to do.
Just saying "but cardinality" isn't good enough either btw. Why does that matter? Why should a user of IPLD Schemas care? Why is it such an immovable object that is preventing them from getting _moar power_ from schemas?

rvagg:
<details>
<summary>Re codegen and its relationship to keeping IPLD Schemas simple. (Collapsible block because it's a bit off-topic.)</summary>

This integer problem only lightly touches on the kinds of problems you're going to have trying to do codegen from schemas. But if that's one of the primary use-cases of schemas while we're also trying to (a) keep them strictly to the data model and (b) avoid dependent types or anything that smells dependent-ish, then we're in for a world of pain. They're way too loosely defined to give you the kind of solid output you'll want. Int representations are just one of the problems but we'll quickly get to things like:

* My Int should be at least `2`
* The postcode field should be either 4 or 6 characters
* If field X exists then field Y should be constrained in some way
* My `Bytes` field is a bitfield ordered in some specific way
* The float/int `.0` kinded-inspection problem we've discussed before

Further to things like:

* Use a custom unmarshaller to turn type Foo into an object because the built-in doesn't do precisely what I want
* Use a custom marshaller for type Bar because it doesn't exist in memory but can be inferred from some other data structure.

We went through all of this with [XMLSchemas](https://www.w3.org/XML/Schema), which are extremely expressive due to the desire to do codegen, you can express all sorts of bounds for types and values, and dependents abound. But even that isn't quite enough, so JAXB has a bunch of additional extensions for Java codegen (even then it's never quite enough and you end up in the ORM problem of having no elbow room to move and the entire edifice breaking down, but that's another discussion for the future of this stuff). Have a browse through https://github.com/eclipse-ee4j/jaxb-ri/tree/master/jaxb-ri/samples/src/main/samples for some fun, and remember that these are only _simple_ samples, not realistic ones.
</details>

design-history/exploration-reports/2019.07-int-ranges-as-adjuncts.md
warpfork: Aye.  I've been deferring this because it's a thorny topic, but you're right, it's certainly an important one, and it does keep coming up.

It's also a topic that I actually hesitate to write much about because users shouldn't *need* to hear or read anything about dependent types in order to use or understand the system, especially if the prospective user didn't already know that term and concept.  (Did protobufs need a declaration of not including dependent types?  Did graphql?  Did.... for a farther reach, java?  No.  It might show up in blogs and comments on the internet afterward, and that's fine, but also a different form of information flow than putting it centrally in the docs of the thing itself.)  

You're right though -- at minimum, for how often this comes up, we would benefit from a simple declaration of goals and things-that-may-be-a-nice-idea-but-boy-there-is-only-so-much-complexity-budget.  Or, it certainly deserves its own little file in our exploration and design history notes, here.  (And thanks for the [comment below](https://github.com/ipld/specs/pull/139#issuecomment-509447225) that has some lovely historical references for getting that rolling!)

warpfork:
I'll probably move to merge this in a day or so if there's no objections.  I don't foresee doing any more incremental edits on this -- it's not the final chapter on this subject, but it's been a good exploration, and I'm done with it and wanna park it.

PR #141, opened 2019-07-10
Move exploratory specifications to directory

mikeal:
I also renamed the index for schemas to `README.md` and bumped the stage to draft. 

Even though improvements are still being made, schemas are relatively stable now, as the base syntax and concepts stabilized some time ago. The remaining changes we plan to make seem to fit well with the “draft stage.”


hannahhoward:
Suggestion: selectors is moving a lot, I think they it's exploratory

mikeal:
@hannahhoward good idea, are there any unmerged PR’s we should finish out before we move?

warpfork:
#### Re: selectors:

Let's... leave those be for this PR, if that's alright.  Those are kind of a mixed bag.

Some of the earlier rounds, such as [PR 116](https://github.com/ipld/specs/pull/116#issuecomment-500041656) and then a bunch of intermediates I only hoisted as gists, absolutely would've belonged as exploratory stuff (and I hoisted them as gists because we hadn't started this pattern yet!), as well as lots of earlier stuff, especially when there were multiple parallel nascent proposals.

At this point though (#133 and beyond), they're oscillating in towards draft level, I think: we're aiming to implement a real not-just-a-PoC next and there's no comparably fleshed-out alternatives in flight.

And yet there's a bunch of stuff in those files that's still from earlier exploratory thought and doesn't belong very much in the final spec-track document.  So I agree maybe that stuff should get re-homed now -- I've been looking at that file and trying to figure out what to do with it in fact, so, this is a good idea.  To get it there, though: Maybe let's leave it where it be now, and I'll do a PR gingerly splitting it up (possibly reviving some of the older files as a snapshot as well) soon?

warpfork:
#### folders

I'd consider moving the `exploratory/README.md` file to `design-history/exploration-reports/README.md` (or just `design-history/README.md` with a pointer to the exploration directory beneath)?

Or moving those existing folders up one.  Potato potato.

FWIW, the thinking that led me to make two folders deep earlier: I figured we'd want to have exactly _one_ folder at the top of the repo that's for early stuff -- and everything else would be for live, current, at-least-draft and/or final material -- and then the second depth of folder was for if we came up with ideas of other kinds of note we want to keep that aren't that exploratory report style.

#### suggesting structure

Should we add a short blurb to the README for this dir that suggests some naming structure for the files inside?

IMO the thing that's really major :key: is to make sure we set things up so that we can have one *or more* exploratory drafts up at the same time -- proposing a new exploration shouldn't "take up a slot" or produce an energy barrier that another exploration has to *overcome* before being able to be shared.

In order to make that flow work, we might also want to create a short blurb about using a dated pattern in filenames -- or something!  (comparable: XEP numbers?) -- which gives us a very easy-to-reach for, intentionally *un*-semantic, un-emotive way of starting efforts that may-or-may-not be parallel.

(... despite just floating "XEP numbers", I actually strongly favor dates over an incrementing number.  Proposal number style does nothing to forestall continued attempts to append necrophilic proposals and carry them across some sort of finish line; by contrast, dates encourage writing fresh takes, which may or may not choose to lift the good ideas from earlier rounds.  The latter is substantially better, IMO.  Proposal numbers can work fine as a suffixer within one train of thought -- rc1, rc2, etc -- but they're a bad idea to use as the primary key.)

mikeal:
> Or moving those existing folders up one. Potato potato.

What do you think about moving `design-history` into `exploratory`?

warpfork:
>  What do you think about moving `design-history` into `exploratory`?

Also potato potato, I guess?

I kinda like having the word "history" be in the dir we see in the root.  Makes it clear instantaneously that this is _not_ where you're gonna find the latest most authoritative stuff.  "exploratory" arguably communicates that as well, but in my brain, not quite as intensely.  I dunno.  /2c.  I'm pretty much at "semantic satiation" for the actual words.

warpfork:
To make this the least complicated to merge, I'd suggest:

- move files in #126 to existing directories; merge it
- move readme in this PR to existing directories; merge it
- I need to finish some detangling in #142 which already uses existing directories; merge it
- ... *then* if we want to use new folders, open a bikeshed for it?

I really don't care one way or the other, but we gotta stop with this thing where we propose a new name for a thing without enumerating issues with the old name.  Churn with a reason is fine.  Churn for churn's sake gets old fast.  Introducing this with new paths just seems like the oddest possible sequencing of diffs.

README.md
```diff
@@ -123,6 +123,7 @@ Specifications:
 
  * Exploratory Stage
    * Authors can merge changes into exploratory specifications at their own discretion
+   * Exploratory specifications live in the `exploratory` directory until they reach the draft stage.
```
rvagg: link

rvagg (commented):
Can't bring myself to register a strong opinion on this sorry, -0 if anything, I'm really not looking forward to this hugely cluttered desgin-history and/or exploratory directory thing. Most of what we are working on right now is exploratory tbh.

vmx:
> I kinda like having the word "history" be in the dir we see in the root.

I'm with @warpfork here. I think having a single dir with "history" in its name is a good indicator that you shouldn't look into it, unless you know what you're doing.

mikeal:
I have an issue with putting things in a directory marked “history” that are still under heavy and active development after they are merged. It miss-communicates the state of the documents.

vmx:
@mikeal I think I misunderstood. I thought the exploration things will end up in the "history" directory once they are fully explored and finished. During exploration it doesn't make much sense to put them inside such a dir.

mikeal:
after a long talk with @warpfork I think the best thing to do is use a `design` directory for exploratory work w/ a `history` directory in it for what we’ve been doing as far as documenting design thinking.

also, the readme should note that exploratory specs include a `-[Letter]` like `-A` to note that they are still under heavy discussion and alternatives can be proposed as alternative specs OR changes to the existing specs.

mikeal:
ok, pushed a rename to `design` and it seems much nicer. also moved the composites spec and added the required `-A` to it.

rvagg:
I've just noticed we have a selectors/ top level directory again now, although I can't quite figure out which PR did that. Can we find a directory to put these layer-spanning specifications rather than cluttering up the top level more? Some more hierarchy would be helpful here.

mikeal:
@rvagg good point, but based on prior comments I’m going to leave selectors out of this PR. we can address in another PR.

warpfork:
... I don't really understand what happened in git in this PR.  It says there's thirteen commits here, but most of them are already on master but showing up here (and cluttering the diff) anyway...?  Is this something that can be cleaned up?

vmx:
@warpfork A `git rebase master` and force push with lease did the trick.

warpfork (approved):

warpfork:
... I'm gonna see what needs be done about this one conflict github is warning us has emerged, do it, and get this merged.  Seems like we're close enough to consensus that this is appropriate, and I suspect we're going to be making a lot of new links to this stuff soon, so I want it done before the move accumulates any _more_ future link-fixing debt.

PR #142, opened 2019-07-12
Factor out Selector design history from present.

warpfork:
Extricate design history of Selectors from the present state, as per triggered by discussion in https://github.com/ipld/specs/pull/141#issuecomment-510777921 .

Recent discussion has highlighted the fact that the main file describing Selectors has grown a bit unwieldy, and contains a confusing mixture of present-tense fact as well as historical (and now often resolved!) questions and design thought.  Simultaneously, there's lots of good content in those earlier designs, and we'd like to still be able to easily link to them.

Therefore, this PR attempts to extricate the two; and moves the historical and more exploratory design goals and discussion to the  `./design-history` folder.  I've also added some extra connective tissue of linking to various other in-progress and evolutionary phases, just so they're a bit more discoverable.  (I did enough splunking through git history that I figured I'd might as well save an index of it for anyone going through the trouble again!  Especially since it turns out `git log` will fail to consider the newly separated file a "move" or `-F`-followable through history.)

### DIFF
--- a/design-history/exploration-reports/2018.10-selectors-design-goals.md
+++ b/design-history/exploration-reports/2018.10-selectors-design-goals.md
@@ -1,0 +1,139 @@
+# IPLD Selectors
+
+This document is a designdoc for IPLD Selectors.
+
+## [Meta: Status of this doc]
+
+- This was written around 2018-10-16 ([video presentation](https://drive.google.com/file/d/1NbbVxZQFKXwW6mdodxgTaftsI8eID-c1/view))
+- It narrows down the decision space enough to make significant progress.
+- good enough for trying out an implementation to learn more and make choices.
+- But it is not complete.
+- some choices that need to be made:
+  - [ ] select general binary and string format structure for selectors. (options given here)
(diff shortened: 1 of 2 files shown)


warpfork:
I want to add a little more exposition back to the top of `selectors.md` before merging this -- it's pretty sparse right now as it stands with all of the earlier stuff scooped out.  It'll take me a day of editing though; there's gems in the existing text to pull from, but very little of it that doesn't need polishing and de-burring to remove historical and now-resolved questions from it.

vmx (approved):
Thanks a lot for taking the time cleaning this up. I think it's also a great idea to not only keep those things somewhere in the version history, but also as actual files.

hannahhoward (approved):
+1 All about clean specs!
And, great to have the historical record in a seperate file.

hannahhoward:
@warpfork can we merge this? I mean I can merge this myself, but it's your PR.

warpfork:
Added slightly more exposition -- I feel like much more could be done here, but you're right, it's better to merge what's done and keep moving.  Will do.

PR #146, opened 2019-07-25
restore the requirements doc

Stebalien:
This document isn't ancient history, it was written 8 months ago as a concise, guiding document for what IPLD must and must not do.

Please _read_ it.

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -1,0 +1,160 @@
+# IPLD Requirements
+
+This document outlines parts of IPLD that should and should not be changed to
+ensure the success of future improvements (especially type systems).
+
+**Definitions**
+
+* Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
+* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
+* Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
+* Link: A link is an IPLD Node that points to another IPLD Node.
+* Path: A paths a human readable pointer to an IPLD Node.
+
+## Linked
+
+IPLD must support linking to any IPLD node (even if the node is in the middle of
+a block). That is, IPLD must support arbitrary IPLD paths in links.
+
+**Motivation:** Considering this in the context of programming languages, not
+being able to *store* a pointer to a struct *inside* of another struct would be
+severely limiting.
+
+NOTE: We don't currently support arbitrary paths but, in the context of
+programming, we really need to.
+
+## Immutable
+
+IPLD links must be immutable. We'll likely define a mutable link spec on top of
+IPLD but there needs to be an immutable layer at the bottom.
+
+**Motivation:** *Having* an immutable layer is important for a lot of analysis,
+memoization, type checking, etc.
+
+## Multicodecs Are Not Types
+
+It's impossible to understand IPLD data at a *structural* level if we don't know
+the format. Therefore, we should avoid introducing new formats unnecessarily as
+*every* IPLD implementation needs to support these new formats.
+
+## No Non-Local Reasoning
+
+An IPLD block should never be interpreted in the context of *anything* not
+contained in the block (and CID).
+
+For example, assuming we add support for relative links, the following
+definition of `foo` would not be a valid IPLD block:
+
+```
+var foo = {
+  // points outside of the current block, into the parent's "baz" field.
+  "baz": {"/": "../../baz"}
+}
+var bar = {
+  "foo": CidOf(foo),
+  // `/foo/baz` points here.
+  "baz": "something"
+}
+
+// resolution throug block `foo` depends on block `bar`.
+Resolve("/ipld/${CidOf(bar)}/foo/baz/")
+```
+
+For the same reason, IPLD links can't rely on an authority (e.g., a blockchain).
+
+Note: Links like this can still be encoded at the application level but they
+won't be handled by the IPLD resolver (and won't get the special "link" type).
+
+**Motivation:** IPLD needs to be easy to reason about.
+
+**Caveat:**
+
+We *may* want to relax this if we want to move schemas into separate,
+deduplicated blocks (referenced by CID). If we do that, we'd need to fetch a
+block's schema before being able to interpret the it.
+
+However, we need to *thoroughly* discuss any changes to this requirement.
+
+1. The space savings may not be worth it given the size of CIDs (>40 bytes),
+   compression, smart transports, and smart datastores.
+2. This change would introduce some weird interface complexities and potential
+   network dependencies.
+
+## No Cycles
+
+IPLD links must not be cyclic, even if we add support for relative links.
+
+**Motivation:**
+
+1. Security: Forbidding link cycles ensures that any graph traversal terminates.
+   This makes it easier to correctly and securely implement some graph
+   algorithms, even on potentially untrusted data in a distributed system.
+2. Consistency: Without complex hacks, it's impossible to create link cycles
+   *between* hash-linked blocks. Allowing link cycles one one level (e.g., in
+   relative links within a single block) but not on a larger scale (between
+   blocks) is inconsistent (even though there's nothing we can do about it).
+
+## Stable Pathing
+
+An IPLD path always means the same thing, everywhere, every time. Importantly,
+users shouldn't be able to configure their IPLD library to change how path
+resolution works. They should be able to use alternative path resolution
+algorithms for *non*-IPLD paths (e.g., IPFS paths) but those should be built
+on top of IPLD.
+
+**Motivation:** Deterministic computations on top of a IPLD need to produce the
+same result every time.
+
+## Link Transparent Pathing
+
+Path resolution must transparently traverse links.
+
+**Motivation:** Deduplication and composability. If links are *not* transparent,
+programmers will have to either make many small objects (lots of hashing) or
+inline data into large objects (lots of duplication and copying).
+
+## Primitives
+
+The "recommended" IPLD format (currently DagCBOR) needs to support *at a minimum*:
+
+* 32/64 bit integers without losing information.
+* 32/64 bit floats without losing information.
+* Unicode strings.
+* Binary strings.
+* Objects (with string keys, at least).
+* Arrays.
+* Booleans.
+* A bottom type (null).
+
+**Motivation:** Convenience, really.
+
+## Non-Cyclic, Block-Local Relative Links
+
+That is, relative links that don't traverse out the back of an object. See the
+conclusions from: [#1](https://github.com/ipld/specs/issues/1).
+
+**Motivation:** This is required to efficiently represent a highly connected DAG
+of tiny nodes.
+
+**Caveat:** This brings in some sticky issues around mutability. Depending on
+the implementation, relative links within an object may be act like mutable
+links (from the perspective of the user). The concern here is that we don't want
+users to bundle nodes together into single block *because* they want this
+mutability.
+
+# To Do
+
+Working through this, I realized we have a few things we really need to finish a few things before we can
+call IPLD ready.
+
+* **Path links.** Pointers that can only point to objects at block boundaries
+  are useful but severely gimped. We've been fine up till now because we
+  generally don't *edit* complicated datastructures but this will change.
+  ([#83](https://github.com/ipld/specs/issues/83))
+* **Slice links.** For the same reason, we really should support
+  `/ipld/QmID/start..stop` as a syntax for slicing an array. Most programming
+  languages support this so *not* supporting it would be a bit awkward.
+  ([#84](https://github.com/ipld/specs/issues/84))
+* **Link Spec.** We need to specify a complete and formal link spec and stick
+  with it. 
+* **Relative Links.** [#1](https://github.com/ipld/specs/issues/1).


Stebalien:
Partially reverts https://github.com/ipld/specs/pull/125.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Link Transparent Pathing
+
+Path resolution must transparently traverse links.
+
+**Motivation:** Deduplication and composability. If links are *not* transparent,
+programmers will have to either make many small objects (lots of hashing) or
+inline data into large objects (lots of duplication and copying).
+
+## Primitives
```
vmx, line 116: We could refer to the [Data Model spec](https://github.com/ipld/specs/blob/8d7dbe757a1b42c5bb3704d29cf2fffdf8639d4d/data-model-layer/data-model.md).

vmx (commented):
I'm not sure if it is the right things to comment on this PR or if we should rather merge it and then modify it.

I agree with @Stebalien that it contains a lot of information that isn't found in other documents yet. I'm not sure if `requirements` is a good name, but I'm OK with it.

Stebalien:
For additional context, this document is the output of https://docs.google.com/spreadsheets/d/1Lfd91hi3nFlLRS1r-FHvK2ip2Ll6ukraufCgepw43bw/edit#gid=35874512&range=23:23. We could also just call it `design-constraints.md`.

rvagg (approved):
Sorry, my bad!
When I read this it didn't seem to connect very well with more recent activity and it's really just a dangling document (and there is a lot of redundant info in here, so my original reasoning isn't completely bunk). I don't mind so much what it's called but some additional context would be great to explain its purpose and how it connects to everything else. A few lines & link in the README would be great, otherwise it'll just be dangling and the next new person may have the same experience.

Stebalien:
You're right, there's nothing there explaining this document and that's my fault. I'll add an explanation and a link from the README.

*edit: although I'm not really sure _where_ that should go*

> and there is a lot of redundant info in here

This document was an initial attempt to pull constraints from all over the IPLD repos into one central place so people could check new design ideas against previous design decisions/rationals. But there are several decisions around pathing, cycles, etc. that aren't replicated.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
+# IPLD Requirements
+
+This document outlines parts of IPLD that should and should not be changed to
+ensure the success of future improvements (especially type systems).
+
+**Definitions**
+
+* Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
+* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
```
warpfork, line 9: I don't think we've had much cause to use this term recently.  Node and Block come up constantly.  Fragment hasn't... at all, really, in my working memory; or if it does, it's been colloquial and not in a way that really seems to deserve a glossary entry.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
+# IPLD Requirements
+
+This document outlines parts of IPLD that should and should not be changed to
+ensure the success of future improvements (especially type systems).
```
warpfork, line 4: I'd like to rehash this sentence slightly somehow.  (Starting with dropping the parenthetical about type systems; IMO that doesn't need special emphasis.)  Probably this goes in the same vein as @vmx's comment about the document title as a whole.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Definitions**
+
+* Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
+* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
+* Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
+* Link: A link is an IPLD Node that points to another IPLD Node.
+* Path: A paths a human readable pointer to an IPLD Node.
```
warpfork, line 12: This definition should expand to say something about "segments".  Paths having segments, which each segment can be popped off and it still refers to a Node, seems to be a fairly critical concept.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+IPLD but there needs to be an immutable layer at the bottom.
+
+**Motivation:** *Having* an immutable layer is important for a lot of analysis,
+memoization, type checking, etc.
+
+## Multicodecs Are Not Types
+
+It's impossible to understand IPLD data at a *structural* level if we don't know
+the format. Therefore, we should avoid introducing new formats unnecessarily as
+*every* IPLD implementation needs to support these new formats.
```
warpfork, line 38: This is a very pleasing paragraph, but an odd title (it alludes to a definition of "types" which isn't offered anywhere in scope).  Maybe a little massaging could make it better -- "Multicodecs are not meant to offer types", "Multicodecs are only for codecs", something in that vein which makes it seem less like "types" is a specific proper noun?

The text about "structural level" seems to resonate strongly with the concept we now call "data model", and maybe we can again link that.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+
+## No Non-Local Reasoning
+
+An IPLD block should never be interpreted in the context of *anything* not
+contained in the block (and CID).
```
warpfork, line 43: I think we *do* actually want to go more with the "caveat" under this heading.  We absolutely *know* that there are things that can and will use contextual information to get work done; our challenge is only in how to describe this.

A more confined statement might be:  "Parsing an IPLD block into the IPLD Data Model should never require any additional information not contained in the block and CID".  It would also follow from that statement that IPLD Paths must always work over the Data Model must do so without addition information outside of the block and CID.  But it doesn't rule out that higher levels of application can interpret things contextually (which is important, because that's like trying to outlaw sunshine; the big ball of fusion disagrees -- and we have so, so many examples of this already: the unixfsv2 symlinks discussion, for example).

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+
+For example, assuming we add support for relative links, the following
+definition of `foo` would not be a valid IPLD block:
+
+```
+var foo = {
+  // points outside of the current block, into the parent's "baz" field.
+  "baz": {"/": "../../baz"}
```
warpfork, line 51: This is a tetchy example, because although I see what it's trying to say, as commented above, we have to be careful to be clear that an application having "../.." strings which it munges in some way is *fine* (see unixfsv2 symlinks); it's just that we'll refuse to parse such cyclic references for *the purpose of IPLD Paths*.

Although this example is trying to get at that, I think it needs a lot of clarification.  It's a subtle point.  (And I think we can observe this subtle but important point having a tendency to derail things badly and recurringly if it's lost (again, unixfsv2 symlinks, among other conversations...)).

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+Note: Links like this can still be encoded at the application level but they
+won't be handled by the IPLD resolver (and won't get the special "link" type).
+
+**Motivation:** IPLD needs to be easy to reason about.
+
+**Caveat:**
+
+We *may* want to relax this if we want to move schemas into separate,
+deduplicated blocks (referenced by CID). If we do that, we'd need to fetch a
+block's schema before being able to interpret the it.
```
warpfork, line 74: I think we need to upgrade this block from "caveat" to statement of fact.  We want to do this.  We are doing it.  (And with apologies for the risk of repeating myself: it's a phenomenon for more that the IPLD Schemas work: contextual information is simply a fact of life in many other applications like Advanced Data Layouts such as HAMTs, and in application logic like symlinks in unixfsv2, and so on, and so on.)

But we've also carved the way to do this without contradicting any of the other goals :tada: -- we can always interpret something *as far as the Data Model layer* (perhaps aka "interpret it structurally", to try on some parlance from elsewhere in the document), and do that without schemas.  Even, critically, in data that's *also* meant to be used with schemas or ADLs or application logic that uses contextual concepts, etc.

E.g., we can still have replication and hashing and DAG traversal and all the rest of the important promises of the IPLD Data Model regardless of schemas or other context.  And that's the important part.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+   compression, smart transports, and smart datastores.
+2. This change would introduce some weird interface complexities and potential
+   network dependencies.
+
+## No Cycles
+
+IPLD links must not be cyclic, even if we add support for relative links.
+
+**Motivation:**
+
```
warpfork, line 88: I kinda want to add an item  `0.` here which is just called "Nice Properties: DAGs are algorithmically pleasing to work with :)" haha...

(But really.)

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+   relative links within a single block) but not on a larger scale (between
+   blocks) is inconsistent (even though there's nothing we can do about it).
+
+## Stable Pathing
+
+An IPLD path always means the same thing, everywhere, every time. Importantly,
+users shouldn't be able to configure their IPLD library to change how path
+resolution works. They should be able to use alternative path resolution
+algorithms for *non*-IPLD paths (e.g., IPFS paths) but those should be built
+on top of IPLD.
```
warpfork, line 103: This is an interesting one.

I think some of our recent discussions include the idea that we might want to do "pathing" over data structures which use Advanced Data Layouts, such as HAMTs.  I also want to be able to "path" using struct field names when handling data with a Schema, even if the representation is tuple (e.g. becomes a list when encoded, and thus has indexes instead of string keys).  Nobody wants to dispute that "stable pathing" is important and useful, I think; but these other features are *also* distinctly important and useful too.  What to do?

Perhaps what we're seeing is another instance where we could benefit from a split, wherein the strictest form of this rule applies at the Data Model layer; and the spirit of the rule sticks around, but the technicalities are a bit relaxed with some of the other higher-level tools in the core ecosystem.

---

Another way to look at this is that the "alternative path resolution algorithms [...] (e.g., IPFS paths)" are now something that we can describe *slightly* more generally: e.g. we expect to be able to make "IPFS" paths work with a combination of the {IPLD Data Model + ADLs such as HAMT + a Schema saying where to expect it all}, and while that's then an "IPFS" application choice-set, it requires very very little glue code (in fact none: all this is 'declarative', except for any logic inside the ADL, and even that we're making spec-first/declarative-ish) to connect, since we've made these components reusable and composable through interfaces we had for IPLD anyway (e.g. Node).

REQUIREMENTS.md
warpfork: I *really* like this heading, aside from the nits.  We could use more of these "X is not solving purpose Y" headings like this -- they add meaningful clarity, and especially constrain scope-creep and defend against epistemologischewanderungsangst.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+* Arrays.
+* Booleans.
+* A bottom type (null).
+
+**Motivation:** Convenience, really.
+
+## Non-Cyclic, Block-Local Relative Links
+
+That is, relative links that don't traverse out the back of an object. See the
+conclusions from: [#1](https://github.com/ipld/specs/issues/1).
```
warpfork, line 134: This heading sounds more like a wishlist or an exploratory thought that something we're actually *doing* at present or immediately-anticipatable future (and the caveats section is... pretty serious) -- I think we should elide it from this document and maybe put something about it in an exploratory doc instead.  Whatever comes of this would need serious incubation.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+
+Working through this, I realized we have a few things we really need to finish a few things before we can
+call IPLD ready.
+
+* **Path links.** Pointers that can only point to objects at block boundaries
+  are useful but severely gimped. We've been fine up till now because we
+  generally don't *edit* complicated datastructures but this will change.
+  ([#83](https://github.com/ipld/specs/issues/83))
+* **Slice links.** For the same reason, we really should support
+  `/ipld/QmID/start..stop` as a syntax for slicing an array. Most programming
```
warpfork, line 155: I think we can drop this from here.  We have Selectors to do this now.

warpfork: Which reminds me: I'd actually like to propose clarifying that *Links* go to **one** Node.

Much of the tower of abstractions either come tumbling down or becomes very alarmingly baroque without this rule... so much so that I think we assume this rule already.  We should say it.

warpfork (commented):
We do need some more documents like this to keep a central definition of direction, so I'm really happy to look at this diff.  I also have a lot of comments.

---

There are a couple of other holistic (vibe, @rvagg) things I'd like to capture after reading this:
primarily, that IPL**D** is about **D**ata.  That seems to underlie all of these points at a very fundamental level but hasn't been called out very explicitly.
I think it might be productive to try to make this more explicit.
(Admittedly, finding words to do this clearly is *hard*.)

For example, in the heading about Stable Pathing, there's a segment of text that says
"An IPLD path always means the same thing, everywhere, every time."
This is interesting, because A) of course!  B) at the same time, we can think of
counter-examples to this which we seem to agree are useful enough that the rule bends.
Advanced Data Layouts like HAMTs are proposed to expose themselves as a map,
which allows them to be pathed across as if they're a single node, for example; and we seem to feel that that's reasonable, and that making it an acknowledged part of IPLD provides huge value and fits the existing abstractions reasonably well.  (See comment at https://github.com/ipld/specs/pull/146#discussion_r308740520 which also explores this.)

This nailing down of the core vibe is important for saying how well something "fits" under the umbrella of IPLD or not.

It seems that many of the key contracts that we want from IPLD are satisfied (at their strictest (which is good)) at the Data Model layer; and, we've also got several features that we consider very very important and useful (Selectors; Advanced Data Layouts (like HAMTs); Schemas) to the overall ecosystem being productive...
and these latter systems generally involve relaxing the contracts *a little* (such as: ADLs and Schemas alike seem to want to provide suggestions to pathing!).  Yet, these systems also still remain as true as possible to the general philosophy of declarative / data-driven / non-computing behaviors; and it seems that because of this adherence to the general philosophy, we're still comfortable acknowledging them as fitting under the umbrella of IPLD (and making them batteries-included in core libraries, e.g. as Selectors are in the new Go libraries, where they power very basic essential traversals).

My very general impression is "declarative: yes; computation: no".  (Maybe the `D` in IPLD also stands nicely for "declarative" as well as "data"!)  But are these things recognizable and clear enough?  How can we clarify this further?

warpfork:
>  I'm not sure if `requirements` is a good name,

"Cornerstones"?  The things that define what is and isn't going inside the floorplan.

REQUIREMENTS.md
Stebalien: Possibly. IIRC, I defined it because I needed it but I probably just ended up removing its use.

REQUIREMENTS.md
```diff
@@ -0,0 +1,160 @@
...
+* Path: A paths a human readable pointer to an IPLD Node.
+
+## Linked
+
+IPLD must support linking to any IPLD node (even if the node is in the middle of
+a block). That is, IPLD must support arbitrary IPLD paths in links.
+
+**Motivation:** Considering this in the context of programming languages, not
+being able to *store* a pointer to a struct *inside* of another struct would be
+severely limiting.
```
warpfork, line 21: This is actually a tricky subject.

Links that land on a Node in the middle of a Block (because they're a tuple of CID+Path, say) are easy _to read_.  They're basically meaningless if we wanted to try to do an _update_ of a value on the far side of such a Link and propagate it, as far as I can figure, which is a bit of a hurdle.

It'd be nice to have a paragraph extolling the virtues of linking first... and then tangoing with this tricky subject of links to mid-Block Nodes in a later heading.

rvagg:
OK, lots of good discussion here but it's not going to help make progress on this PR since it's a lot all at once. So I've got #149 that takes this commit, adds a note to the README and renames it to "IPLD Foundational Principles". Let's get some agreement over those small steps (README & naming) and then move the other possible changes to separate PRs for more contained discussion.

rvagg:
sorry, didn't see #148 before pushing #149, but my point still stands on large chunks of work - it might be more productive to bite off smaller pieces than trying to get consensus on fundamentals

warpfork:
I'm down for that.  I'll rebase #148 to come after #149 and put the diffs into that filename.  (I only refrained from a move in #148 because I didn't want it to steamroll the diff into a single large pair of red/greens.  "not move" after another PR that does the move also accomplishes that.)

rvagg:
merged in #149

REQUIREMENTS.md
warpfork: ... I should perhaps clarify that when I say "a bit of hurdle", what I mean is "this is such a huge barrier to clear definition that I think it's nonviable until resolved".

Sometimes I'm a master of understatement.

Stebalien: In most cases, I'd copy the node out of the block and update it as a new block. If I can use a path to point to something but I can't _store_ that pointer within IPLD, that pointer is pretty useless.

REQUIREMENTS.md
Stebalien: Slice links were a form of transformation and I agree that they may be beyond the scope of the IPLD data model. However, selectors aren't _quite_ right (unless we're treating them as transforms as well).

Note, slice links would create a "virtual" array (like a slice in go, rust, etc.). The link itself would still point to one "node" (the virtual array), not the elements.

REQUIREMENTS.md
Stebalien: This item is a requirement for what I'd consider to be a "feature complete" IPLD. Without it, we're forced to split nodes out into their own blocks if we want to point to them.

REQUIREMENTS.md
warpfork: So in this model, let's say I start with one block where the nodes inside contain a link to another block and then the link contains several path segments so we reach a mid-block node.  (So far this is all just setup statements.)

If I do an update, using my library that has a neat function which builds a new node for the updated value, and propagates it up copy-on-write style through all the values I pathed over to reach it, and includes replacing any links with updated link hashes...

... then the very first time I use this update feature, I'd end up rewriting the link to be a CID-only link, since I copied(/replaced) that reached node out of the block?

This seems like it would end up with a lot of data that... wouldn't use this feature, because it would tend to shake itself back out of the data anyway.  Links with path segments that reach a mid-block node would only ever be generated by some application which has opinions about that.  Maybe that's not a problem, but it seems interesting.

rvagg: I wonder is this a drawback of trying to abstract away the block entirely? If you retain the concept of block along with its constituent node then wouldn't this be a bit easier to reason about?

PR #147, opened 2019-07-30
Content-addressability is a concept.

warpfork:
It's useful to introduce related to the block/codec layer, because it's the bytes themselves that we hash in order to get our content addresses... but the concept itself is certainly recurring: for example, deterministic ordering remains important at many other levels for the same core conceptual reason.

So should we move it to 'concepts'?


vmx (approved):

rvagg (commented):
I'm not convinced but if you feel strongly enough then go ahead

PR #148, opened 2019-07-30
Iterate on requirements/cornerstones

warpfork:
This is mostly addressing the huge batch of comments I left in the
vicinity of https://github.com/ipld/specs/pull/146#pullrequestreview-268375431 .

Some of this text might be good; some might need more iteration.  In particular I also didn't yet make any thrust at addressing the "IPL**D** is about (**D**eclarative) **D**ata" key concept, which is _absolutely_ one of our critical cornerstones that underlies all of these other points.  I would still like to try to make this more explicit, though; help wanted.

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -1,27 +1,24 @@
-# IPLD Requirements
+# IPLD Cornerstones
 
 This document outlines parts of IPLD that should and should not be changed to
-ensure the success of future improvements (especially type systems).
+ensure the success of future improvements and the continuity of direction.
 
 **Definitions**
 
 * Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
-* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
 * Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
-* Link: A link is an IPLD Node that points to another IPLD Node.
-* Path: A paths a human readable pointer to an IPLD Node.
+  Many nodes can exist encoded inside one Block.
+* Link: A link is a kind of IPLD Node that points to another IPLD Node.
+* Path: A path is composed of segments which each specify a step across an IPLD Node.
 
 ## Linked
 
-IPLD must support linking to any IPLD node (even if the node is in the middle of
-a block). That is, IPLD must support arbitrary IPLD paths in links.
+The IPLD Data Model includes Links. A Link can be resolved to reach another IPLD Node.
 
-**Motivation:** Considering this in the context of programming languages, not
-being able to *store* a pointer to a struct *inside* of another struct would be
-severely limiting.
-
-NOTE: We don't currently support arbitrary paths but, in the context of
-programming, we really need to.
+**Motivation:** Linking makes it possible to build data structures which are
+theoretically unbounded in size, while still being traversable, consistent,
+authenticated and immutable.  This unlocks the potential for a host of
+decentralized applications and is part of IPLD's fundamental purpose.
 
 ## Immutable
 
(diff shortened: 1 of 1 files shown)


REQUIREMENTS.md
```diff
@@ -31,54 +28,89 @@ IPLD but there needs to be an immutable layer at the bottom.
...
-Note: Links like this can still be encoded at the application level but they
-won't be handled by the IPLD resolver (and won't get the special "link" type).
+**Note:** Concepts that seem similar to relative linking can still be encoded
+at the application level.  This is fine, but distinct from "IPLD Links", because
+such linking won't be interpreted by IPLD path and link resolution (e.g. they
+won't get the special "link" type, and won't violate the constraints that the
+IPLD Data Model expresses a DAG, etc).
 
-**Motivation:** IPLD needs to be easy to reason about.
+### Moving beyond local reasoning
```
vmx, line 84: Good idea making putting those things into a new section.

REQUIREMENTS.md
```diff
@@ -105,6 +137,34 @@ on top of IPLD.
...
+when a struct uses "tuple" representation, causing it to be transformed into
+a list rather than a map when encoded).  In these cases, we can "path" across
+data interpreted in context of a Schema using the field names, even if at the
+Data Model layer it's been represented as a list (and thus has indexes instead
+of map keys corresponding to the field names).  This kind of pathing can be
+stable and predictable, but (as with the Advanced Data Layouts story), that
+stability now requires more: holding the Schema declaration.
+
+Note that regular, core Data Model still maintains stable pathing even in these
+examples of higher level systems with alternative rules.
```
vmx, line 166: The original sections all have a "Motivation" which I quite like. Here the motivation could be about different views on the data like IPFS is doing it with UnixFS.

REQUIREMENTS.md
```diff
@@ -31,54 +28,89 @@ IPLD but there needs to be an immutable layer at the bottom.
...
+relaxed for these high-level systems, and the scope of "local" can be
+understood more broadly.
+
+Since we can always interpret block structurally (e.g., parse them at least to
+the Data Model layer) -- even in data that's also meant to be used with
+Advanced Data Layouts or Schemas other application logic that uses contextual
+concepts, etc -- we can still have replication and hashing and DAG traversal
+and all the rest of the important promises of the IPLD Data Model regardless of
+that other context, meaning these systems are purely value-add and do not
+compromise any of the other core promises of IPLD.
```
vmx, line 113: A possible "motivation" could be building higher level data structures (like HAMT) on top of the core IPLD Data Model.

vmx (commented):
I really like where this is going, good work. Thanks @warpfork (and @Stebalien for the original doc).

REQUIREMENTS.md
```diff
@@ -1,27 +1,24 @@
...
 **Definitions**
 
 * Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
-* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
 * Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
-* Link: A link is an IPLD Node that points to another IPLD Node.
-* Path: A paths a human readable pointer to an IPLD Node.
+  Many nodes can exist encoded inside one Block.
+* Link: A link is a kind of IPLD Node that points to another IPLD Node.
+* Path: A path is composed of segments which each specify a step across an IPLD Node.
```
lanzafame, line 12: I am struggling to grok `a step across an IPLD Node.` How does one step across a _point_? Would it be more accurate to say `step between IPLD Node(s)`?

lanzafame: Actually, I think it just clicked.
A Node is:
```
(A)
```
A Link is:
```
(A) -> (B)
    ^
   Link
```
A Path of `/A/B/C` represents:
```
(A) -> (B) -> (C)
 |_____________^
        ^
(B) gets stepped over 
```
So maybe `step over IPLD Node(s)` is a better way of saying this. @warpfork?

REQUIREMENTS.md
warpfork: Good idea.  I should keep that up.

REQUIREMENTS.md
warpfork: Yeah, I agree, this bit of text struggled.  "between" might do better.

I'm not sure about that diagram, because it's not necessary that it be a link in-between; rememeber that a whole tree of Nodes can be in a Block.

Maybe I should also break this down into bullet points for "Path" and "Path Segment", and that might help?  E.g. one Path Segment moves from one Node to the next (but only can describe movement of a single stride: e.g. looking up one map key, or stepping into an list by index).  A Path is just a collection of Path Segments, which we often use to describe a walk of a specific, er, well, *path* down a Node tree.

warpfork: Made another shot at this in: #152 -- does that one seems clearer to you?

REQUIREMENTS.md
```diff
@@ -31,54 +28,89 @@ IPLD but there needs to be an immutable layer at the bottom.
 **Motivation:** *Having* an immutable layer is important for a lot of analysis,
 memoization, type checking, etc.
 
-## Multicodecs Are Not Types
+## Multicodecs Are Not Meant to Act As Types
```
mikeal, line 31: This is great. Probably the most concise text I’ve read so far on “why you don’t need to write a new codec.”

Recently, I’ve run into some areas where I legitimately did need a new block format and codec for an application specific use case. Until I read this I didn’t have great language to distinguish *why* this case was different (it wasn’t about types, the motivation was compactness and limiting dependencies).

warpfork: Heh...  Yeah, we could almost make a "## Multicodec nonproliferation treaty" and put this motivation hunk under that.  "Multicodecs are not meant to act as types" could just have a "see above" for its motivation.

REQUIREMENTS.md
lanzafame: @warpfork Much better, made a note on a slight clarification.

warpfork:
Alright, closing this one -- It's stale, and it's been broken up into smaller PRs in... well, all the other PRs that can be seen referencing this one lately.

PR #149, opened 2019-08-01
REQUIREMENTS -> IPLD Foundational Principles

rvagg:
Some minor progress over #146, restore the doc, rename it and put a note in the README.

### DIFF
--- a/FOUNDATIONS.md
+++ b/FOUNDATIONS.md
@@ -1,4 +1,4 @@
-# IPLD Requirements
+# IPLD Foundational Principles
 
 This document outlines parts of IPLD that should and should not be changed to
 ensure the success of future improvements (especially type systems).
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@ IPLD Specifications
 
 The goal of IPLD is to enable decentralized data-structures that are universally addressable and linkable which in turn will enable more decentralized applications.
 These data-structures allow us to do for data what URLs and links did for HTML web pages.
+Read more about the principles that are guiding the ongoing development of IPLD in **[IPLD Foundational Principles](FOUNDATIONS.md)**.
 
 IPLD is not a single specification it is a set of specifications.
 Many of the specifications in IPLD are inter-dependent.
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -1,0 +1,160 @@
+# IPLD Requirements
+
+This document outlines parts of IPLD that should and should not be changed to
+ensure the success of future improvements (especially type systems).
+
+**Definitions**
+
+* Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
+* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
+* Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
+* Link: A link is an IPLD Node that points to another IPLD Node.
+* Path: A paths a human readable pointer to an IPLD Node.
+
+## Linked
+
+IPLD must support linking to any IPLD node (even if the node is in the middle of
+a block). That is, IPLD must support arbitrary IPLD paths in links.
+
+**Motivation:** Considering this in the context of programming languages, not
+being able to *store* a pointer to a struct *inside* of another struct would be
+severely limiting.
+
+NOTE: We don't currently support arbitrary paths but, in the context of
+programming, we really need to.
+
+## Immutable
+
+IPLD links must be immutable. We'll likely define a mutable link spec on top of
+IPLD but there needs to be an immutable layer at the bottom.
+
+**Motivation:** *Having* an immutable layer is important for a lot of analysis,
+memoization, type checking, etc.
+
+## Multicodecs Are Not Types
+
+It's impossible to understand IPLD data at a *structural* level if we don't know
+the format. Therefore, we should avoid introducing new formats unnecessarily as
+*every* IPLD implementation needs to support these new formats.
+
+## No Non-Local Reasoning
+
+An IPLD block should never be interpreted in the context of *anything* not
+contained in the block (and CID).
+
+For example, assuming we add support for relative links, the following
+definition of `foo` would not be a valid IPLD block:
+
+```
+var foo = {
+  // points outside of the current block, into the parent's "baz" field.
+  "baz": {"/": "../../baz"}
+}
+var bar = {
+  "foo": CidOf(foo),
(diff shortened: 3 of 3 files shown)


warpfork (approved):

warpfork:
I believe @vmx also already said his only question about #146 was whether it's more useful to "comment on this PR or if we should rather merge it and then modify it" -- so maybe let's consider it decided for the second one?, and go ahead and merge this.

rvagg:
want to contest "foundational principles" vs "cornerstones" vs "requirements" or is this acceptable as is?

warpfork:
No contest.  "Foundational principles" does it for me :+1:

FOUNDATIONS.md
```diff
@@ -0,0 +1,160 @@
+# IPLD Foundational Principles
+
+This document outlines parts of IPLD that should and should not be changed to
+ensure the success of future improvements (especially type systems).
+
+**Definitions**
+
+* Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
+* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
```
mikeal, line 9: What distinguishes a “fragment” from a “node?”

FOUNDATIONS.md
```diff
@@ -0,0 +1,160 @@
...
+* Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
+* Link: A link is an IPLD Node that points to another IPLD Node.
+* Path: A paths a human readable pointer to an IPLD Node.
+
+## Linked
+
+IPLD must support linking to any IPLD node (even if the node is in the middle of
+a block). That is, IPLD must support arbitrary IPLD paths in links.
```
mikeal, line 17: This isn’t always true. In “pure data model” we are only able to link by CID and not (CID + Path). Everything we’ve done above the data model (paths, selectors, composites, etc) support this though.

FOUNDATIONS.md
warpfork: Don't worry about that one ;) https://github.com/ipld/specs/pull/152

FOUNDATIONS.md
```diff
@@ -0,0 +1,160 @@
...
+IPLD but there needs to be an immutable layer at the bottom.
+
+**Motivation:** *Having* an immutable layer is important for a lot of analysis,
+memoization, type checking, etc.
+
+## Multicodecs Are Not Types
+
+It's impossible to understand IPLD data at a *structural* level if we don't know
+the format. Therefore, we should avoid introducing new formats unnecessarily as
+*every* IPLD implementation needs to support these new formats.
```
mikeal, line 38: I need to write a larger post about this, but I’ve become deeply skeptical that we can curb codec/block-format proliferation in a meaningful way. Given that, we probably need to find a solution to “supporting all these formats” in a more dynamic way.

FOUNDATIONS.md
```diff
@@ -0,0 +1,160 @@
...
+Note: Links like this can still be encoded at the application level but they
+won't be handled by the IPLD resolver (and won't get the special "link" type).
+
+**Motivation:** IPLD needs to be easy to reason about.
+
+**Caveat:**
+
+We *may* want to relax this if we want to move schemas into separate,
+deduplicated blocks (referenced by CID). If we do that, we'd need to fetch a
+block's schema before being able to interpret the it.
```
mikeal, line 74: I don’t actually see what the contradiction here is.

For one thing, if a block *without* a schema can be decoded at the block level then it’s always valid and understandable. The fact that schemas greatly *improve* the interpretation should not invalidate the previous statements.

I do think you may want to iterate a bit on the above statements regarding context. Data *inside* of a block is atomic and does not *rely* on outside context, but outside context can still be **applied** to the data within a block. Think about a file encoded in IPFS, you could look at the list of links to raw blocks and they contain a lot of atomic information on their own, they are an ordered list of raw binary chunks, but the fact that they represent a file isn’t contained *in* that data, the only context in which you can interpret them as parts of a file is from having read the file metadata which points to this list of raw binary chunks.

Like so much with DAGs, context can only be applied in one direction ;)

FOUNDATIONS.md
```diff
@@ -0,0 +1,160 @@
...
+   This makes it easier to correctly and securely implement some graph
+   algorithms, even on potentially untrusted data in a distributed system.
+2. Consistency: Without complex hacks, it's impossible to create link cycles
+   *between* hash-linked blocks. Allowing link cycles one one level (e.g., in
+   relative links within a single block) but not on a larger scale (between
+   blocks) is inconsistent (even though there's nothing we can do about it).
+
+## Stable Pathing
+
+An IPLD path always means the same thing, everywhere, every time. Importantly,
```
mikeal, line 99: We have some prior verbiage about “path levels” that we should probably get rid of, but I’m interpreting this statement as belonging to “path level zero” only. Meaning, these are the raw paths through only the IPLD data model.

FOUNDATIONS.md
```diff
@@ -0,0 +1,160 @@
...
+# To Do
+
+Working through this, I realized we have a few things we really need to finish a few things before we can
+call IPLD ready.
+
+* **Path links.** Pointers that can only point to objects at block boundaries
+  are useful but severely gimped. We've been fine up till now because we
+  generally don't *edit* complicated datastructures but this will change.
+  ([#83](https://github.com/ipld/specs/issues/83))
+* **Slice links.** For the same reason, we really should support
```
mikeal, line 154: At what point does this just become selectors?

FOUNDATIONS.md
warpfork: Yeah, I tried to tear into some of that in https://github.com/ipld/specs/pull/156 .  Does that read better?

I like this example about files as well, maybe we should roll that into the diff as well.

> For one thing, if a block *without* a schema can be decoded at the block level then it’s always valid and understandable. The fact that schemas greatly *improve* the interpretation should not invalidate the previous statements.

:heavy_plus_sign: :100: 

> Like so much with DAGs, context can only be applied in one direction ;)

Super good statement.  I've been trying to come up with a way to phrase that for a while now, and I like this statement of it.

Some of the other negative examples in this vicinity use examples of cyclic links, and while we can all agree *those* are problematic, it's no argument against accumulating understanding as we follow *down* a DAG!

mikeal: Ah, yes, the new PR looks great. Sorry, my email backlog is in chronological order ;)

warpfork: :heart:

PR #152, opened 2019-08-01
Tweak foundations definitions: path segments, etc.

warpfork:
Dropped mention of "Fragments".  We don't seem to use that term much.  It's not a bad term; it just doesn't seem to have reached the frequency of usage which makes it seem important to include in the  glossary of most critical terms.

Inserted the word "kind" into the description of links.

Reworked "Path" to describe "Path" and "Path Segment" separately. This bit is a bit touchy to phrase well, and has already gone through some massaging in another PR; hopefully we're getting *closer*.

Extracts some content from #148, and regards earlier review in https://github.com/ipld/specs/pull/146#discussion_r308709312 , and also incorporates a bit of feedback from https://github.com/ipld/specs/pull/148/files/bea80f71e78fe97799526b1414931904010a397b#r309488666 .

### DIFF
--- a/FOUNDATIONS.md
+++ b/FOUNDATIONS.md
@@ -5,13 +5,16 @@ ensure the success of future improvements (especially type systems).
 
 **Definitions**
 
-* Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
-* Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
+* Block: A block is a chunk of an IPLD DAG, encoded in a format.
+  Blocks have [CID](./block-layer/CID.md)s.
+* Node: A node is a *point* in an IPLD DAG (map, list, number, etc.).
   Many nodes can exist encoded inside one Block.
 * Link: A link is a kind of IPLD Node that points to another IPLD Node.
-* Path Segment: A path segment a piece of information that describes a move from
-  one Node to another Node (effectively, either a map key or a list index).
-  A path segment only describes movement of a single stride.
+  Links are what make IPLD data a DAG rather than only a tree.
+  Links are content-addressable -- see [CID](./block-layer/CID.md).
+* Path Segment: A path segment is a piece of information that describes a move
+  from one Node to a directly connected child Node.  (In other words, a
+  Path Segment is either a map key or a list index.)
 * Path: A path is composed of Path Segments, thereby describing a traversal
   from one Node to another Node somewhere deeper in the DAG.
 
--- a/FOUNDATIONS.md
+++ b/FOUNDATIONS.md
@@ -6,10 +6,14 @@ ensure the success of future improvements (especially type systems).
 **Definitions**
 
 * Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
-* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
 * Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
-* Link: A link is an IPLD Node that points to another IPLD Node.
-* Path: A paths a human readable pointer to an IPLD Node.
+  Many nodes can exist encoded inside one Block.
+* Link: A link is a kind of IPLD Node that points to another IPLD Node.
+* Path Segment: A path segment a piece of information that describes a move from
+  one Node to another Node (effectively, either a map key or a list index).
+  A path segment only describes movement of a single stride.
+* Path: A path is composed of Path Segments, thereby describing a traversal
+  from one Node to another Node somewhere deeper in the DAG.
 
 ## Linked
 


FOUNDATIONS.md
```diff
@@ -6,10 +6,14 @@ ensure the success of future improvements (especially type systems).
...
 * Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
-* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
 * Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
-* Link: A link is an IPLD Node that points to another IPLD Node.
-* Path: A paths a human readable pointer to an IPLD Node.
+  Many nodes can exist encoded inside one Block.
+* Link: A link is a kind of IPLD Node that points to another IPLD Node.
+* Path Segment: A path segment a piece of information that describes a move from
+  one Node to another Node (effectively, either a map key or a list index).
+  A path segment only describes movement of a single stride.
```
vmx: I had to look up "stride". So perhaps it could be "step"?

FOUNDATIONS.md
```diff
@@ -6,10 +6,14 @@ ensure the success of future improvements (especially type systems).
...
 
 * Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
-* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
 * Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
-* Link: A link is an IPLD Node that points to another IPLD Node.
-* Path: A paths a human readable pointer to an IPLD Node.
+  Many nodes can exist encoded inside one Block.
+* Link: A link is a kind of IPLD Node that points to another IPLD Node.
+* Path Segment: A path segment a piece of information that describes a move from
+  one Node to another Node (effectively, either a map key or a list index).
```
lanzafame:
```suggestion
  the current Node to another Node (effectively, either a map key or a list index).
```
A **single** Path Segment defines a move **relative** to the current position.

lanzafame (commented):
Other than the one comment, this is a much clearer. I also agree with @vmx on changing stride to step.

FOUNDATIONS.md
```diff
@@ -6,10 +6,14 @@ ensure the success of future improvements (especially type systems).
 **Definitions**
 
 * Block: A block is a chunk of an IPLD DAG, encoded in a format. Blocks have CIDs.
-* Fragment: A piece of an IPLD DAG. Blocks contain fragments.
 * Node: A node is a *point* in an IPLD DAG (object, array, number, etc.).
-* Link: A link is an IPLD Node that points to another IPLD Node.
-* Path: A paths a human readable pointer to an IPLD Node.
+  Many nodes can exist encoded inside one Block.
+* Link: A link is a kind of IPLD Node that points to another IPLD Node.
```
rvagg, line 12: For the sake of the unfamiliar reader, would it be ok to add reference to CID in here? We do have the spec in this repo after all.

> "Currently an Link is represented as a [CID](https://github.com/ipld/specs/blob/master/block-layer/CID.md) although additional Link representations may be introduced in future."

FOUNDATIONS.md
rvagg: I'm not sure "the current" is necessary, if a Path is made up of many segments, they each relate to different Nodes, not "the current". But it does need to be tightened so it doesn't sound like arbitrary hops constitute a "Path Segment". Maybe "one Node to an adjacent Node", or "one Node to a directly connected Node"?

Also .. are we leaving the advanced layout case out of it for now, pretending that Paths only operate on the data model and that's all?

rvagg: Re last point about advanced layouts addressed in #157 but probably suggests that #157 _shouldn't_ be entirely in a separate doc, will comment there.

vmx: > if a Path is made up of many segments, they each relate to different Nodes

Yes, but each Path Segment itself points from its current node to another. So I'm in favour of "the current Node".

lanzafame: So in the context of a `Path Segment` (highlighting the singular), you have `./a`, which describes the move from `.` (the current node) to `a`, the root being where we are currently with a relative movement to `a`.

warpfork: I also think "current" is a little weird without having been introduced, but I'll massage this a bit more to try to incorporate the idea.

warpfork: Another thing that would be nice to capture here is that the step is, er, "downward".  E.g., `".."` is not a path segment that means "go up one", even though people might expect that because of the history of unixy filesystemy behavior.  But this is also tricking without having introduced "up" vs "down" directions.

It seems like there's a limit on how much meaning we can pack into these short introducers here.  We should probably start thinking about what the tersest expressions of these concepts are, and then linking to fuller specs elsewhere (maybe even later in the same document, if we're trying to make this one fairly freestanding).

FOUNDATIONS.md
warpfork: Hm.  You know, my first reaction was "yes", but then on further thought, I almost wanted to push back on this and say "the concept of a link should be clear without referring to the implementation details of CIDs".

But we already have the phrase "blocks have CIDs" just a few lines above, so... might as well be consistent about it, in that case.  Added it.

Stebalien (approved):

vmx (approved):

rvagg (approved):

lanzafame (approved):

warpfork:
I've never had so many checkmarks on a PR before.  Bless you all  :heart_decoration:

PR #154, opened 2019-08-01
Tweak foundations positioning of multicodecs.

warpfork:
Previous paragraph was good; this diff just updates it to use more references and an explicit call-out to the concept of the Data Model (which seems to be basically the same thing previously discussed and was already alluded to as "structurally" understanding the data).

Extracts some content from https://github.com/ipld/specs/pull/148 and regards earlier review in https://github.com/ipld/specs/pull/146#discussion_r308710837 .

### DIFF
--- a/FOUNDATIONS.md
+++ b/FOUNDATIONS.md
@@ -31,11 +31,22 @@ IPLD but there needs to be an immutable layer at the bottom.
 **Motivation:** *Having* an immutable layer is important for a lot of analysis,
 memoization, type checking, etc.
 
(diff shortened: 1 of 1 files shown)


vmx (approved):

rvagg (approved):

warpfork:
Okay, plus a thumbs-upping sound from Mikeal over in https://github.com/ipld/specs/pull/148/files#r309795339 this one seems like a slam dunk at least -- mergin'

PR #155, opened 2019-08-01
Rename to lowercase filename

vmx:
I find lowercase filenames easier to read and also to use on a terminal.

PS: There's no reason to scream.


warpfork:
I'm indifferent.  IMO when things are "about as important as the README itself", then they can capslock scream.  Therefore I'm fine with this one.

At this concrete moment especially though: plz no -- I have four open PRs against this filename right now and three more in unpushed draft to come later today.

We should get in the habit of doing renames as issues first, not PRs, and ideally I'd even propose writing them up as a series of `mv` and `sed` commands, because that makes them easy to apply later.  Rebasing rename PRs (or other PRs around them) sucks *mightily*.

vmx:
I'm closing this one and opening a new one, once the situation around the foundational document has been calmed down a bit. Though in general I prefer lowercase names. To me README is only uppercase due to historic reasons (whatever these might be).

PR #156, opened 2019-08-01
Refine foundational statements about nonlocality.

warpfork:
This extracts some content from #148, and regards earlier review in https://github.com/ipld/specs/pull/146#discussion_r308713535 and https://github.com/ipld/specs/pull/146#discussion_r308718009 .

The example is also substantially rewritten, to attempt to clarify two things: that it's a *negative* example (!!), and also to make it more obvious what we're proposing is a link (since the dag-json syntax isn't introduced anywhere in this immediate vicinity, and it can be a bit non-obvious what's going on if that wasn't in your mind's eye already as you were reading).  This addresses https://github.com/ipld/specs/pull/146#discussion_r308715053 .

From the earlier draft in #148, I've also added a bit more tagged "motivations" per @vmx's suggestions; and also a segment attempting to make it very clear how these refinements to the rules are also of limited scope themselves, and are designed to work within rather than compromise or override the foundational limits.

### DIFF
--- a/FOUNDATIONS.md
+++ b/FOUNDATIONS.md
@@ -39,46 +39,75 @@ the format. Therefore, we should avoid introducing new formats unnecessarily as
 
 ## No Non-Local Reasoning
 
-An IPLD block should never be interpreted in the context of *anything* not
-contained in the block (and CID).
+Transforming content of a Block into Nodes conforming to the IPLD
+[Data Model](./data-model-layer/data-model.md) should never require
+interpretation in the context of anything not contained in the Block plus CID.
 
-For example, assuming we add support for relative links, the following
-definition of `foo` would not be a valid IPLD block:
+Similarly, traversing an IPLD Node according to a Path should not require
+interpretation in the context of anything not already contained in that Node plus Path.
 
-```
+**Motivation:** IPLD needs to be easy to reason about.  Systems which use
+nonlocal reasoning become harder to reason about, and much harder (if not
+impossible) to compose smoothly or with predictable results; therefore we
+should avoid this.
+
+**Negative Examples:**
+
+```javascript
+// This is an example of what is NOT possible.
 var foo = {
-  // points outside of the current block, into the parent's "baz" field.
-  "baz": {"/": "../../baz"}
+  "baz": Link("../../zot") // NOT legal: makes a non-local reference.
 }
 var bar = {
   "foo": CidOf(foo),
-  // `/foo/baz` points here.
-  "baz": "something"
+  "zot": "something" // `./foo/baz` imagines pointing here.
 }
 
-// resolution throug block `foo` depends on block `bar`.
+// resolution through block `foo` depends on block `bar`...
 Resolve("/ipld/${CidOf(bar)}/foo/baz/")
+
+// meaning this would be undefined, which is why relative links are NOT allowed:
+Resolve("/ipld/${CidOf(foo)}/baz/")
 ```
 
 For the same reason, IPLD links can't rely on an authority (e.g., a blockchain).
 
-Note: Links like this can still be encoded at the application level but they
-won't be handled by the IPLD resolver (and won't get the special "link" type).
+**Note:** Concepts that seem similar to relative linking can still be encoded
+at the application level.  This is fine, but distinct from "IPLD Links", because
+such linking won't be interpreted by IPLD path and link resolution (e.g. they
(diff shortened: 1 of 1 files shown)


FOUNDATIONS.md
```diff
@@ -39,46 +39,81 @@ the format. Therefore, we should avoid introducing new formats unnecessarily as
 
 ## No Non-Local Reasoning
 
-An IPLD block should never be interpreted in the context of *anything* not
-contained in the block (and CID).
+Transforming content of a Block into Nodes conforming to the IPLD
```
rvagg, line 42: I'd like to see clarification that the original authors of the text see the same thing here. The whole discussion about "context" makes me think that there could be some disagreement boiling beneath this one. I like the way you've made it more specific but it seems that it might be more specific than others may want.

rvagg (approved):
OK, so I like this but I wonder how much meaning this actually has now that it's been made specific. Other than the relative link example are there other obvious ways that someone might be tempted to break this?
I'd prefer to see an original author sign off on this.
Also, is it OK to define "Block" as existing above the format? Because that's another thing that's needed to interpret the data.

vmx (approved):

FOUNDATIONS.md
warpfork: +1 to would enjoy more eyes on this.  We've **got** to do something about the "anything" and broadness of the original, though, in some way or other.

"interpreted" is also a trigger word for me: data can be interpreted multiple ways simultaneously, and therefore that's a very tricky word to use here -- some sort of "interpreted for {purpose}" clause is necessary, imo.

warpfork: If we can add _more_ "doing X must not require interpretation with any more information than contained in Y+Z" entries, I'd see that as a very welcome improvement.

warpfork: I guess I can see how the way I've rephased this can be seen as being a blacklist, whereas "always" in computer science the better answer is to whitelist.  Unfortunately, I don't see any choice: due to the "data can be interpreted multiple ways" fact, I don't see any other way to go about this which doesn't result in trying to outlaw the sun shining.  (How shall I interpret an S-expr serialized in IPLD?  In the context of a Lisp Machine, of course.  But also as the text: homoiconicity!  *thumps SICP unironically for once*)

warpfork:
Had some brief exchange with @Stebalien about this a moment ago and (if I can recall/transcribe accurately):

- definitely on the same page that this is tricky to phrase :beers: 
- seems to be agreed that there are some places where we want this rule very strictly (basic parsing into Data Model definitely an example)
- seems to be agreed there are also places where this becomes more of a philosophic issue, and the defn of "local" starts working differently...
  - example was ~"imagine doing computation and that the function should only work with the information in its arguments -- no hazy other surrounding information" -- in this situation, "local" means "the arguments"... but that can be quite big, and (of particular note for this discussion) isn't directly correlated with blocks (it could certainly cross them).

(Hopefully that contains no wild misstatements or misquotes.)

So ISTM maybe one way to describe this is: the important thing in the generalized/philosophic form of this rule is: whatever is being used in some interpretation process must be finite and cannot refer to things "outside" -- but deeper "inside" is fine.  And in the case of our concrete examples about the interpretations we make core to getting things mapped into the IPLD Data Model, the "deeper inside" amendment also just isn't really necessary in practice, because in practice, we're doing a bunch of things that aren't going to do any kind of lookahead because that would be slow and silly, so it's just not worth mentioning.

Stebalien:
Basically, an IPLD operations should be stateless (or, at least, the state should be clearly specified?).

For example, the function that looks up paths in a specific block format should take a path and a block and then return a path. It should update any implicit information in some global context that then informs future operations.

warpfork:
So, this chunk as updated does take a heading that's fairly "principle"/"north star"-oriented, and includes some content under it that's at the fairly concrete end of the spectrum, and I'd very much like to find more wording in the future to do a bit more of the "principle" stuff.  In particular, I think we're floating some sort of meta-concept of 'scope' (?) which we seem to agree might not always bind tightly to "Block" (although in some cases, like parsing content to the Data Model level, it absolutely does!), and it'd be cool to expand on that.  This seems expansive and like it will take some work to phrase well, though, and deserves another future PR.

Meanwhile, I think this diff is moving in the right direction, and is double-approved, and clarifies how some of our important advancements relate to these principles -- so, merging.  (And will love to keep thinking about how we can further refine our text on this subject.  Onward!)

PR #161, opened 2019-08-12
move schemas overview to correct place

rvagg:
I suspect I borked this. I integrated the README contents and expanded it in schema-layer/schemas/README.md and linked to the overview.md there, but it's now sitting alone in its own directory. The README needs to be deleted and overview.md needs to live with the other content.


warpfork:
Closed instead by https://github.com/ipld/specs/pull/178 .  (I didn't check and find this first -- my bad :/ )

Issue #59, opened 2018-04-04
Document restrictions on keys of maps


vmx:
When you store data serialized as CBOR in IPLD there are restrictions on they keys that can be used in maps. This needs to be documented in the spec.

The restrictions are:

 - Keys must be strings
 - If they should be "pathable" (traversed with a path), they must not contain a slash

This information is based on the discussions at https://github.com/ipld/specs/issues/58.

PR #163, opened 2019-08-13
DagCBOR: Cleanup the spec

vmx:
The major change is that only strings are allowed as map keys,
without any further restrictions. Slashes are allowed within keys.
The issue of escaping those should be solved in a pathing layer
and not on the codec layer.

You better review the individual commits as the renaming of the file trips up diff.


vmx:
That's not the final version. But I'd like to take small steps in hope we can move faster this way.

block-layer/codecs/dag-cbor.md
```diff
@@ -0,0 +1,41 @@
...
+
+As with all IPLD formats, DagCBOR must be able to encode [Links].
+In DagCBOR, links are encoded using the raw-binary identity [Multibase] in a
+field with a byte-string type (major type 2), with the tag 42.
+
+(the inclusion of the Multibase exists for historical reasons)
+
+## Map Key Restriction
+
+In DagCBOR, map keys must be strings.
```
mikeal, line 25: Does CBOR even allow non-string keys?

I know we need to add this statement to the data model but I didn’t realize we needed to clear it up here as well.

Stebalien: Yes, it does. Also, It would be _really_ nice to support number keys (and I'm pretty sure it won't cause any problems with pathing).

vmx: We talked about keys in maps in the last sync meeting and thought about restricting the data model to string keys only (for now. Keeping things simple, we can revise that in the future). The issue to track this is https://github.com/ipld/specs/issues/59.

block-layer/codecs/dag-cbor.md
```diff
@@ -0,0 +1,41 @@
...
+## Format
+
+The CBOR IPLD format is called DagCBOR to disambiguate it from regular CBOR.
+Most CBOR objects are valid DagCBOR. The only hard restriction is that any field
+with the tag 42 must be a valid CID.
+
+## Link Format
+
+As with all IPLD formats, DagCBOR must be able to encode [Links].
+In DagCBOR, links are encoded using the raw-binary identity [Multibase] in a
```
rvagg, line 18: original:
>  DagCBOR, links are encoded using the raw-binary (identity, NUL) multibase in a field with ...

The `(identity, NUL)` bit is probably important to keep, or at least highlight in some way, as @warpfork found out [recently](https://github.com/ipld/go-ipld-prime/pull/23), although it's worded very unclearly. Adding a bracketed clarification might be enough, something like "(effectively prepending a `0` byte to the CID bytes)".

rvagg (changes requested):

block-layer/codecs/dag-cbor.md
vmx: My idea was that we are referring to "identity" in the multibase spec, so it should be clear. But I'm happy adding it back in.

vmx: @rvagg Would this be clearer:

In DagCBOR, links are [CIDs] encoded using the raw-binary identity [Multibase], i.e. the raw bytes prefixed with `0x00`. They are stored as byte-string type (major type 2), with the tag 42.

Stebalien (approved):

warpfork (approved):

PR #168, opened 2019-08-14
fix: specify dag-json serialization requirements

### LINKED ISSUE #108: DAG-JSON canonicalizing and whitespace
We need a spec about handling of whitespace in dag-json encodings.

I recently ran into a fun situation.  I have some code which emits json, in pretty print format.  This pretty printing includes a trailing line break, as is generally sensible for printing.  This is hashed and stored.

With the trailing linebreak.

When loading this, my deserializer consumed all the json tokens, then reported it's "done" with the json.  A hasher has been fed by a TeeReader, so, with exactly the same bytes that the deserializer consumed; and after the deserializer is done, the final hash is summed and then checked for integrity.

Which failed.  Because the trailing linebreak wasn't consumed by my deserializer.  So it wasn't rolled into the hasher.  So of course the hash wouldn't match.

Ouch.

Probably the answer is simple: "don't emit with whitespace".  But since I did, and now we're here...

What is our spec here?  https://github.com/ipld/specs/blob/master/Codecs/DAG-JSON.md doesn't currently have much to say.

Should we be Postel'y: accept more inputs (e.g. with whitespace variation) than we emit?  Perhaps also supporting a "--strict" mode (or, the contrary, "--lax" mode) behind feature flags of some kind?

Can we phrase this in terms of some clear higher-level rules, like "The content as known by the Data Model layer plus the multicodec value must fully specify all bytes", or, the contrary, "Byte sequences are acceptable as long as they result in one full JSON object, and may contain leading or trailing bytes so long as they do not yield any JSON token (e.g. are whitespace)"?

---

(n.b. we probably want to keep this separated from A) CBOR and B) canonical ordering of map entries; and C) anything else at all; canonicalization is hard enough one spec detail at a time.)

mikeal:
fixes #108

### DIFF
--- a/block-layer/codecs/DAG-JSON.md
+++ b/block-layer/codecs/DAG-JSON.md
@@ -6,6 +6,17 @@ DAG-JSON supports the full [IPLD Data Model](../data-model-layer/data-model.md).
 
 ## Format
 
+### Serialization
+
+While it is unlikely de-serializers will enforce the following rules, codec implementors
+**SHOULD** do the following in order to ensure hashes consistently match for the same block data.
+
+* utf8 sort object keys
+* strip whitespace
+
+This produces the most compact and consistent representation which will ensure that two codecs
+producing the same data end up with matching block hashes.
+
 ### Simple Types
 
 All simple types except binary are supported natively by JSON.
@@ -20,7 +31,7 @@ of `dag-json` can't use the native JSON parser and serializer.
 {"/": { "base64": String }}
 ```
 
-### Link Type
+#### Link Type
 
 ```javascript
 {"/": String /* base encoded CID */}
--- a/block-layer/codecs/DAG-JSON.md
+++ b/block-layer/codecs/DAG-JSON.md
@@ -8,8 +8,7 @@ DAG-JSON supports the full [IPLD Data Model](../data-model-layer/data-model.md).
 
 ### Serialization
 
-While it is unlikely de-serializers will enforce the following rules, codec implementors
-**SHOULD** do the following in order to ensure hashes consistently match for the same block data.
+Codec implementors **MUST** do the following in order to ensure hashes consistently match for the same block data.
 
  - Sort object keys by their (UTF-8) encoded representation, i.e. with byte comparisons
  - Strip whitespace
--- a/block-layer/codecs/DAG-JSON.md
+++ b/block-layer/codecs/DAG-JSON.md
@@ -11,8 +11,8 @@ DAG-JSON supports the full [IPLD Data Model](../data-model-layer/data-model.md).
 While it is unlikely de-serializers will enforce the following rules, codec implementors
 **SHOULD** do the following in order to ensure hashes consistently match for the same block data.
 
-* utf8 sort object keys
-* strip whitespace
+ - Sort object keys by their (UTF-8) encoded representation, i.e. with byte comparisons
+ - Strip whitespace
 
 This produces the most compact and consistent representation which will ensure that two codecs
 producing the same data end up with matching block hashes.


block-layer/codecs/DAG-JSON.md
```diff
@@ -6,6 +6,17 @@ DAG-JSON supports the full [IPLD Data Model](../data-model-layer/data-model.md).
 
 ## Format
 
+### Serialization
+
+While it is unlikely de-serializers will enforce the following rules, codec implementors
+**SHOULD** do the following in order to ensure hashes consistently match for the same block data.
+
+* utf8 sort object keys
```
vmx: "utf8 sort" isn't really specific enough. Do you mean the Unicode Collation with the DUCET table? With the roor collation? Or sorting by code points?

I think we should go for something like http://gibson042.github.io/canonicaljson-spec/. Where those things are clearly specified (though also there it took me a while to understand how they actually sort).

vmx (changes requested):

block-layer/codecs/DAG-JSON.md
Stebalien: I think we just sort lexicographically by encoded utf8. Note: we don't _currently_ bother normalizing anything first.

We need to handle that (for _all_ paths in _all_ formats) but that's a can of worms we've repeatedly opened and then re-shut).

vmx: > I think we just sort lexicographically by encoded utf8.

What does lexicographically mean? Which locale?

Stebalien: It's sorted using the _encoded_ (utf8) representation. That is, we just sort the bytes. See https://tools.ietf.org/html/rfc7049#section-3.9

vmx: I see. Could we then change this to "sort UTF-8 encoded objects keys bytewise" or something similar?

mikeal: Let’s use the language [from wikipedia](https://en.m.wikipedia.org/wiki/UTF-8): “Sorting order: The chosen values of the leading bytes means that a list of UTF-8 strings can be sorted in code point order by sorting the corresponding byte sequences.”

rvagg (approved):

block-layer/codecs/DAG-JSON.md
vmx: @mikeal sounds good to me. As long as there's a bit more information what "utf8 sort" is supposed to mean.

warpfork: I both like the language from wikipedia, but also support a flat, clear-speak statement like 

> That is, we just sort the bytes

... because it's true.  And character encodings are terrifying enough by their nature that clearspeak about the subject should be highly prized.

mikeal: “Sort by bytes” is going to be rather confusing to many developers since most of them don’t know that this is how utf-8 strings sort already, which may be what the keys are already encoded into. We don’t want to imply that it’s necessary to convert the string to binary in order to sort properly. We also need to worry about languages that default to ascii encoded strings (Python 2.x), because those won’t sort properly without being converted to utf-8.

I’m normally in favor of rather terse statements but I’m beginning to think this needs a bit more text than I’ve been giving it so far.

vmx: > “Sort by bytes” is going to be rather confusing to many developers

I find that really clear. Look at the UTF-8 string byte by byte and compare it.

>  most of them don’t know that this is how utf-8 strings sort already

I don't understand what you mean by this. Perhaps I've been to deep into Unicode ordering/collation. I have no clue how utf-8 strings sort. I would need instructions, e.g. sort them by byte order. Or do proper collation with a locale. I don't see how there is a "default sort order".

mikeal: > I don't understand what you mean by this. Perhaps I've been to deep into Unicode ordering/collation. I have no clue how utf-8 strings sort. I would need instructions, e.g. sort them by byte order. Or do proper collation with a locale. I don't see how there is a "default sort order".

You don’t need to because most languages and/or standard libraries have sort functions that do this for you :) If the keys are already utf8, this is what people should use and they should **not** convert them to a binary object and do the sorting themselves ;)

vmx:
We had a long discussion about how to formulate the sort-order. I went ahead rebased and pushed a new formulation I hope everyone can live with, please review again @warpfork @mikeal @Stebalien.

block-layer/codecs/DAG-JSON.md
```diff
@@ -6,6 +6,17 @@ DAG-JSON supports the full [IPLD Data Model](../data-model-layer/data-model.md).
 
 ## Format
 
+### Serialization
+
+While it is unlikely de-serializers will enforce the following rules, codec implementors
```
Stebalien: If this is a SHOULD requirement, de-serializers MUST NOT enforce this rule.

vmx: What about removing this part? Why not talking only about Serializers in this section? So it would be "Codec implementors **SHOULD**…".

mikeal: I think we should phrase it as “de-serializers SHOULD preserve the encoded ordering, encoders MUST ensure keys are ordered using ....”

Specifying any sort ordering that doesn’t conform to the encoded ordering is problematic if you want to support streaming decoders.

Any MUST regarding map ordering during de-serialization is problematic across languages at this layer, because native maps in different languages do different things to the sort ordering and we don’t want to prohibit native maps from being used in IPLD. In Selectors we can use MUST language to describe the ordering because we don’t necessarily have to use native maps when de-serializing and deterministic order is a requirement for Selectors.

Taken as a whole, these constraints leave the only good option to require deterministic ordering at encode time and attempt to preserve it during decode.

rvagg: > I think we should phrase it as “de-serializers SHOULD preserve the encoded ordering, encoders MUST ensure keys are ordered using ....”

yes, but, I don't know if it's useful to get into the layer above the codec here, those are weeds that maybe we can't afford to bring into this spec. It's the second part of that statement that I think should go in. Discussing what map-ordering occurs once you're out of the serialized form is a question for the layers above this one and we should deal with that there rather than discussing it in all of these codec docs.

vmx:
New version. I agree with @rvagg and don't want to get into the business of de-serializers. I also changed to `MUST`. Please review again.

Stebalien (approved):
Note: we should also specify a regular (non-dag) JSON that doesn't have these requirements.

block-layer/codecs/DAG-JSON.md
```diff
@@ -20,8 +30,8 @@ of `dag-json` can't use the native JSON parser and serializer.
 {"/": { "base64": String }}
 ```
 
-### Link Type
+#### Link Type
 
 ```javascript
 {"/": String /* base encoded CID */}
```
Stebalien, line 36: If we want determinism, what's the required base encoding?

vmx: That's my next PR, I've already written it, I'll push it once we've merged this one :)

mikeal:
I can’t official review the PR since I created it orginally, but I’m +1 on merging.

mikeal:
> Note: we should also specify a regular (non-dag) JSON that doesn't have these requirements.

Can’t we just consider that entry the multiformat table to be defined by the JSON RFC and not have a specific IPLD representation?

Stebalien:
> Can’t we just consider that entry the multiformat table to be defined by the JSON RFC and not have a specific IPLD representation?

Well, yes. It would be defined by the JSON RFC. My point is that we should consider treating "json" as a valid IPLD codec as well.

mikeal:
> My point is that we should consider treating "json" as a valid IPLD codec as well.

yup, part of the refactoring i’m doing as part of `js-multiformats` will provide this in the basic set since it’s effectively free (can be implement in about 3 lines of code with no deps).

PR #170, opened 2019-08-20
HashMap: don't serialize `bitWidth`, infer it from `map` length

rvagg:
An outstanding item from https://github.com/ipld/specs/pull/131, I've implemented this, in JavaScript at least, and it's 👌.

Note that the minimum `bitWidth` of `3` is now a "must" not a "should" as less than that would give us a `map` length of less than 1-byte, breaking our ability to infer `bitWidth` from `map` length.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,9 @@ The Schema Layer will also serve as an enabling layer for complex multi-block da
 
 |     |      |
 |-----|------|
-| [Specification: IPLD Schemas](schemas/README.md) | [schemas/README.md](schemas/README.md) |
 | [Concept: IPLD Multi-block Collections](schema-layer/data-structures/multiblock-collections.md) | [schema-layer/data-structures/multiblock-collections.md](schema-layer/data-structures/multiblock-collections.md) |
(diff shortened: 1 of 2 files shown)


rvagg:
pushed a couple of minor, unrelated, fixes:
* add a note about the linked vs inline child node in the schema to explain the weird kinded `Element` union
* removed a TODO near that top that wanted to figure out whether this was an "advanced layout", "generic" or something else. I fixed it to "advanced data layout" for now, we can change later if this terminology evolves.

rvagg:
Synced my abstract JS implementation with this version of the spec @ https://github.com/rvagg/iamap/pull/7, it's complete except that it doesn't support inline child nodes in any way. That'll need to come at some point, for UnixFSv2 I guess.

I'm also iffy about the kinded union, differentiating between kind=array (bucket), kind=link (link to child node) and kind=map (inlined child node). It works, it just doesn't _feel_ right, like it's _too_ compact. But it is that, we're saving the additional `{link:CID}` in this version, it's just `CID` now.

warpfork:
> it's *too* compact

haha

This seems like an acceptable problem :)

rvagg:
_inferring_ implicit support for this

PR #171, opened 2019-08-20
Add exploration report about merkle segments

### LINKED ISSUE #67: Spec refining: specify valid merklepath segments and encoding

vmx:
This report was originally issue #67.

Closes #67.

### DIFF
--- a/design/history/exploration-reports/2018.07-merkle-segments.md
+++ b/design/history/exploration-reports/2018.07-merkle-segments.md
@@ -1,0 +1,66 @@
+This document was archived from https://github.com/ipld/specs/issues/67. There was agreement that UTF-8 NFC should be the canonical encoding. Below is the original text of the issue written by @warpfork.
+
+---
+
+## Mission
+
+It's important to specify precisely what is a valid merklepath segement in IPLD.
+
+The spec currently contains a "TODO: list path resolving restrictions" and this could be improved :)
+
+---
+
+## Why
+
+First, a quick clarification: "merklepath segments" are a distinct concept from "IPLD Selectors".  Merklepaths are a specific and limited implementation of IPLD Selectors; they can only specify a traversal to a single object; and importantly, we want them to be serializable in a way that's easy for humans to operate.  To quote the current spec for motivations:
+
+> IPLD paths MUST layer cleanly over UNIX and The Web (use `/`, have deterministic transforms for ASCII systems).
+
+(Perhaps "ASCII" is a little over-constrained there.  The spec *also* says "
+IPLD paths MUST be universal and avoid oppressing non-english societies (e.g. use UTF-8, not ASCII)" -- we might want to refine those two lines after we tackle the rest of this issue.)
+
+Second of all, just a *list* of other issues that are fairly closely related to a need for clarity on this subject:
+
+ - https://github.com/ipfs/go-ipfs/issues/1710 -- "**IPFS permits undesirable paths**" -- lots of good discussion on encodings, and examples of valid but problematic characters
+- https://github.com/ipld/specs/issues/59 -- "**Document restrictions on keys of maps**"
+- https://github.com/ipld/specs/issues/58 -- "**Non-string keys**" -- this one has a particularly interesting detail quoted: "The original intention was to actually be quite restrictive: map keys must be unicode strings with no slashes. However, we've loosened that so that they can contain slashes, it's just that those keys can't be pathed through.".  (n.b. *this* issue here is named "merklepath segments", not "IPLD keys", specifically to note this same distinction.)
+- https://github.com/ipld/specs/issues/55 -- "**Spec out DagPB path resolution**"
+- https://github.com/ipld/specs/issues/37 -- "**Spec refining: make sure that an attribute cannot be named `.`**"
+- https://github.com/ipld/specs/issues/20 -- "**Spec refining: Merkle-paths to an array's index**"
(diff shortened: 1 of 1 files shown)


warpfork:
... There's a lot of positions in this that I'd probably update now, having had more time to slow-cook on the ideas.

Roughly:

- the idea that there's a distinction between IPLD keys and path segments is probably *a bug*

- I think the number one thing to get straight that makes everything else fall in line is to nail a spec for paths which starts with the cornerstone "Paths are composed of PathSegments" and then define validity around PathSegment.  Then we can nudge about the serialization and escaping of paths and their segments for various encodings as a layer atop that, and things will be much clearer.  This idea is missing from this snapshot.

- As much as I still think "UTF-8, NFC" is the defacto correct choice for character encodings when a character encoding must be chosen,  I increasingly see the honor and simplicity of unixy "strings are really just bytes" approach.  Normalization causes a loss of generality, full stop.

... *but*, that's the point of putting this in a "history" file with a timestamp right upfront in the filename.  These are dated opinions that we're recording, not forcing to be up-to-date nor ratified right now.  We can write further snapshots of updated thinking in the future.  I'll merge this snapshot now.

PR #172, opened 2019-08-20
Add exploration report about type convention

### LINKED ISSUE #71: IPLD Type Convention
Motivation 1: I'd like to be able to look at an IPLD object and know, approximately, it's intended interpretation (without guessing or using context).

Motivation 2: I'd like to be able to define or extend a type system for *my* IPLD application without having it completely fail to interop with other IPLD type systems.

Motivation 3: I'd like to buy some time to figure out the perfect type system.

---

We've been discussing IPLD type *systems* but these discussions usually boil down to implementing the perfect system. I'd like to propose an alternative: the IPLD type convention.

Proposal: `@type: $something` denotes a type. What this type *means* depends on the type's type (if specified).

Why @? It's less likely to conflict but I'm not fixated on it.

Why "the IPLD type convention"? This isn't a specification. Basically, I'm giving in to JSON duck-typing and calling it "good enough".

Why is it good enough? This is a decentralized system so we'll have to check everything *anyways*. Trying to prescribe structure on users tends to lead to more trouble than it's worth (IMO). If we need more structure, we can always give the type a type to make sure we're operating within the correct type system.

How will this work with existing formats:

1. CBOR/JSON: Do nothing. For existing objects without a `@type`, these objects simply don't have types (within this system). Type systems that need to give everything *some* type can just give these some 
2. Git (tree, commit, etc), Eth, etc: I'd *like* to retroactively add in a type field. Thoughts? I kind of doubt this will break anything.

---

We've *also* talked about adding a new format with the structure `<CidOf(type)><data>`. That is, introduce a new format where we put all the type and schema information in a separate object, prepending the CID of this separate object to the *actual* object (the value).

After considering this for a bit, I've realized we should treat these as separate concerns: we're conflating *types* with *schemas*. There's no reason we can't introduce this new, compressed format at some later date even if we go with the above "type convention" proposal.

---

Disclaimer: this was *not* my idea, I've just finally convinced myself that it's probably "good enough".

Thoughts @jonnycrunch (you're the one who told me to look into the JSON-LD spec), @diasdavid, @davidad, @whyrusleeping?

---

While I'd like to avoid prescribing *too* much, I'd like to define a set of conventions that users *should* follow. For example:

* `@type: CID` -  CID points to the actual type.
* `@type: {}`: inline type. This will often be used for type "constructors". For example: `{@type: {@type: "generic", constructor: CidOfConstructor, parameters: [...]}`.
* `@type: "string"`: A human readable string/path. IMO, this should *usually* be used to specify the type *system*.
* `@type: 1234`: A multicodec. A reasonable type-of function would look this multicodec up in the multicodec table to map it to a CID.
* `@type: [thing1, thing2, thing3]`: multiple types.

vmx:
This report was originally issue #71.

Closes #71.

### DIFF
--- a/design/history/exploration-reports/2018.09-type-convention.md
+++ b/design/history/exploration-reports/2018.09-type-convention.md
@@ -1,0 +1,165 @@
+This document was archived from https://github.com/ipld/specs/issues/71.
+
+#71: IPLD Type Convention
+-------------------------
+Opened 2018-09-20T22:57:15Z by Stebalien
+
+Motivation 1: I'd like to be able to look at an IPLD object and know, approximately, it's intended interpretation (without guessing or using context).
+
+Motivation 2: I'd like to be able to define or extend a type system for *my* IPLD application without having it completely fail to interop with other IPLD type systems.
+
+Motivation 3: I'd like to buy some time to figure out the perfect type system.
+
+---
+
+We've been discussing IPLD type *systems* but these discussions usually boil down to implementing the perfect system. I'd like to propose an alternative: the IPLD type convention.
+
+Proposal: `@type: $something` denotes a type. What this type *means* depends on the type's type (if specified).
+
+Why @? It's less likely to conflict but I'm not fixated on it.
+
+Why "the IPLD type convention"? This isn't a specification. Basically, I'm giving in to JSON duck-typing and calling it "good enough".
+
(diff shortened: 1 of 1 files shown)


Stebalien:
Can't we just move this to an IPLD/notes? This effectively closes the conversation.

If we're _really_ going to do this (which I think is a colossal waste of time and effort and no better than just having a notes repo full of dead issues), go ahead.

PR #176, opened 2019-08-20
Archive issue as exploration report about Graphsync optimizations

### LINKED ISSUE #120: GraphSync: Data Deduplication Across Simultaneous Queries
## Overview
There are a few important cases where it would be very useful to have multiple queries that access similar data be bunched together. These can be handled at the either at the Selectors layer or at the GraphSync layer. Given both my interest in data deduplication, as opposed to query traversal, and the concerns with CID selectors raised in #116 I propose handling this at the GraphSync layer.

## Useful Examples
To understand the utility of handling a query of these simultaneous queries let's look at two examples:

### Resuming Interrupted Queries
We ask for the full graph starting at Root:
```
       Root
   /    |    \ 
  A0   B0   C0
  |     |     |
  A1   ...   ...
  |
  A2 
```

We then get interrupted in our DFS resolution of the graph and only receive nodes `Root, A0, A1`. If we want to resume our session, whether with the same peer or a different one, we would like to not get repeats of the nodes we already have. There are a variety of ways to describe how to resume our sync, they are of the form: I want the graph rooted at `Root` and I
1. Have first 3 nodes of the query
2. Have nodes `Root, A0, A1` (or some condensed form like `A1` and all parents)
3. Want to start with `A2, B0, C0`

While option 1 is the most efficient it requires both peers doing the traversals in the same order which we might not be prepared to guarantee - something I think has already been looked into and that I'm probably just lacking context for. Options 2 and 3 are similar as they are essentially inverses.

**Option 3** would be greatly helped by allowing data deduplication across queries/queries being grouped together.

### Versioned Data (Git, Operational Transform, CRDTs, ....)
We start with a graph `G` at state `G0`, and then created two modified versions of the graph `G1` and `G2` both of which reference the previous state, `G0`. We then give a friend the CIDs to the latest versions of `G` (which as far as we know are `G1` and `G2`). The friend then uses GraphSync to ask us (or some other peer) for the full graphs under `G1` and `G2`. We would like to not get the data under `G0` more than once.

```
G1      G2
  \    /
    G0
``` 

## How we might handle this
Fundamentally what's being asked for is that GraphSync retain the list of sent blocks per peer longer. The balancing act required in the solution is how we support this while not allowing a single peer to eat up all of a GraphSync responder's resources.  

### Send Bundled Requests
Allow sending multiple requests together such that GraphSync knows not to GC its list of sent blocks per peer. The protocol already allows for sending multiple requests together so all we could easily support this by just A) holding off on the GC until the full multi-request is complete or B) add a flag saying that for this particular multi-request we should hold off the GC until the multi-request is complete.

### Allow Append Requests
All subrequests are still processed individually, but the requests can have a field tying those requests to other on-going requests for GC purposes.

vmx:
This report was originally issue #120.

Closes #120.

### DIFF
--- a/design/history/exploration-reports/2019.05-graphsync-deduplication-across-queries.md
+++ b/design/history/exploration-reports/2019.05-graphsync-deduplication-across-queries.md
@@ -1,0 +1,123 @@
+This document was archived from https://github.com/ipld/specs/issues/120.
+
+#120: GraphSync: Data Deduplication Across Simultaneous Queries
+---------------------------------------------------------------
+Opened 2019-05-03T19:02:15Z by aschmahmann
+
+## Overview
+There are a few important cases where it would be very useful to have multiple queries that access similar data be bunched together. These can be handled at the either at the Selectors layer or at the GraphSync layer. Given both my interest in data deduplication, as opposed to query traversal, and the concerns with CID selectors raised in #116 I propose handling this at the GraphSync layer.
+
+## Useful Examples
+To understand the utility of handling a query of these simultaneous queries let's look at two examples:
+
+### Resuming Interrupted Queries
+We ask for the full graph starting at Root:
+```
+Root
+/    |    \
+A0   B0   C0
+|     |     |
+A1   ...   ...
+|
+A2
+```
+
+We then get interrupted in our DFS resolution of the graph and only receive nodes `Root, A0, A1`. If we want to resume our session, whether with the same peer or a different one, we would like to not get repeats of the nodes we already have. There are a variety of ways to describe how to resume our sync, they are of the form: I want the graph rooted at `Root` and I
+1. Have first 3 nodes of the query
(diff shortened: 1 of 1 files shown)


vmx:
@aschmahmann Please approve this archival of your exploration report.

aschmahmann:
@vmx Whatever you think makes sense. That issue is referenced by ipfs/go-graphsync#31, and still has some low hanging fruit to be implemented.

I'm happy to have future discussion (when the time and priority arise) in #120, a new IPLD issue, or in go-graphsync.

vmx:
@aschmahmann In the [a sync sync meeting](https://github.com/ipld/team-mgmt/blob/3b434079fe2b723a6fdc592605a723743377a0f2/meeting-notes/2019/2019-08-12--ipld-sync.md#spec-repo-pruning) we talked about what we do about issues like that. And decided to archive them as they contain valuable information. In case we want to tackle those again, or parts of it, new issues should be opened which will then probably be more focused.

PR #177, opened 2019-08-20
Archive issue as exploration report about advanced layouts design

### LINKED ISSUE #130: Identifying the critical decisions in IPLD advanced data structure design

vmx:
This report was originally issue #130.

Closes #130.

### DIFF
--- a/design/history/exploration-reports/2019.06-advanced-layouts-critical-design-decisions.md
+++ b/design/history/exploration-reports/2019.06-advanced-layouts-critical-design-decisions.md
@@ -21,7 +21,7 @@ engagement at all, rather than simply having another regular map, list, or
 byte sequence.
 
 Note that "signaling" refers *only* to the conditional branch that decides
-we're going to do *something* -- it doesn't define what the something *is;
+we're going to do *something* -- it doesn't define what the something *is*;
 that's the "referencing" decision.
 
 (It's useful to regard signaling as a distinct issue because we should be
--- a/design/history/exploration-reports/2019.06-advanced-layouts-critical-design-decisions.md
+++ b/design/history/exploration-reports/2019.06-advanced-layouts-critical-design-decisions.md
@@ -1,0 +1,258 @@
+This document was archived from https://github.com/ipld/specs/issues/130.
+
+#130: Identifying the critical decisions in IPLD advanced data structure design
+-------------------------------------------------------------------------------
+Opened 2019-06-11T19:10:13Z by warpfork
+
+(Forward: this is a collection of a *lot* of stuff, and I'm trying to be comprehensive, but will inevitably fail.  If you want to read this and visualize semi-continuous injections of "TODO: expand" and "TODO: more alternatives?", that will probably be appropriate!)
+
+---
+
+There are approximately three distinct and critical decisions we need to regard:
+
+1. Signaling
+2. Referencing
+3. Having the implementation
+
+### Signaling
+
+"Signaling" refers to deciding / detecting that we need any advanced layout
+engagement at all, rather than simply having another regular map, list, or
+byte sequence.
+
+Note that "signaling" refers *only* to the conditional branch that decides
+we're going to do *something* -- it doesn't define what the something *is;
+that's the "referencing" decision.
+
+(It's useful to regard signaling as a distinct issue because we should be
+able to point on a flow control map to the exact position where signaling
+occurs -- it's _one point_.)
+
+### Referencing
+
+(alternate titles: "rendezvous", "implementation selection")
+
+"Referencing" refers to the need to pick a particular implementation of code
+we're about to use to understand the data layout.
+
+Roughly, "referencing" is about how something says "I'm a hamt!" or
+"i'm a rabin chunked byte sequence!".
+
+Note that referencing is about *which* of those things a data structure is;
+the "signaling" decision already handled whether or not *something* special
+is about to happen.  (Referencing can involve more complex decision trees,
+whereas signaling is a pure boolean.)
+
+### Having the implementation
+
+"Having the implementation" refers to the need to have actual executable code
+which understands and manipulates the data layout.  This might refer to a
+plugin registration system, or a bytecode interpreting system, or... etc.
+
+### Other
+
+Some advanced layout algorithms have additional parameters than merely their reference;
+for example, HAMTs can have a 'bitwidth' parameter.
+
+These additional parameters may be worthy of special consideration, because
+they have the interesting property of being variable without needing *new code*
+in the client libraries acting on the data.  (This also means that parameters
+as recorded in serial data can vary in ways that don't affect whether or not
+we "have the implementation", which may be useful; however, by the other side
+of the same coin, it poses further interesting questions.)
+
+We could also choose to remove these from the field of play entirely and
+simply state that the must be part of the Referencing data overall and
+opaque to the rest of the system.
+
+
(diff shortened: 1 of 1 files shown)


rvagg (approved):

warpfork (approved):
Okay... I was thinking there were some more details other alternatives I might like to inject into another generation of writeup on this... but I seem to have misplaced any notes I had on that.  So perhaps at this point I'll just thumbs up that we merge this as is as a snapshot that's nicely linkable; and if another generation of writeup on this is prompted by future questions, it'll get written when those questions are more pressing.

design/history/exploration-reports/2019.06-advanced-layouts-critical-design-decisions.md
```diff
@@ -0,0 +1,258 @@
...
+3. Having the implementation
+
+### Signaling
+
+"Signaling" refers to deciding / detecting that we need any advanced layout
+engagement at all, rather than simply having another regular map, list, or
+byte sequence.
+
+Note that "signaling" refers *only* to the conditional branch that decides
+we're going to do *something* -- it doesn't define what the something *is;
```
warpfork:
```suggestion
we're going to do *something* -- it doesn't define what the something *is*;
```

missing end of italics

Issue #164, opened 2019-08-13
Check PRs for broken links


vmx:
I just found a wonderful tool called [remark-validate-links](https://github.com/remarkjs/remark-validate-links) which checks for local links. I plan to add a TravisCI job to run that on this repo to make sure we don't introduce any breakage with new PRs.

Currently quite a few links are broken. I will fix those before enabling the CI job.

warpfork:
I also just tried https://github.com/raviqqe/liche and it did the trick.  They look pretty similar?

Here's a list of features I checked for:

- works on markdown files (thus locally)
- checks for `#headings`
- can check absolute links against some root

Looks like both `liche` and `remark-validate-links` do all of these.

warpfork:
`remark-validate-links` also looks like it's doing a lot, uh, more?  According to its readme, "git remote and git rev-parse also runs for processed files", and I cannot imagine what the purpose of this is.

I'm just starting to use `liche`, but I've previously used the related https://github.com/raviqqe/muffet, and would note that it's been extremely fast and no-nonsense.

vmx:
As I'm a fan of tooling and automating things, I'll take this issue and when working on it, I can just try both and see which one does a better job.

PR #178, opened 2019-08-22
Fixing 404's and links around schema docs.

warpfork:
All schema content in one place now.  Not sure sure how it got split.

Went with the shorter of the two dir structures; also in part because we've had some conversations lately about the "layercake" model having some tendencies to confuse, so, figuring that as long as I'm going to do the footwork on work on paths and links anyway, eliding that word seems like the right move.

(N.b. I know this is another single patch to organization, and I'd rather do draft-whole-org-first approach than spattered patching... but @whyrusleeping is asking for where docs outside of our own should link to for this stuff, and I feel obliged to answer with something that isn't 80% broken links.)

I'm using a link checker on this stuff now and in the future.  I've used https://github.com/raviqqe/muffet before to great success on web content; now, https://github.com/raviqqe/liche seems to be a comparable tool that works directly on markdown, which makes it much easier to use *before* publishing, so I'm going to start using it.  Heavily.  (This PR doesn't fix every broken link, because I want to go to bed, but I'd like to do that soon, and then maybe put a link checker in CI.)


rvagg (approved):
closes https://github.com/ipld/specs/pull/161
as long as this pulls the newer schema-layer/schemas/README.md to replace the older schemas/README.md as #161 was trying to do before pulling it back the other way.

vmx:
@warpfork in regards to link checking tools, see: https://github.com/ipld/specs/issues/164

concepts/cardinality.md
```diff
@@ -13,5 +13,5 @@ Cardinality concepts are also useful for clear documentation and understanding s
 modifier keywords.
 
 You can see some examples of cardinality-counting used to describe `nullable` vs `optional` vs `defaults`
-in the schema docs: [Cardinality Examples](../schema/schema-kinds.md#cardinality-examples).
+in the schema docs: [Cardinality Examples](/schemas/schema-kinds.md#cardinality-examples).
```
vmx: I'm not sure if absolute paths are a good idea. I'm a huge fan of links that work independent of where the files are stored.

vmx (changes requested):

concepts/cardinality.md
warpfork: So I actually raised an eyebrow at this too, because I first saw it in some of Rod's docs, and I thought "noooo, surely that doesn't work?!".

And so I poked it and it turns out it works better than I thought.  Github actually transforms it into correct links using the "tree/master" prefix, or etc, as contextually appropriate.

Does this change your opinion?  (I could go either way, since I can't make an error-prone-ness argument one way or the other given we're going to add link checkers to CI soon.)

warpfork:
Yeah, I teased apart the readme files manually and stitched together the superset of content.

concepts/cardinality.md
vmx: I'd feel more comfortable having relative links (perhaps I've been bitten too much by all those static website generators and JS build tools that expect the output to be stored at a domain root directory which totally breaks things if you deploy it on example.com/tmp/).

warpfork: I feel ya.  Will change.

vmx (approved):

PR #179, opened 2019-08-23
Readmes covering intent of design and history directories

warpfork:
A lot of this has been discussed in various channels and at various levels of meta, but not gathered into a single document before, so, I've attempted it.  And it seems like covering what is and is not imagined to go under the umbrella of these directories is going to be a frequently asked question, so... I wrote most of it as an FAQ.

I tried to write this in the most forward-progress-focused way possible.  A lot of it is talking about setting ourselves up with the right incentives, and trying to outmaneuver potential bad habits in advance, which makes it... tricky to phrase clearly.  I've done my best; edits welcome.


design/history/README.md
```diff
@@ -0,0 +1,248 @@
...
+are not themselves suitable to being treated as artifacts for reference.
+
+There are several reasons we believe a directory of files to be superior to
+github issues for our purposes:
+
+- It's preferable to save github issues for real issues that need active engagement;
+- it's possible to link to specific headings in markdown in files in github,
+  while it is *not* possible to do this to headings within in issue bodies;
+- and it's important and desirable that conversations reach halting states;
+  we want to emphasize that "thread necros" are *bad*, and rebooting a
```
warpfork: This in particular could use better statement.  If you've heard the phrase 'thread necro" before, it unpacks instantly; if you haven't, it's *extremely* inscrutable.

rvagg:
Sorry to do this, but it's a bit too wordy for a README as it is. Could you boil it a little longer to extract the concentrated essence of it?

warpfork:
... Yeah, you're right, this needs some work.  I think I'll try to tease it apart into a few more files -- a "but-why.md" for most of the long stuff, and then the readme can link more tersely to headings inside that.

The one thing I'm sure of is that we need something to link to which states basics like

- Yes, this isn't github issues
- *Yes, we thought about that*
- Truly, there's a separate purpose and flow in mind
- No, not everything that used to be a github issue goes here
- Yes, avoiding topic prefixed filenames is intentful

... and while organizing it for discoverability might be tricky, the questions are getting asked, so they need answers; and unfortunately most of those don't lend well to terse or flippant answers.

design/history/README.md
vmx: I'd really like this one to be rephrased. Perhaps:

> we want to emphasize that reviving discussions that haven't seen any update for a long time are *bad*

design/history/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+If we want the "open" and "closed" states to have any meaning, it's critical
+that things move to the closed state.  Attempting to do design processes
+entirely in Github issues doesn't necessarily obstruct this -- but
+observationally, in practice, *it does*:
+design discussions in issues almost invariably receive protest when closed,
+even if the issue has been open (sometimes for *years*!) without comment.
+
+Yet closing issues is necessary to prevent the issue list from becoming
+a graveyard full of ghastly distracting vuvuzelas.  What to do?
```
vmx: I kind of like the "graveyard full of ghastly distracting vuvuzelas", but I think we should stick to simpler phrasing. Perhaps simply:

> …from becoming a distraction.

vmx (changes requested):
I agree that making it shorter/denser makes sense.

Though I commented on two things I'd like to see changed (I'm just making propositions, other phrasing is welcome).

warpfork:
Some rearranging.  Vuvuzelas are gone.  Necromantic references also gone.  Readme now shorter due to moving most content to FAQ, and though I'm not sure the crosslinking is dense enough, it seems better now.

vmx (approved):

PR #183, opened 2019-08-29
schema: advanced layout parameterization

rvagg:
This is a PR against #182, you should review that one first. This one adds the idea that advanced layouts may have parameters, perhaps these things are exposed in their codegen APIs, or perhaps you use them within schemas to lock them down. Like in the Filecoin case of wanting to dictate that a HAMT will always have the same parameters forevermore, there is only one HAMT shape acceptable. New HAMTs will use those parameters, existing ones are expected to have those parameters or they are gibberish.

### DIFF
--- a/schemas/advanced-layouts.md
+++ b/schemas/advanced-layouts.md
@@ -1,0 +1,120 @@
+Advanced Layouts for IPLD Schemas
+---------------------------------
+
+An "advanced layout" is a special type in an IPLD Schema that denotes an abstract node that may _act_ as one of the standard IPLD [schema kinds](./schema-kinds.md) when interacted with above the schema layer, but be represented in entirely different data forms at the data model layer. Advanced layouts allow us to use transformational logic to present consistent interfaces to significantly more complex data than the data model otherwise allows or significantly larger data than reasonable block sizes allow.
+
+Data structures such as maps, lists or byte arrays may be implemented in novel forms with the data model and may even transparently cross block boundaries to present data that is potentially both larger and more complex than a standard data model kind while affording the standard data model API interaction modes of their respective kinds.
+
+Example uses of advanced layouts include:
+
+* Very large byte arrays that may cross many block boundaries but still present a standard `bytes` kind interface.
+* Maps that can be sharded across many blocks allowing efficient access via standard `map` kind interfaces while allowing extremely large storage.
+* Encrypted block mechanisms that are able to encode and decode data at the data model layer (perhaps in `byte` or `string` form) while presenting as unencrypted kinds when accessed through the schema layer.
+* Novel kind representation formats, such as high-precision or large numbers or alternative string encoding formats. Although size constraints at the language layer for number kinds may represent challenges for large or high-precision number formats.
+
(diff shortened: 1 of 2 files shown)


schemas/advanced-layouts.md
```diff
@@ -118,3 +118,28 @@ If we were to take an IPLD node, and assert that it is of type `BigConfidentialC
...
 If `ROT13` and `HashMap` were to omit their `root` descriptor, we could only make assertions 1 and 2 above.
+
+## Representation parameterization
+
+The IPLD HashMap spec defines some possible parameters that may be used at instantiation time. These parameters are stored in the root block so the parameters used to write the `HashMap` are known when reading the `HashMap`. Such parameters are generally provided programatically, but we may use our advanced layout definition to dictate strict requirements.
+
+```ipldsch
+advanced HashMap {
+  root HashMapRoot
+  parameter hashAlg String
```
rvagg, line 129: I'm tempted to suggest an "optional" on the end here too, I can see use cases for these things being always required by a schema reference and sometimes being left up to other layers (programatic) or allowing for flexibility in the data that's loaded.

rvagg:
And an additional thought re the Filecoin usage: parameterisation and strict definition when an advanced layout is used could be the path to resolving that concern that @whyrusleeping raised of not wanting to clutter up root blocks with the HAMT parameters when they're always going to be fixed within the Filecoin. We revisit the HashMap spec, stick an `optional` on the end of `bucketSize` and `hashAlg` and let the HashMap logic decide when these are required based on what it receives from the schema ("receives" being door to a much larger discussion about how these things interact, but codegen being an obvious starting point).

warpfork:
I like the use of the word "receives" in that last comment.

Mild variation on that idea I'd suggest toying with: two separate, but nearly identical schemas: one of which has those fields (required, not optional) and another which doesn't.  This would better reflect that a dataset which had a `bucketSize` but was missing a `hashAlg` is nonsense, for example.

I don't know if this would seem "convenient" or ergonomically well-supported at present, but it's an idea I'd like to roll around and see if we can't build towards support for some day, because it seems like it would describe a lot of real-world desires well.

rvagg:
> Mild variation on that idea I'd suggest toying with: two separate, but nearly identical schemas: one of which has those fields (required, not optional) and another which doesn't. This would better reflect that a dataset which had a bucketSize but was missing a hashAlg is nonsense, for example.

So, by "fields" you mean fields in the actual block data? So a form that receives these parameters entirely from the schema (again, _somehow_, :handwave:) without baking them into the root block like the current HashMap spec does.

I don't know if it's worthwhile but I'm starting to think of IPLD as carving out two types of paths—or maybe it's a spectrum of two types of uses—where on one hand you may want to lock down to _one way_, like in Filecoin, where it's an entire monolith of these pieces put together into a firmly specified form that isn't intended to change over time. And on the other hand we are offering more generic tools that can be pieced together in flexible ways, with dials to tweak. You could be in one camp or the other or you could migrate from the latter to the former as your usecase solidifies and you figure out the shape of the data for your application.

Specifying parameters in your schema certainly locks them in and allows for no variation in data that's found along the way. I suppose that's a good way to use these things. My personal need for flexibility makes me want to leave options open, partly as an acknowledgement that we honestly don't know how all of these data structures are going to perform with real data in their varied uses. There's so many variables that would make you tweak these dials, it's entirely case dependent but we don't have good ways of quantifying that yet and may not for some time, it's mostly guesswork and supposition. (I think this is a digression, but it's related to the motivations for wanting these options and whether there's a "right" way).

vmx (approved):
I really like how this turns out.

PR #185, opened 2019-08-30
schemas: fully switch to "implicit", use "discriminant" consistently

rvagg:
1. Completes the transition to "implicit" from "default".
2. Uses "discriminant" consistently ("discriminantKey", "discriminantTable"). Trying to clearly identify the standard usage meaning of "discriminant" and "discriminator" is a bit tricky, but @warpfork preferred the former and it seems to lean slightly more in the right direction.

### DIFF
--- a/schemas/schema-schema.ipldsch
+++ b/schemas/schema-schema.ipldsch
@@ -54,9 +54,9 @@ type Schema union {
 ##
 ## The Type union is serialized using "inline" union representation,
 ## which means all of its members have map representations, and there will be
-## an entry in that map called "type" which contains the union discriminator.
+## an entry in that map called "type" which contains the union discriminant.
 ##
-## Some of the kinds of type are so simple the union discriminator is the only
+## Some of the kinds of type are so simple the union discriminant is the only
 ## content at all, e.g. strings:
 ##
 ## ```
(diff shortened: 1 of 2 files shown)


rvagg:
Matching JS schema parser/validator changes @ https://github.com/rvagg/js-ipld-schema/pull/5

schemas/schema-schema.ipldsch
```diff
@@ -365,42 +366,42 @@ type StructRepresentation union {
...
-## Note that fields with defaults are distinct from fields which are optional!
+## Note that fields with implicits are distinct from fields which are optional!
 ## The cardinality of membership of an optional field is is incremented:
 ## e.g., the cardinality of "fieldname Bool" is 2; "fieldname optional Bool" is
 ## membership cardinality *3*, because it may also be undefined.
-## By contrast, the cardinality of membership of a field with a default value
+## By contrast, the cardinality of membership of a field with an implicit value
 ## remains unchanged; there is serial state which can map to an undefined value.
 ##
 ## Note that 'alias' supports exactly one string, and not a list: this is
```
warpfork, line 389: Not necessarily to tack onto this PR, but while I'm looking at it: @dignifiedquire suggested `rename` as an alternative name (ahheh) for `alias`.  ("Alias" is a bit problematic because for many readers it connotes that one of two names will be accepted.  "Rename" is what rust calls this.)

warpfork (approved):

schemas/schema-schema.ipldsch
mikeal: I’m +1 on the change to rename, but want to note that this should go in ASAP so that I can use it before we finalize the unixfsv2 schema.

rvagg: it's in :thumbsup:

PR #186, opened 2019-09-02
schemas: complete map and struct representations for schema-schema

rvagg:
Brings schema-schema into parity with https://github.com/ipld/specs/blob/master/schemas/representations.md

Adds stringpairs, listpairs, stringjoin for map and struct as appropriate.

The `Representation_StringPairs` and `Representation_ListPairs` structs are without the prefixes (`Struct` and `Map`) because they are shared between both. This works out nicely in code too but I'm not sure if it's the best way to do it here?

schema-schema.ipldsch.json generated by js-ipld-schema

### DIFF
--- a/schemas/schema-schema.ipldsch
+++ b/schemas/schema-schema.ipldsch
@@ -177,8 +177,8 @@ type TypeMap struct {
 ##
 type MapRepresentation union {
 	| MapRepresentation_Map "map"
-	| Representation_StringPairs "stringpairs"
-	| Representation_ListPairs "listpairs"
+	| MapRepresentation_StringPairs "stringpairs"
+	| MapRepresentation_ListPairs "listpairs"
 } representation keyed
 
 ## MapRepresentation_Map describes that a map should be encoded as
@@ -186,8 +186,8 @@ type MapRepresentation union {
 ##
 type MapRepresentation_Map struct {}
 
-## Representation_StringPairs describes that a map, or struct, should
-## be encoded as a string of delimited "k/v" entries, e.g. "k1=v1,k2=v2".
+## MapRepresentation_StringPairs describes that a map should be encoded as a
+## string of delimited "k/v" entries, e.g. "k1=v1,k2=v2".
 ## The separating delimiter may be specified with "entryDelim", and the k/v
 ## delimiter may be specified with "innerDelim". So a "k=v" naive
 ## comma-separated form would use an "innerDelim" of "=" and an "entryDelim"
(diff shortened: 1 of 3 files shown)


schemas/schema-schema.ipldsch
```diff
@@ -405,6 +452,21 @@ type StructRepresentation_Tuple struct {
...
+## values are concatenated, in order, and separated by a "join" delimiter.
+## For example, specifying ":" as the "join": "v1,v2,v3".
+## stringjoin is necessarily restrictive and therefore only valid for structs
+## whose values may all be encoded to string form without conflicts in "join"
+## character. It is recommended, therefore, that its use be limited to structs
+## containing values with the basic data model kinds that exclude multiple
+## values (i.e. no maps, lists, and therefore structs or unions).
+type StructRepresentation_StringJoin struct {
+	join String
+	fieldOrder optional [String]
```
warpfork: I wonder if we should consider introducing a `type FieldName string`, for roughly the same reasons it's been useful to have `type TypeTerm string`.

schemas/schema-schema.ipldsch
```diff
@@ -361,6 +405,9 @@ type InlineDefn union {
 type StructRepresentation union {
 	| StructRepresentation_Map "map"
 	| StructRepresentation_Tuple "tuple"
+	| Representation_StringPairs "stringpairs"
+	| StructRepresentation_StringJoin "stringjoin"
+	| Representation_listpairs "listpairs"
```
warpfork: Temptation: to consider a 'type alias' feature.

I dunno if I wanna broach that at the moment though without putting in some work to come up with more examples of what that could be like.

rvagg:
(from discussion elsewhere) letting this PR bake for a bit to get broader mindshare, it's in our docs so we've flagged it as a feature, but maybe there's reasonable objection to even allowing this kind of thing in (see edge cases notes, they're pretty edgy).

rvagg:
@warpfork have a look at 18a3dce where I've tried a `FieldName`. I'm still unsure about its place in `fieldOrder` though and whether that means that we allow barewords or quoted strings in there. Is it `fieldOrder ["Foo", "Bar"]` or `fieldOrder [Foo, Bar]`? Currently imposing the former but does switching it to `FieldName` imply something new?

warpfork:
Yeah, I think introducing FieldName makes all those things it touches look better.  Big thumbs up.

`fieldOrder ["Foo", "Bar"]`, as currently, still seems clear.  Coercing `"foo"` to a `FieldName` is just fitting a string to a string.

schemas/schema-schema.ipldsch
warpfork: Coming back to this with fresh eyes: I think I *would* rather "duplicate" things (e.g. have StructRepresentation_StringPairs and also MapRepresentation_StringPairs).

Reason: it's "coincidental" (relatively) that similar strategies exist for representing maps and structs.  The fact that structurally similar configurations are required for them presently doesn't indicate they should be coupled.

rvagg:
rebased, squashed, and added a35f676 which separates out the Map and Struct forms of StringPairs and ListPairs as per https://github.com/ipld/specs/pull/186#discussion_r322006869. FWIW I disagree with the reasoning, it's not coincidental as they are both essentially map encodings. But for clarity it's fine and it allows for distinct documentation.

rvagg:
ready to land, will do so tomorrow unless someone raises a concern

vmx (approved):

warpfork:
> it's not coincidental as they are both essentially map encodings. But for clarity it's fine and it allows for distinct documentation.

I'm still oscillating on this as well, it's true they're both essentially map encodings and that makes it less of a 'coincidence'.  But the clarity and distinct docs arguments seem to still be winners.

warpfork (approved):

PR #187, opened 2019-09-03
schemas: replace "alias" with "rename"

rvagg:
This has been floating around and mostly adopted in Goland. A better word than "alias" for a translation of the key to the serialized form.

JS support @ https://github.com/rvagg/js-ipld-schema/pull/6

Test fixture of

```ipldsch
  type StructAsMapWithRenames struct {
    bar Bool (rename "b")
    boom String
    baz String (rename "z")
    foo Int (implicit "0" rename "f")
  }
```
comes out as

```json
  {
    "StructAsMapWithRenames": {
      "kind": "struct",
      "fields": {
        "bar": {
          "type": "Bool"
        },
        "boom": {
          "type": "String"
        },
        "baz": {
          "type": "String"
        },
        "foo": {
          "type": "Int"
        }
      },
      "representation": {
        "map": {
          "fields": {
            "bar": { "rename": "b" },
            "baz": { "rename": "z" },
            "foo": {
              "implicit": 0,
              "rename": "f"
            }
          }
        }
      }
    }
```

### DIFF
--- a/schemas/overview.md
+++ b/schemas/overview.md
@@ -43,8 +43,9 @@ All of this maps over the IPLD Data Model:
 
 - every type has a representation;
 - e.g. maps are represented as maps; structs are also represented as maps!
-- representations can specify aliases for how to serialize things; e.g. provide
-  shorter serial strings to use in place of longer struct names.
+- representations can specify a mapping for serialized field names, which can
+  be useful in cases such as requiring serialization compactness but preferring
+  verbosity in the schema for descriptive purposes.
 - representations are customizable, including mapping a type onto a different
   representation kind entirely!  e.g. a struct can be declared to have a
   list representation (making it shorter to serialize, if less self-describing);
--- a/schemas/schema-schema.ipldsch
+++ b/schemas/schema-schema.ipldsch
@@ -364,19 +364,27 @@ type InlineDefn union {
 ## StructRepresentation describes how a struct type should be mapped onto
 ## its IPLD Data Model representation.  Typically, maps are the representation
 ## kind, but other kinds and details can be configured.
+##
 type StructRepresentation union {
 	| StructRepresentation_Map "map"
 	| StructRepresentation_Tuple "tuple"
 } representation keyed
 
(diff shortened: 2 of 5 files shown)


schemas/overview.md
```diff
@@ -43,8 +43,9 @@ All of this maps over the IPLD Data Model:
 
 - every type has a representation;
 - e.g. maps are represented as maps; structs are also represented as maps!
-- representations can specify aliases for how to serialize things; e.g. provide
-  shorter serial strings to use in place of longer struct names.
+- representations can specify aliases (referred to as a "rename" in the schema
```
warpfork: Oh, boy, that's stilted now.

Maybe we can try out some phrasing using the word "remapped"?  Idea spawned by looking at other text elsewhere, also commented there.

schemas/schema-schema.ipldsch
```diff
@@ -370,13 +370,13 @@ type StructRepresentation union {
 } representation keyed
 
 ## StructRepresentation_Map describes a way to map a struct type onto a map
-## representation.  For example, fields may be aliased,
-## or implicit values associated.
+## representation.  For example, fields may be renamed, or implicit values
```
warpfork: How's this sound:

> For example, the keys that fields are emitted under may be remapped using the 'rename' instruction,

schemas/overview.md
rvagg: > - representations can specify a mapping for serialized field names, which can
>  be useful in cases such as requiring serialization compactness but preferring
>  verbosity in the schema for descriptive purposes.

schemas/schema-schema.ipldsch
rvagg: nah, don't like that either, how about this:

> Field serialization options may optionally be configured to enable mapping serialized keys using the 'rename' option, or implicit values specified where the field is omitted from the serialized form using the 'implicit' option.

rvagg:
Merging this to unblock other progress, I expect that in the small likelihood that there's outstanding quibbles that they are about minor language tweaks. We can address in subsequent PRs to this same area.

PR #188, opened 2019-09-03
schemas: specify Any and AnyScalar

rvagg:
First attempt at specifying `Any`, also introduces `AnyScalar` and adds some words about `&` usage into `TypeLink` comments. The example.ipldsch is also adjusted to include links and it has the missing `"expectedType"` where it's `Any` in the first instance.

### DIFF
--- a/schemas/schema-schema.ipldsch
+++ b/schemas/schema-schema.ipldsch
@@ -135,6 +135,18 @@ type RepresentationKind enum {
 	| "link"
 }
 
+## AnyScalar defines a union of the basic non-complex kinds.
+##
+## Useful defining usage of IPLD nodes that do compose from other nodes.
+##
+type AnyScalar union {
+	| Bool bool
+	| String string
+	| Bytes bytes
+	| Int int
+	| Float float
+} representation kinded
+
 ## TypeBool describes a simple boolean type.
 ## It has no details.
 ##
@@ -178,15 +190,25 @@ type TypeList struct {
 } representation map
 
 ## TypeLink describes a hash linking to another object (a CID).
+##
 ## A link also has an "expectedType" that provides a hinting mechanism
 ## suggesting what we should find if we were to follow the link. This
 ## cannot be strictly enforced by a node or block-level schema
 ## validation but may be enforced elsewhere in an application relying on
 ## a schema.
+##
 ## The expectedType is specified with the `&Any` link shorthand, where
 ## `Any` may be replaced with a specific type.
+##
+## Unlike other kinds, we use `&Type` to denote a link Type rather than
+## `Link`. In this usage, we replace `Type` the expected Type, with `&Any`
+## being shorthand for "a link which may resolve to a type of any kind".
+##
+## `expectedType` is a String, but it should validate as "Any" or a TypeName
+## found somewhere in the schema.
+##
 type TypeLink struct {
-	expectedType TypeTerm (implicit "Any")
+	expectedType String (implicit "Any")
 }
 
 ## TypeUnion describes a union (sometimes called a "sum type", or
@@ -406,7 +428,7 @@ type StructRepresentation_Map struct {
 ##
 type StructRepresentation_Map_FieldDetails struct {
 	rename optional String
-	implicit optional Any # Review: may be better to introduce a small kinded union here which has the essential scalars as members.
+	implicit optional AnyScalar
 }
 
 ## StructRepresentation_Tuple describes a way to map a struct type into a list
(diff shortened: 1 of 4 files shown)


vmx:
Sounds good to me, but it should really be approved by @warpfork.

schemas/schema-schema.ipldsch
```diff
@@ -135,6 +135,35 @@ type RepresentationKind enum {
 	| "link"
 }
 
+## Any is a union of the basic kind types (minus links). Any is used to
+## specify that an IPLD node may be of any of the basic data model kinds.
+##
+## Links are left out of this union for now, and therefore any element in
+## the schema DSL (`&Any`) that uses it. The placement of links should be
+## intentional when designing schemas.
+##
```
warpfork:
```suggestion
+## Note that "Any" also shows up as a TypeName string literal in some parts
+## of the schema-schema, such as being the implicit value of the TypeName
+## for the expectedType field of a TypeLink.
+##
```

schemas/schema-schema.ipldsch
```diff
@@ -135,6 +135,35 @@ type RepresentationKind enum {
...
+type Any union {
+	| TypeBool "bool"
+	| TypeString "string"
+	| TypeBytes "bytes"
+	| TypeInt "int"
+	| TypeFloat "float"
+	| TypeMap "map"
+	| TypeList "list"
+} representation kinded
```
warpfork: Here and below, wee typo: kinded unions need no quotes, because kinds are keywords.  (The quoted strings wouldn't show up anywhere in serial form, either, so quoting them would be confusing.)

warpfork: Also, wrong level of meta.   'Any' contains any *value*, not the description of any *type*.  Thus:

```suggestion
-       | TypeBool "bool"
-       | TypeString "string"
-       | TypeBytes "bytes"
-       | TypeInt "int"
-       | TypeFloat "float"
-       | TypeMap "map"
-       | TypeList "list"
+       | Bool bool
+       | String string
+       | Bytes bytes
+       | Int int
+       | Float float
+       | Map map
+       | List list
```

warpfork: Oh shoot.  There's no great definition of `Map` for this purpose.  It would have to be something like `{String:Any}` but that seems a bit shakey itself for putting into the prelude of known type names.  I'm not sure what to do about this.

warpfork:
@rvagg, I think you'd suggested in earlier discussion that you're suspicious "Any" is more complicated than this, and might need a special construct, and I'm thinking you were right.

A link of '&Any' can certainly point to a union or a struct.  This doesn't fit with the idea of a kinded union.  We'd seem to need a special construct for that.

And that does seem to be what the use in e.g. HAMTs internals requires.  Link targets in general, too.

`AnyScalar`, on the other hand, seems to still be clear without this (nor run into any of the other tricky issues).  So maybe we should proceed with just that for now -- and it is, fortunately, the only thing we really want in the schema-schema itself.

rvagg:
I've removed `Any` but left `AnyScalar`.

I've also changed `Link`'s `expectedType` to `String` instead of `TypeTerm` because we can't supply an implicit for a non-scalar. Parsers are just going to have to special-case that one. schema-schema.ipldsch now parsers properly and the .json is updated too here (it didn't parse properly with an implicit for `TypeTerm`).

@warpfork see lastest changes.

warpfork:
Ah right that's the other thing I was trying to remember to mention.

We can make `Link`'s `expectedType` a `TypeName` rather than a `TypeTerm`, and that clears that up because then the implicit is just scalar content again.  I was a doofus in saying `TypeTerm` at earlier points -- forgot the difference between the two.

rvagg:
> We can make Link's expectedType a TypeName rather than a TypeTerm

That still introduces complications. The parser has to see the type that the "implicit" needs to be coerced to and make sure it can perform such coercion. If we restrict "implicit" usage to only AnyScalar types then it's straightforward. If we want to make TypeName able to have an implicit then we need to either have TypeName declared before this block (first introduction of ordering, first declare-before-use) or do a double pass through the parsed "AST" to do the type checking for implicits after we've loaded all the types and can determine that TypeName is really just a string kind and we can coerce that.

So ... How about we just a restriction for now that you can only use "implicit" on AnyScalar typed struct properties and let Link's expectedType be a string to satisfy that?

warpfork:
Hm.  I see the concern about a double pass, but think it might be fine and justified -- I'm pretty sure there's other parts of schema validation that *only* are possible in a second pass anyway, so we'll already have one (e.g. the reference checks for type names already poke the exact same nerve, and declare-before-use isn't an option at all since that case is freely recursive).

But I'm also fine with pushing that off down the road if you'd rather.  You're closer to the implementation of that than I am at the moment, I think.

warpfork (approved):

rvagg:
> But I'm also fine with pushing that off down the road if you'd rather

:thumbsup: this change doesn't preclude further iteration in that direction.

PR #189, opened 2019-09-05
Add exploration report about compression

### LINKED ISSUE #76: IPLD and compression
Compression came already up [several](https://github.com/ipfs/ipfs/issues/47) [times](/discuss.ipfs.io/t/any-work-on-compression-of-ipld-data/2505). I think there should be some kind of compression for IPLD.

I can think of two different ways which serve different purposes.

### On the application level

Add a new container format which is about compression which embeds the original one. There's several ways doing that, one could be by just adding a new format for any compression method, the format implementation would then extract the data and you could access the underlying one. Hashes would then be the one of the compressed data.

Use cases for doing it this way are when you are rather collecting data (e.g. with sensors) and don't look at them too often.

### Deep inside IPLD/IPFS

Compression could also happen on a layer below IPLD. Then the hashes would be based on the actual (uncompressed) content. The idea had is that you would implement it on two layers, the storage and the transport.

The storage layer (the repo) would be enabled with compression and just compress all the data that comes in. When you retrieve the data, it would be uncompressed by default. So things would work as they do today. Though there could be also an option to return the compressed data, which could then be passed on the transportation layer

The transportation layer takes an uncompressed block by default, compresses it for the transport and uncompressed it again on the other side. Though there could be an option that it takes the compressed data directly of the storage layer and transmit that data. Then it would only need to be de-compressed on the receiving side to verify the hashes match.

vmx:
This report was originally issue #76.

Closes #76.

### DIFF
--- a/design/history/exploration-reports/2018.11-compression.md
+++ b/design/history/exploration-reports/2018.11-compression.md
@@ -1,0 +1,157 @@
+This document was archived from https://github.com/ipld/specs/issues/76.
+
+#76: IPLD and compression
+-------------------------
+Opened 2018-11-05T16:24:20Z by vmx
+
+Compression came already up [several](https://github.com/ipfs/ipfs/issues/47) [times](/discuss.ipfs.io/t/any-work-on-compression-of-ipld-data/2505). I think there should be some kind of compression for IPLD.
+
+I can think of two different ways which serve different purposes.
+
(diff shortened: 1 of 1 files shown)


mikeal:
LGTM

rvagg (approved):

PR #194, opened 2019-09-13
schema: introduce byteprefix union representation strategy

rvagg:
Introduced in Filecoin, not an unreasonable representation strategy for a union types that are all represented as bytes.

Ref: https://github.com/filecoin-project/specs/

### DIFF
--- a/schemas/representations.md
+++ b/schemas/representations.md
@@ -43,6 +43,7 @@ Link is also not described in this section, as it's a rather unique business.)
 	- `kinded` representation -- transcribes to varying (!) kinds in the Data Model.
 	- `envelope` representation -- transcribes to a dual-entry `map` in the Data Model.
 	- `inline` representation -- transcribes to a `map` in the Data Model (and has additional limitations).
+	- `byteprefix` representation -- transcribes to `bytes` in the Data Model, only usable for unions of `bytes`.
 - Struct
 	- `map` representation -- the default -- transcribes to `map` in the Data Model.
 	- `tuple` representation -- transcribes to `list` in the Data Model.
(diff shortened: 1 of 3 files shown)


vmx (approved):

rvagg:
Switched from `discriminantTable {Int:TypeName}` to `discriminantTable {TypeName:Int}` .. because we don't have string keys.

So as per https://github.com/ipld/js-ipld-schema/pull/11:

```ipldsch
  type Bls12_381Signature bytes
  type Secp256k1Signature bytes
  type Signature union {
    | Secp256k1Signature 0
    | Bls12_381Signature 1
  } representation byteprefix
```

becomes:

```json
   {
    "Bls12_381Signature": {
      "kind": "bytes"
    },
    "Secp256k1Signature": {
      "kind": "bytes"
    },
    "Signature": {
      "kind": "union",
      "representation": {
        "byteprefix": {
          "Secp256k1Signature": 0,
          "Bls12_381Signature": 1
        }
      }
    }
```

which is flipped from the other union tables

rvagg:
Rebased and squashed. @warpfork this needs your blessing or alternative suggestions.

I'm thinking that this is similar to #196 in that the thing we want to look up is the value rather than the key (int enums where we get the int from the data model and want to convert it to some other entity). They're both going to require either some more efficient conversion mechanism in codegen and/or internal representation data that provides for more efficient lookup—in a similar way to the ordering retention problem we have whenever maps get involved.

warpfork (approved):
I'm kind of frustrated by the inversion of the map order, as you noted above.  I don't currently have a better mechanism for solving it in mind, though, so overall, I agree.

I pushed a commit nudging some docs around ever so slightly, so you might want to recheck those to see if they pass your own muster, but otherwise I'd say this is good to go.

PR #196, opened 2019-09-13
schema: int representation for enums

rvagg:
I've done some non-trivial surgery here to make this happy. I copied the union representation structure in schema-schema and removed the innards of `enum`, pushing the "members" into the representations. So in schema-schema we have this:

```ipldsch
type RepresentationKind enum {
	| "bool"
	| "string"
	| "bytes"
	| "int"
	| "float"
	| "map"
	| "list"
	| "link"
}
```

maps to this

```json
		"RepresentationKind": {
			"kind": "enum",
			"representation": {
				"string": {
					"bool": null,
					"string": null,
					"bytes": null,
					"int": null,
					"float": null,
					"map": null,
					"list": null,
					"link": null
				}
			}
```

And as per the docs, I'm proposing that this (from Filecoin):

```ipldsch
type Options enum {
    | Blocks 0
    | Messages 1
    | BlocksAndMessages 2
} representation int
```

(note the unquoted strings this time)

would map to this:

```json
		"Options": {
			"kind": "enum",
			"representation": {
				"int": {
					"Blocks": 0,
					"Messages": 1,
					"BlocksAndMessages": 2
				}
			}
```

The mapping of `{String:Int}` isn't as convenient for lookups but I'm sure codegen could make that cost go away.

I also made some cosmetic alterations to representations.md while I was in there.

### DIFF
--- a/schemas/representations.md
+++ b/schemas/representations.md
@@ -1,6 +1,20 @@
 Kinds and their Representations
 -------------------------------
 
+* [Kinds and their Representations](#kinds-and-their-representations)
+* [Available representations](#available-representations)
+* [Representation Strategy Reference](#representation-strategy-reference)
+  * [struct map representation](#struct-map-representation)
+  * [struct tuple representation](#struct-tuple-representation)
+  * [struct stringjoin representation](#struct-stringjoin-representation)
+  * [map stringpairs representation](#map-stringpairs-representation)
+  * [union keyed representation](#union-keyed-representation)
+  * [union kinded representation](#union-kinded-representation)
+  * [union envelope representation](#union-envelope-representation)
+  * [union inline representation](#union-inline-representation)
+  * [enum string representation](#enum-string-representation)
+  * [enum int representation](#enum-int-representation)
+
 A type at the Schema level must be mapped onto a representation expressible
 within the Data Model.
 
@@ -26,7 +40,7 @@ more detailed parameters which are specific to the strategy (e.g., the struct
 map representation has a concept of 'default' values which is unique to that
 strategy).
 
-### table
+## Available representations
 
 (Kinds for which no meaningful customization is possible -- Null, Boolean,
 Integer, Float, String, Bytes, and List -- are elided.
@@ -52,19 +66,17 @@ Link is also not described in this section, as it's a rather unique business.)
 	- `listpairs` representation -- transcribes to `list` (of lists) in the Data Model.
 - Enum
 	- `string` representation -- the default -- transcribes the enum values as `string` in the Data Model.
+	- `int` representation -- transcribes the enum values as `int` in the Data Model.
 
 Each of these representation strategies will be explored in detail in the
 "Representation Strategy Reference" section, below.
 
-
-Representation Strategy Reference
----------------------------------
+## Representation Strategy Reference
 
 In this section, we'll go into detail on each representation strategy,
 and include examples of both schemas using it and data matching it.
 See the "Kinds and their Representations" section for an introduction to
 representation strategies.
-
 
 ### struct map representation
 
@@ -72,7 +84,7 @@ Map representation of structs means that the struct is represented as a map,
 where the keys are the names of the struct fields.  This is a common and
 natural way to represent structs; it's thus their default representation.
 
-#### struct map representation example
+**Example**
 
 ```ipldsch
 type Foo struct {
@@ -93,7 +105,6 @@ Some data matching the `Foo` struct (shown as JSON) is:
 }
 ```
 
-
 ### struct tuple representation
 
 Tuple representation allows structs to be packed into a list representation.
@@ -106,7 +117,7 @@ in the way of "self-describing" information, tuple representations can make for
 very fragile protocols, increase the difficulty of migrations, and make
 serialized data incomprehensible without the schema information in hand.
 
-#### struct tuple representation example
+**Example**
 
 ```ipldsch
 type Foo struct {
@@ -126,7 +137,7 @@ Notice how this is the same data as in the
 it's just much more compact than it was in the map representation (and
 beware -- correspondingly less self-describing!).
 
-#### struct stringjoin representation example
+### struct stringjoin representation
 
 ```ipldsch
 ## Fizzlebop is a pair of fields which serializes as "value-of-a:value-of-b" as a string.
@@ -147,7 +158,7 @@ Some data matching the `Fizzlebop` struct (shown as JSON) is:
 Since this is a struct, and none of the fields are optional, it *must* have two fields:
 therefore, a string with no `":"` characters would be rejected as not matching.
 
-#### map stringpairs representation example
+### map stringpairs representation
 
 Say we're doing something awfully like the mount options in an /etc/fstab file:
 
@@ -167,7 +178,7 @@ Some data matching the `MountOptions` struct (shown as JSON) is:
 "keys=values,serialized=thusly"
 ```
 
-#### union keyed representation example
+### union keyed representation
 
 ```ipldsch
 type MyKeyedUnion union {
@@ -191,7 +202,7 @@ This data would also match, as the other type:
 {"bar": 12}
 ```
 
-#### union kinded representation example
+### union kinded representation
 
 ```ipldsch
 type MyKindedUnion union {
@@ -233,7 +244,7 @@ Note that the kinds valid in a kinded union are all Data Model-layer
 `struct` is not.  The kind that a union member type is tagged with must
 match that type's representation kind or the schema is invalid.
 
-#### union envelope representation example
+### union envelope representation
 
 ```ipldsch
 type MyEnvelopeUnion union {
@@ -260,7 +271,7 @@ This data would also match, as the other type:
 {"tag":"bar", "msg":12}
 ```
 
-#### union inline representation example
+### union inline representation
 
 ```ipldsch
 type MyInlineUnion union {
@@ -308,3 +319,44 @@ type Bls12_381Signature bytes
 At the block level, this presents as a byte array, where the first byte is the discriminator (`0x00` or `0x01`) and the remainder is sliced to form either of the two types depending on the discriminator.
 
 `byteprefix` is not valid for unions where any of the constitutive types are _not_ `Bytes`.
+
+### enum string representation
+
+By default, an enum is simply represented as a string in the data model. An
+`enum` in the schema simply defines the list of possible strings that could be
+used in that position.
+
+**Example**
+
+```ipldsch
+type Status enum {
+	| NOPE
+	| YEP
+	| MAYBE
+}
+```
+
+This enum dictates that where `Status` is used, we should find one of `"NOPE"`,
+`"YEP"` or `"MAYBE"`.  No other value is valid where `Status` is used.
+
+### enum int representation
+
+An alternative representation for enums is the "int" representation that is
+closer to what users may expect from enums in certain programming languages
+that map enum values to integers. In IPLD Schemas we explicitly define the
+mapping to integers, so the user can dictate the appropriate data model values.
+
+**Example**
+
+```ipldsch
+type Status enum {
+	| NOPE  0
+	| YEP   1
+	| MAYBE 100
+} representation int
+```
+
+Note that for an "int" representation enum we don't quote the strings as they
+are no longer our data model values. Instead, the ints provided after each of
+the names are the values we find in the data model. In this case, this enum
+dictates that where `Status` is used, we should find one of `0`, `1` or `100`.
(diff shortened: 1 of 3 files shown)


schemas/representations.md
```diff
@@ -51,27 +69,25 @@ Link is also not described in this section, as it's a rather unique business.)
...
 representation strategies.
 
-
 ### struct map representation
 
 Map representation of structs means that the struct is represented as a map,
 where the keys are the names of the struct fields.  This is a common and
 natural way to represent structs; it's thus their default representation.
 
-#### struct map representation example
```
warpfork, line 75: undesirable.

I know these read redundantly as heck, and I'm not a huge fan of that either... but they are as such so that `thepage.md#struct-map-representation-example` works.  :/

warpfork: we could try to do our own `<a ...>` tags, but the little get-me-a-link button that github draws wouldn't work with that either, sadly.

rvagg: yes, but they suck verbose like that, so nice hierarchy is more desirable to me

rvagg: how about just removing that level of heading and going with bolded `**Example**`, that renders nicely in MD

schemas/representations.md
```diff
@@ -291,3 +306,44 @@ This is because inline unions are only a defined concept when working with types
...
+`enum` in the schema simply defines the list of possible strings that could be
+used in that position.
+
+#### Example
+
+```ipldsch
+type Status enum {
+	| "NOPE"
+	| "YEP"
+	| "MAYBE"
```
warpfork: Hm.  Should we maybe ... infer strings as the serial form by default, but make the syntax for all the representations use something that looks more "like" a TypeName?

E.g.:

```
type Status enum {
	| Nope
	| Yep
	| Maybe
}
```

and this is equivalent to:

```
type Status enum {
	| Nope "Nope"
	| Yep "Yep"
	| Maybe "Maybe"
}
```

I'm thinking of this in part because of the need of codegen to generate reasonable symbols, such as in this case:

```go
const (
	Status_Nope  = Status("Nope")
	Status_Yep   = Status("Yep")
	Status_Maybe = Status("Maybe")
)
```

schemas/schema-schema.ipldsch.json
```diff
@@ -39,31 +39,35 @@
...
-				"bytes": null,
-				"int": null,
-				"float": null,
-				"map": null,
-				"list": null,
-				"link": null,
-				"union": null,
-				"struct": null,
-				"enum": null
+			"representation": {
```
warpfork, line 70: Hrm.  I'm a little disquieted by having the whole of the semantic content end up in the representation block, and none in directly on the type.

But I can't think of any alternatives that don't make for fairly staggering redundancy.  So... hrm.  (This might be just a lamenting note rather than a request for anything different.)

schemas/representations.md
rvagg: Yes, I was heading in that direction when I introduced `type EnumValue string` to mimmick `type TypeName string` which is unquoted, but backed out some of my changes. It's awkward not having ways to express this in schema-schema.
If you're OK with this then I can go ahead and unquote.

schemas/schema-schema.ipldsch.json
rvagg: I've felt the same way about union types.
My first attempt here was to change `members` to `{String:EnumValue}` and have `EnumValue` be a kinded union of `Null null` and `IntValue int` .. but then remembered that we have the same awkward situation with unions and we can more carefully enforce the types if we branch by representation and keep the kinded union out of the mappings.

schemas/representations.md
rvagg: See latest fixup commit and view that: https://github.com/ipld/specs/blob/19691d41261bfda89a4eab63267e057aa21b9ddd/schemas/representations.md

The sections are small enough that linking straight into an example isn't so helpful. Just link to the section.

rvagg:
Cleaned up a bit, removed quotes from enum values as per discussion above.

Put up a JS implementation, test fixtures are interesting to look at:

https://github.com/ipld/js-ipld-schema/blob/rvagg/enum-int/test/fixtures/bulk/enum-int.yml

and string enum:

https://github.com/ipld/js-ipld-schema/blob/rvagg/enum-int/test/fixtures/bulk/enum.yml

@warpfork this needs your blessing.

schemas/schema-schema.ipldsch
```diff
@@ -116,17 +116,17 @@ type Type union {
...
-	| "bytes"
-	| "int"
-	| "float"
-	| "map"
-	| "list"
-	| "link"
-	| "union"
-	| "struct"
-	| "enum"
+	| bool
```
warpfork: My first inclination would be to make these Snake/TitleCase, on the logic that so far lowercase is showing up in keywords and field names.  Thoughts?

(Docs above also have SCREAMING case, which I'm just never a fan of, personally.)

warpfork:
My slightly-alternate proposal would be to stay strong on membership being stored in the type body, and then handling any custom string names or int representations in much the same pattern as how we treat 'rename' directives on struct fields.

E.g.,

```
type Woo enum {
	| Nope "n"
	| Yep "y"
	| Maybe
}
```

would result in:

```

"SimpleEnum": {
	"kind": "enum",
	"members": {
		"Nope": null,
		"Yep": null,
		"Maybe": null
	},
	"representation": {
		"string": {
			"Nope": "n",
			"Yep": "y"
			// note lack of 'maybe' key
		}
	}
```

(The behavior for ints would be similar, except since there's no sane default, it would be invalid to be lacking a representation map entry for any of the members.)

This is more redundant in the case of many custom names (or, yes, for all 'int' representations).  However, there's a choice to weigh here: is avoiding redundancy in the schema AST something we need to significantly optimize for, or do we give priority to the "type semantics in the type block, serial semantics in the representation block" rule?

I tend to think avoiding redundancy in the AST ranks pretty low.

---

There might be some implications for round-tripping DSL->AST->DSL:

```
type Whee enum {
	| Foo "Foo" # redundant -- noncanonical?
	| Bar
}
```

But perhaps the guideline of "act like struct field renames" should again apply.  In this case that would mean redundantly stating the same string as would be inferred by default is simply not handled specially (and is thus retained).

warpfork:
Speaking of the "type semantics in the type block, serial semantics in the representation block" rule --

There's an argument to be made that we might want to put the strings or ints for the representation in parens, for consistency again with how we use parans elsewhere (namely struct fields and the 'rename' feature there) to denote things that are representational, but got syntactically translocated up to the type block in the DSL for human-author convenience.

E.g.:

```
type Woo enum {
	| Nope (serial "n")
	| Yep (serial "y")
	| Maybe
}
```

and

```
type Whee enum {
	| Foo (serial 1)
	| Bar (serial 4)
} representation int
```

Otoh, that's getting awfully high in character count, and since I can't imagine any other features appearing in the future which would also reside in those parens, it seems net boilerplatey.  So, I'm posting this as a thinking-out-loud, but not sure I'm actually advocating for it.

rvagg:
I'm fine with the JSON proposal except for the discrepancy with the way we handle unions. Can you think up some words re why we're doing the double-up in enums but unions get everything shifted into their representations?

Re `| Foo (serial 1)`, I'm fine with that and don't think it's an entirely unreasonable ask for schema authors. It's nice and clear what's going on too (I wish I could come up with a better word than "serial" though).

Note that in `implicit` we are quoting the value, regardless of kind, so maybe we need to follow that here too? `| Foo (serial "1")`, where `representation int` is left to do the work of forcing these to integers.

One possibility to simplify this is to just drop the keyword and paren the serialized value: `| Foo ("1")`, or `| Nope ("nopez")`. That's a mild departure from previous parsing rules for parens but not a terrible departure and would still allow additional representation rules be appended after a first quoted one.

Let's try that on for size:

```
type Woo enum {
	| Nope ("n")
	| Yep ("y")
	| Maybe
}

type Whee enum {
	| Foo ("1")
	| Bar ("4")
} representation int
```

warpfork:
**discrepancy with unions**: shoot.  You're right.  That *is* discrepant, and we should fix that one way or another.

I seem to have originally written unions this way as well, but then dropped it way back in https://github.com/ipld/go-ipld-prime/commit/763174ea9f27140eaf7d9a7becd85b155124944f with... not as much reasoning in the commit message as I usually expect from myself.  The parent commit https://github.com/ipld/go-ipld-prime/commit/4bd496a92c54e3891aa371c1a5278c9d4d2f6e8f is where union guts were introduced for the very first time at all (!), so this is pretty old stuff in general, and maybe worth a review at this point.  It looks like map key questions were a part of it, but choosing to have a map at the type level with arguable redundancy means that keys are answered, so, uh, I'm not sure my past self's logic was very consistent here.

So you raise a good point, and I'm not sure which way we should hew and which of these two should change to remove the discrepancy.  More thought needed.

---

**parens**: cool, let's do it then.

I also think your idea to shorten them down to one value in this case sounds completely fine.  :+1:   (It probably introduces another small mode in the syntax parser?  But it's not ambiguous or anything.  Go for it.)

---

**quotes around the value**: I'd actually officially like to punt this choice (both here and how `implicit` does it) to you :)

rvagg:
> discrepancy with unions

Seems like we need to resolve this now then: (1) adopt different approaches for unions and enums, (2) use the current unions approach for both, or (3) use your proposed approach here for both.

My guess is that you switched because of the duplication and verbosity in JSON; combined with the fact that the different types of unions are almost entirely different beasts so their overlapping needs are minimal. Enums are a bit different though, we're not really proposing entirely different things, just that there's a mapping to named programatic things and in one form the serial form is a string, which may or may not be the same as the name, and in another form it's an int.

> quotes around the value

Here's one key thing about that question -- if we go for the shorthand version then we make it easier for the parser if we choose to extend the options in parens for enums. If we go unquoted then we have a hard time differentiating. Consider a string representation mapping unquoted: `| Foo (f)`, then we introduce an option like `| Foo (f alternate foo)` where `alternate` lets you specify another string form in serialization that would also be a `Foo` (so a "f" or "foo" = `Foo`). Now introduce an enum that serializes to "alternate": `| Alt (alternate)` - the signal to the parser that this is a serialized string is that there's no second token in the parens, what about `| Alt (alternate alternate)`? That should be valid and could be parsed, but maybe it was a user error? The lack of clarity is a problem here IMO. Which is why I think I'm in favour of quoting, regardless of int or strings: `| Foo ("f" alternate foo)`, `| Yes ("1")`.

Quoting non-strings does irk me but it's a nice signal to the parser, and possibly the author and reader, that we're breaking out of schema-language into serialized forms.

schemas/schema-schema.ipldsch
rvagg: capitalised first character

warpfork:
Still not 100% confident on the best and most consistent way through on the discrepany with unions.  You've summed it up well.  I'd lean towards saying (3) long run, but we might as well do (1) at present as that's the smaller diff, and if we learn something useful from a brief period running with both strategies, that's great and we can change our minds.  But if you have a strong opinion for the other choices, I'd support it.

Complete thumbs up to your comments on the quoting and DSL format.  `| Yes ("1")` reads right to me.

rvagg:
OK, updated, will merge this if there are no comments in the next day or two.

As per test fixtures in https://github.com/ipld/js-ipld-schema/pull/16 (which is updated according to the latest changes here):

```
  type SimpleEnum enum {
    | Foo ("0")
    | Bar ("1")
    | Baz ("100")
  } representation int
```

yields

```json
  {
    "types": {
      "SimpleEnum": {
        "kind": "enum",
        "members": {
          "Foo": null,
          "Bar": null,
          "Baz": null
        },
        "representation": {
          "int": {
            "Foo": 0,
            "Bar": 1,
            "Baz": 100
          }
        }
      }
    }
  }
```

and

```
  type SimpleEnum enum {
    | Foo
    | Bar
    | Baz
  }
  type SimpleEnumWithValues enum {
    | Foo ("f")
    | Bar
    | Baz ("b")
  }
```

yields

```json
  {
    "types": {
      "SimpleEnum": {
        "kind": "enum",
        "members": {
          "Foo": null,
          "Bar": null,
          "Baz": null
        },
        "representation": {
          "string": {}
        }
      },
      "SimpleEnumWithValues": {
        "kind": "enum",
        "members": {
          "Foo": null,
          "Bar": null,
          "Baz": null
        },
        "representation": {
          "string": {
            "Foo": "f",
            "Baz": "b"
          }
        }
      }
    }
  }
```

rvagg:
oh, and I'm keeping `"representation": { "string": {} }` in there for the default, like we do for struct-map.

schemas/representations.md
```diff
@@ -308,3 +319,64 @@ type Bls12_381Signature bytes
...
+
+By default, an enum is simply represented as a string in the data model. An
+`enum` in the schema simply defines the list of possible strings that could be
+used in that position.
+
+**Example**
+
+```ipldsch
+type Status enum {
+	| NOPE
```
warpfork, line 333: Can we switch these out to TitleCase rather than SCREAMCASE, same as your latest examples in https://github.com/ipld/specs/pull/196#issuecomment-542009802?

warpfork (approved):
This looks super correct to me :+1:

rvagg:
merged

PR #197, opened 2019-09-20
schemas: define a "copy" type

rvagg:
First go at defining a `copy` type. This might naively be thought of as an **alias** and it's what Filecoin is trying to do in https://github.com/filecoin-project/specs/blob/master/network-protocols.md:

```
type StorageDealResponse union {
    | UnknownParams
    | RejectedParams
    | AcceptedParams
    | StartedParams
    | StagedParams
...
} representation keyed
...
type RejectedParams struct {
    message optional String
    proposal &SignedStorageDealProposal
}
type AcceptedParams RejectedParams
type FailedParams RejectedParams
type StagedParams RejectedParams
...
```

So you get the same structure in memory but a different _thing_ (handle?) to assert against and you get more descriptive power in in your schemas as well as reducing duplication.

**BUT** this is not intended to be implemented as an alias, it's a copy in the macro sense. We are taking one type definition, changing the name and creating a new one with the same shape and representation. It's is described here as a short-hand for repeated definitions.

## JSON form

I've introduced a new schema kind of `"copy"` to store this in the JSON form. We want to retain the fact that it's a `copy` and only perform the copying when we implement a thing (codegen, validation, whatever)—we get to a type, find it's a `"copy"` and redirect to the `"fromType"` to get the definition body but with the original name. It would look like this:

```
  "Foo": {
    "kind": "copy",
    "fromType": "Bar"
  }
```

## DSL syntax

Regarding syntax, I think @warpfork originally proposed `type Foo = Bar`, and I'm sympathetic to the idea of introducing a delimiter to make it clear what's going on and avoid confusion with the other schema kinds that can go in that position with an empty definition (`bytes`, `string`, etc.). Here's some variations to consider:

* `type Foo Bar`
* `type Foo = Bar`
* `type Foo copy Bar`
* `type Foo <= Bar`

### DIFF
--- a/schemas/schema-kinds.md
+++ b/schemas/schema-kinds.md
@@ -2,8 +2,13 @@
 
 * [Extending the IPLD Data Model](#extending-the-ipld-data-model)
   * [Data Model Kinds](#data-model-kinds)
-  * [Recursive Kinds](#recursive-kinds)
-  * [Meta Kinds](#meta-kinds)
+  * [Schema Kinds](#schema-kinds)
+    * [List](#list)
+    * [Map](#map)
+    * [Union](#union)
+    * [Struct](#struct)
+    * [Enum](#enum)
+    * [Copy](#copy)
 * [Value Type Modifiers](#value-type-modifiers)
   * [Nullable Values](#nullable-values)
   * [Optional Fields](#optional-fields)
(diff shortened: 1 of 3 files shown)


vmx:
`type Foo = Bar` sounds good to me as it indicates that it is more than just an alias, which I would imply if I would see `type Foo Bar`. Though I'm not as deep into the schemas as the rest of the team is, so I leave the decision to others.

mikeal:
should the parsed form be `{ "kind": "copy", "fromType": "Bar" }` or should it just be a literal copy of the parsed form of Bar?

rvagg:
If we do the copy when parsing we can't go back the other way. A `to-schema` operation would spit out full types with no copies. Leaving the copy operation for later means we retain full info about what the original schema was.

rvagg:
I'm partial to `type Foo <= Bar` but I don't have strong opinions on any of the syntax options here really.

schemas/schema-kinds.md
```diff
@@ -1,26 +1,35 @@
...
+   `envelope` and `inline` representations, and varying data model kinds for
+   `kinded` unions, as described by [representations.md](representations.md).
+ - **Struct** - represented as a **map** in the data model by default but may be
+   used to describe **string** and **list** encodings, as described by
+   [representations.md](representations.md).
+ - **Enum** - represented as either a **string** or **int** in the data model,
+   as described by [representations.md](representations.md).
+ - **Copy** - a special kind that indicates that a type should be implemented
+   and encoded the same as another type but with an alternate name. This is not
+   the same as an alias, but a short-hand to avoid defining multiple types of
```
warpfork: I might drop the "not the same as an alias" part, because there's no definition of "alias" in scope to be strongly disambiguating against, and just proceed with "This is a short-hand to ..."

(It's a point that might be worth noting and exploring elsewhere but seems a stretch for a short summary like this.)

Like the rest.

schemas/schema-kinds.md
```diff
@@ -1,26 +1,35 @@
...
+ - **Link**
+ - **Union** - represented as a **map** in the data model for `keyed`,
+   `envelope` and `inline` representations, and varying data model kinds for
+   `kinded` unions, as described by [representations.md](representations.md).
+ - **Struct** - represented as a **map** in the data model by default but may be
+   used to describe **string** and **list** encodings, as described by
+   [representations.md](representations.md).
+ - **Enum** - represented as either a **string** or **int** in the data model,
+   as described by [representations.md](representations.md).
+ - **Copy** - a special kind that indicates that a type should be implemented
```
warpfork: Hmm.  Maybe `**copy**` should go in a separate list?

It shows up in the schema DSL (if not by name) and in the AST, but it's *not* going to show up in the [`schema.Kind` enum](https://github.com/ipld/go-ipld-prime/blob/53e4ebbe7bf632c9fe73baefee46135e923df8de/schema/kind.go#L13-L28) in the types that people actually use, which makes it a somewhat distinctive beast.

schemas/schema-schema.ipldsch
```diff
@@ -623,3 +625,16 @@ type StructRepresentation_ListPairs struct {}
 type TypeEnum struct {
 	members {String:Null}
 }
+
+## TypeCopy describes a special "copy" unit that indicates that a type name
+## should copy the type descriptor of another type. While similar to an
+## "alias", a TypeCopy doesn't redirect a name to another type, it copies the
```
warpfork: s/similar to/similar to the concept of/

warpfork (approved):
Great breakdown of the subject in the PR text.

2c, I still vote for the `type Foo = Bar` syntax.  Fewer sigils is better than more sigils.  But one sigil is better than none, because it should look visually distinctive from `type Foo keyword` patterns (and more so than a capital letter accomplishes).

I left some docs comments, but broadly agree with this overall.

rvagg:
@warpfork see latest commit, to get "copy" to be a distinctive kind I extended the list into 3 distinct categories with their own headings. Then I added a TOC because it's getting larger and reorganised the hierarchy a bit. It should address all of your comments. I've also added a note about `=` in schema-schema too.

rvagg:
test fixture over in https://github.com/ipld/js-ipld-schema/pull/17/files#diff-09529cb659091175d377ded071dd17f8 has examples of the current proposal for consideration, in DSL and JSON form.

schemas/schema-kinds.md
```diff
@@ -1,29 +1,65 @@
...
+  - **Boolean**
+  - **Integer**
+  - **Float**
+  - **String**
+  - **Bytes**
+  - **List**
+  - **Map**
+  - **Link**
+
+### Recursive Kinds
```
warpfork: I love the docs and reorg into sections you've done here, but this section name is wrong.  Map and List are recursive kinds, and they're in the Data Model Kinds set; therefore this heading can't be "recursive".

I can think of a couple choices for what else we could title this, and maybe you can brainstorm others as well:

- "Type-level kinds"
- "Perceived kinds" ([ancient reference](https://gist.github.com/warpfork/6df17e791936d1f9b0d5e5483678c8bf#file-gistfile1-txt-L3) -- not attached to it tho)
- ...?

We also might want to decide on if we orient most of our phrasing around the two *sets*... or the DM set plus more.  I tend to favor the former approach (naming the two sets), and then maybe don't spent so much words on the naming the ones that differ; just briefly pointing them out might suffice.

warpfork: Though it touches on things beyond the original scope of this PR, I'll throw on for consideration: maps and lists also behave slightly different in the schema layer, because they have to have specified key and value types now.  (One *can* say `[any]` and `{String:any}`, but one now has to... well, say that.)

So this might be an indication that the two-sets style of description might be the clearer route: it gives a place where this shift in map and list behavior can also be briefly discussed.

rvagg: what about "Abstract Kinds"?

rvagg: I don't quite understand what you mean by the last paragraph, is there something concrete in this PR I could change to alleviate that concern?

warpfork: "Type-level kinds" -> would describe all of {int,string,...,map,list,union,enum,struct}.
"Perceived kinds" -> would describe just {union,enum,struct} -- the added ones.
(I suspect "abstract kinds" would lean towards the latter mode as well.)

I tend to think the former route might be preferable, because the latter route results in docs *everywhere else* having to say the mouthful of "the data model kinds plus the perceived kinds".

rvagg: OK, how about this:

-----

**Scalar Kinds**

... something about how this is the data model minus list and map, list them here ...

**Recursive Kinds**

... something like: these kinds build upon the scalars in various ways, maps and lists in the schema layer are not entirely freeform and must be composed of specific types, as specified in a schema ... structs, unions and enums are abstract kinds that describe layouts comprised of data model kinds, depending on the representations specified ...

rvagg: and **Meta Kinds** since you want that to be a separate thing, or do you have a better suggestion for how to handle `copy`?

rvagg: > the former route might be preferable

you're losing me, which route? what headings do you want?

warpfork: Okay, lemme back myself up a bit.

Usually when I visualize these things, I visualize them as a venn diagram.  The Data Model Kinds are a subset of the Type Level Kinds.  So we've got a circle inside a bigger circle.  And in my mental model, those are the two labels I use -- I label each of the two big circles.  I usually _don't_ label the circle-with-a-cutout shape (which is where struct,union,enum live) but if I did, I'd probably call it the "perceived kinds".

The reason I mentally label it this way is that I've found I end up needing to refer to "type-level kinds" in other documentation or code comments at a rate of about 10x more often than I need refer to "perceived kinds".

(And now as of this PR, the venn diagram isn't just two circles, it's now bumped up to three: We've got another wee small circle on the side of the diagram now for the "meta kinds" for this Copy feature.)

So, it seems to me there's two different ways to write this documentation: naming the ~two~ three big circles; or naming the circle, the circle-with-cutout, and the other wee circle.  (This is the 'route' I was referring to).

In either case, the "meta kinds" header is the same.  And in either case, the reference to the data model kinds is there, so that heading is probably the same.  The only heading that changes based on this choice is the middle one -- because that's the thing that's either a circle, or a circle-with-cutout, in the venn diagram.

So we've got two choices: what to call this header, and which approach to the venn diagram we're taking.  But they're actually the same choice.  Does that savvy?

---

I was considering advocating for the name-the-circles route and thus avoiding having to name the circle-with-cutout, because of the aforementioned ratio in usage rate.  (In that route, we'd probably favor a name more like "type-level kinds" (if not that specific string) for this section -- and then the first sentence in this section would be "The type-level kinds include all of the already familiar data-model kinds, plus three new kinds:" and then resume with the rest of the paragraphs below.  See how the "three new kinds" never got a group title themselves?)

But now that I've backed myself up a bit, I see how describing {circle, circle-with-cutout, circle} is probably more obvious when writing it as a bulleted list.  I'm just a bit sad it means having to name the circle-with-cutout region at all.

This just turned into a bunch of braindump, sorry.  Maybe the one good takeaway is that we should actually consider a venn diagram png somewhere near here!

rvagg: I'm thinking that maybe this doc would be better by:

1. Leading with a listing the data model kinds
2. Then moving into a full list of type kinds with:
  a. An explanation of the new restrictions placed on list and map (although the restrictions are hard to be clear about because unions let you have mixed kinds within lists and map values).
  b. How the new forms, struct, enum and union, build compose from the other kinds and how they can be described as "perceived".
  c. Mention "copy" as a meta kind.

vmx (approved):
I already like the changes as they are, but feel free to polish it more :)

rvagg:
heyhey, see my latest _fixup!_ commit for a major rework of the top section. It might be easier to just read the doc straight: https://github.com/ipld/specs/blob/rvagg/schemas-aliases/schemas/schema-kinds.md (just the top section, before "Value Type Modifiers"). @vmx @warpfork

warpfork:
Absolutely freakin stellar.  I love it.

*crowd chants Merge! Merge! Merge!*

rvagg:
phew

PR #198, opened 2019-09-30
Add exploration report about using Serde for Rust IPLD

vmx:
These are my thoughts while trying to implement DagCbor for Rust.
The implementation prototype can be found at https://github.com/vmx/rust-ipld-dag-cbor

### DIFF
--- a/design/history/exploration-reports/2019.09-rust-serde.md
+++ b/design/history/exploration-reports/2019.09-rust-serde.md
@@ -1,0 +1,58 @@
+Using Serde for Rust IPLD
+=========================
+
+**Author**: Volker Mische ([@vmx])
+
+It sounds like a great idea using [Serde] as basis for Rust IPLD. It should be simply to put the [IPLD Data Model] on top of the [Serde Data Model]. So I was looking into implementing DagCbor on top of [serde_cbor].
+
+Missing CBOR tags support in serde_cbor
+---------------------------------------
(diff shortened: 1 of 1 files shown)


rvagg (approved):
Is this now a back-burner thing? something you'll try to get to eventually? or waiting for imminent action per the last paragraph?
Sounds like this could be a pretty strong blocker if no demonstrable progress can be made. I'd like to understand a bit more about the implications of operating outside of Serde, I know there are other serialisation formats that are bypassing Serde entirely, is that feasable?

vmx:
@rvagg I'm going to try to get tag support into Serde immediately.

> I'd like to understand a bit more about the implications of operating outside of Serde, I know there are other serialisation formats that are bypassing Serde entirely, is that feasable?

From my current understanding and looking at [msg-pack-rust](https://github.com/3Hren/msgpack-rust/), I would architecture a serialization format as a low level encoding/decoding of the primitives without a dependency on Serde and then another library which is using Serde on top of that.

If it's done like that, you could swap out Serde with something else, in our case with the IPLD Data Model based serializer.

The nice things about using Serde directly is that we just would need to write the part that converts from the Serde Data Model into our Data Model, without the need to care about the low level serialization of formats.

warpfork (approved):

mikeal (approved):

PR #199, opened 2019-10-01
Add information about migration of schemas

vmx:
This document was copied from
https://github.com/ipld/go-ipld-prime/blob/53e4ebbe7bf632c9fe73baefee46135e923df8de/doc/schema.md

I've made two commits out of it, so that my changes are easier to review.

@warpfork: The last section "Actually Migrating!" is kind of Go implementation specific (talking about `Node`s. Can you please have a look and perhaps reformulate it?

### DIFF
--- a/schemas/migration.md
+++ b/schemas/migration.md
@@ -1,0 +1,165 @@
+Schemas and Migration
+---------------------
+
+Fundamental to our approach to schemas is an understanding:
+
+> Data Never Changes.  Only our interpretation varies.
+
+Data can be created under one schema, and interpreted later under another.
+Data may predate or be created without any kind of schema at all.
+All of this needs to be fine.
+
+Moreover, before talking about migration, it's important to note that we
+don't allow the comforting, easy notion that migration is a one-way process,
+or can be carried out atomically at one magically instantaneous point in time.
+Because data is immutable, and producing updated versions of it doesn't make
+the older version of the data go away, migration is less a thing that you do;
+and more a state of mind.  Migration has to be seamless at any time.
+
+Migration comes in two parts:
+
+1. Understand what data we have;
+2. and having a process to map it into the format of data we want.
+
+We'll spend a few sections on part 1, and then get on to part 2.
+
+### Using Schema Match checking as Version Detection
+
+We don't include any built-in/blessed concepts of versioning in IPLD Schemas.
+It's not necessary: we have rich primitives which can be used to build
+either explicit versioning or version detection, at your option.
+
+Since it's easy to check if a schema fits over a piece of data, it's
+easy to simply probe a series of schemas until finding one that fits.
+Therefore, any constraint a schema makes has the potential to be used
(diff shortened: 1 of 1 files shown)


schemas/migration.md
```diff
@@ -0,0 +1,165 @@
+IPLD Schemas and Migration
```
rvagg: We should pull the heading levels up by one in here, this top one should be `#` or `=====` and the subsequent ones should be `##`.

schemas/migration.md
```diff
@@ -0,0 +1,165 @@
...
+and more a state of mind.  Migration has to be seamless at any time.
+
+Migration comes in two parts:
+
+1. Understand what data we have;
+2. and having a process to map it into the format of data we want.
+
+We'll spend a few sections on part 1, and then get on to part 2.
+
+### Using Schema Match checking as Version Detection
```
rvagg: These headings have inconsistent casing.

My preference (this is not a strong opinion, I mainly want consistency) is for title case for 1st and 2nd level headings where the titles are short enough but this is a long title. The titles could be shortened though, this one could just be "Version Detection" and the next one "Versioning Theory".

schemas/migration.md
```diff
@@ -0,0 +1,165 @@
...
+explicitly linked is certainly one way to address the need to
+"understand what data we have", remember that the definition of migration has a
+second half: "having a process to map data into the format we want".
+
+Unless it just so *happens* that this exact schema is the one you want, and have
+already build your application logic against, etc... an explicitly linked schema
+doesn't necessarily provide more value in terms of migration than any of the
+other forms of versioning; it's essentially the same as using explicit labels.
+
+### Actually Migrating!
```
rvagg: this and the next sentence is too casual, suggest removal of `!` and the entire next sentence.

rvagg:
I have mixed feelings about this doc, mostly I think it's positive to have it here, it's an important piece in the "keep schemas simple" (and "no dependent types") arguments, but on the other hand it's much more verbose than anything else we have on schemas so far and parts of it are a little casual for the specs repo. I'm OK with merging it with the modifications I suggest but it probably needs to be cleaned up and brought into better alignment at some point in the future.

vmx:
I went for sentence case as title case is pretty alien to me as a German native speaker.

rvagg (approved):
ok, whatever, I don't mind too much as long as it's consistent.
I know Eric prefers title case in most places, I'd prefer it _at least_ in the actual title, but I don't care enough to make this more complicated.

schemas/migration.md
warpfork: fwiw, when I first skrived it, "Schema Match" and "Version Detection" got this-is-an-Important-Noun caps... and then "Using" got first-word caps, and so "checking" was just the odd one out.

But I did that without much in the way of supporting context, and leaning on caps that way is not the greatest writing, so I agree with the fixes :thumbs_up:

PR #200, opened 2019-10-01
Selectors examples

vmx:
This revives the selectors examples.

Someone please review the YAML parts in detail, I haven't used schemas much and I'm not 100% sure I got things right.

### DIFF
--- a/selectors/example-selectors.md
+++ b/selectors/example-selectors.md
@@ -19,7 +19,7 @@ For human convenience, we will also pretend there are a few modifications
 to the Selector schema:
 
 - we will disregard all 'rename' shortenings specified by the schema;
-- and we will use type names (snakeCase'd) for union discriminators,
+- and we will use type names for union discriminators,
   instead of the shorter keys specified in the schema for those unions.
 
 Therefore, note that the *real* serialized selectors will be significantly terser!
@@ -53,15 +53,21 @@ Example data (as JSON):
 A Selector to extract the "year" data could look like this:
 
 ```yaml
-selectFields:
-  "characters":
-    selectFields:
-      "kathryn-janeway":
-        selectFields:
-          "birthday":
-            selectFields:
-              "year":
-                matcher: {}
+Selector:
+  ExploreFields:
+    fields:
+      "characters":
+        ExploreFields:
+          fields:
+            "kathryn-janeway":
+              ExploreFields:
+                fields:
+                  "birthday":
+                    ExploreFields:
+                      fields:
+                        "year":
+                          Matcher:
+                            {}
 ```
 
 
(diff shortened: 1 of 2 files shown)


mikeal (approved):

selectors/example-selectors.md
```diff
@@ -0,0 +1,218 @@
...
+#### A quick note about formatting
+
+We will typically show some sample data as JSON encoded documents,
+and show the Selectors as YAML documents.
+**Remember,** any IPLD document can be encoded in any multicodec;
+these examples are JSON for convenience and by convention, but all
+the examples would work the same even if they were in other formats.
+Similarly, **Selectors themselves are just IPLD documents**, so while
+we've written yaml forms here, they could just as well be e.g. JSON or CBOR.
+When we write yaml, we'll also use quoted strings for user-supplied strings,
```
rvagg: probably should s/yaml/YAML`

selectors/example-selectors.md
```diff
@@ -0,0 +1,218 @@
...
+          fields:
+            "kathryn-janeway":
+              ExploreFields:
+                fields:
+                  "birthday":
+                    ExploreFields:
+                      fields:
+                        "year":
+                          Matcher:
+                            {}
```
rvagg, line 70: I guess this is correct, I'm just not sure that YAML is expressive enough to show what's going on here, although that could be that my eyes aren't as comfortable with YAML as some other formats.

selectors/example-selectors.md
```diff
@@ -0,0 +1,218 @@
...
+    maxDepth: 5
+    sequence:
+      ExploreFields:
+        fields:
+          "parent":
+            ExploreRecursiveEdge
+```
+
+This will traverse the same set of nodes as the previous example -- however,
+it has has a *slightly* different effect!
```
rvagg: `!` probably a bit too casual here for a specs doc

rvagg (approved):
+1 sans one nit and some discomfort with the utility of YAML for the examples

selectors/selectors.md
```diff
@@ -70,9 +70,9 @@ type ExploreAll struct {
 ## traversing on typed/schema nodes) and applies a next selector to the
 ## reached nodes.
 ##
-## Note that a concept of "ExplorePath" (e.g. "foo/bar/baz") can be represented
-## as a set of three nexted ExploreFields selectors, each specifying one field.
-## (For this reason, we don't have a special "ExplorePath" feature; use this.)
+## Note that a concept of exploring a whole path (e.g. "foo/bar/baz") can be
+## represented as a set of three nexted ExploreFields selectors, each
+## specifying one field.
```
warpfork, line 75: Excellent rephrase :+1:

warpfork (approved):
Some of the latter examples start getting a little bumpy to read without having example JSON data that they're fitting over that lives within this same file.

But overall, looks about right to me.

warpfork:
(Someday it'd be nice to make these snippets into something we can parse and put through some kind of functional tests too, but it would be a bit of a stretch right now.  Someday, after we get a feature for doing "take this schema, but ignore all the 'rename' directives, then parse the data against that", this'll become a lot easier.  At the moment, it would be problematic to plug these directly into functional tests largely because of the concessions we made to readability; with that feature, we could have our cake and eat it too.)

(Mostly noting this as an example of user story, so I can link to it as a justification for that feature in the (possibly somewhat distant) future :))

selectors/example-selectors.md
warpfork: It ain't pretty, but yeah, it's correct.  It's just a very right-leaning tree.  Hard to make those pretty.

(Fwiw, one of the oldest iterations of this whole Selectors thing actually used lists because we were so eager to avoid the right-leaning tree it seemed like a viable option.  It was not: more complex queries that we instantly had users toss at us were inexpressible in that notation.)

Probably the prettiest format for something like this would take inspiration from XML: the keyed unions here (`ExploreFields`, `Matcher` in this example) are an awfully lot like the tag in XML, and the map inside those is an awfully lot like attributes.  (Or, are the inner maps like nested keys?  Ahha.  This is why we don't use XML... ;)  Fun to think about other syntaxes that would have better visual presentation though!)

vmx: In my editor (vim) it looks much nicer, as the things in quotes are highlighted differently from the ones without quotes.
image

vmx:
Fixed the minor things, hence merging.

Issue #55, opened 2018-03-10
Spec out DagPB path resolution


Stebalien:
Currently, we're playing it fast-and-loose with DagPB path resolution. For context, the *structure* of DagPB IPLD objects is currently *defined* to be:

```js
{
  "Data": "binary data",
  "Links": [
      {
        "Name": "thing", // may not be unique, may be omitted
        "Tsize": 1234, // may be omitted
        "Hash": "QmId"
      }
  ]
}
```

Given this structure, the *correct* path to `QmId` would be `/Links/0/Hash`. However, this isn't very usable.

* js-ipfs actually supports this pathing scheme. However, it *also* supports pathing by name (`/Links/thing/Hash`). While nice and *usable*, this is technically transforming the object.

* go-ipfs uses the pathing scheme `/thing`. That is, it treats the object as if it were a single map of names to CIDs (for the purposes of pathing). This is obviously problematic from an IPLD standpoint as many of the fields aren't addressable.

So, we need a scheme that's both consistent between implementations and consistent with other IPLD formats.

Given that we're basically the only real consumer of this format, I believe we have a *bit* of flexibility.

---

By the way, @diasdavid, your code organization is :100:. <sup>(despite the fact that you're using JavaScript)</sup>

Stebalien:
# Proposal: Simplicity over UX

Given that others are, hopefully at least, *not* using this format, we could just say that all paths must use link *indices*. We'll need to write a helper function to resolve named links.

Stebalien:
# Proposal: Named xor Unnamed

If we require that all links in an object either be named or unnamed, we can redefine the structure to either have a map of link names to links *or* an array of unnamed links. The question is, do we ever *mix* named links with unnamed links?

rvagg:
Re `/Links/thing/Hash`, I don't believe this is possible in the current js-ipld-dag-pb. I supports `/thing` and `/Links/0/Hash` style. (I can't find a historical version that shows clearly that `/Links/thing/Hash` was actually supported tbh).

So with them both supporting `/thing`, an easy fix might to be to define that, plus the `/Links/0/Hash` style as _the way_, then implement `/Links/0/Hash` in he Go implementation, then they'd be unified and there will be peace in the world, no?

vmx:
> So with them both supporting `/thing`, an easy fix might to be to define that, plus the `/Links/0/Hash` style as _the way_, then implement `/Links/0/Hash` in he Go implementation, then they'd be unified and there will be peace in the world, no?

[...]

PR #201, opened 2019-10-02
Add DagPB spec

vmx:
This PR combines #114 and #115 and addresses all code review comments from there.

I then changed the section about pathing to describe the current state in the Go and JS implementations. As this is a descriptive spec, I find this a sensible approach.

Of course it would be desirable to harmonize how pathing works between implementations, but that hasn't happened for a long time and I don't see this happening soon. Hence I find having the current state document is better than keeping discussing it without taking action.

### DIFF
--- a/block-layer/codecs/dag-pb.md
+++ b/block-layer/codecs/dag-pb.md
@@ -1,10 +1,12 @@
 # DagPB Spec
 
+**Status: Descriptive - Draft**
+
 DagPB does not support the full ["IPLD Data Model."](../../data-model-layer/data-model.md)
 
 ## Format
 
-The DagPB IPLD format is a legacy format implemented with a single protobuf.
+The DagPB IPLD format is a format implemented with a single protobuf.
 
 ```protobuf
 // An IPFS MerkleDAG Link
@@ -32,11 +34,13 @@ message PBNode {
 ```
 
 The objects link names are specified in the 'Name' field of the PBLink object.
-All link names in an object must either be blank or unique within the object.
+All link names in an object must either be omitted or unique within the object.
 
 ## Pathing
 
-There is some overlap between the Go and JavaScript implementation of DagPB. Both support pathing with link names: `/<name1>/<name2>/…`.
+The pathing is currently different between implementations. Please see [issue #55] for more information about the harmonization effort. This section describes the current implementations as of September 2019.
+
+The Go and JavaScript implementation both support pathing with link names: `/<name1>/<name2>/…`.
 
 In Go, this is the only way, which implies that is is impossible to path through nodes that don't name their links. Also neither the Data section nor the Links section/metadata are accessible through paths.
 
@@ -57,3 +61,5 @@ Canonical DagPB must:
 Historical Note: The ordering (Links then Data) of the PBNode message is due to
 a bug in the initial protobuf encoder that was used in the first implementation
 of ipfs. Take care to maintain this ordering for full compatibility.
+
+[issue #55]: https://github.com/ipld/specs/issues/55
--- a/block-layer/codecs/dag-pb.md
+++ b/block-layer/codecs/dag-pb.md
@@ -1,6 +1,6 @@
 # DagPB Spec
 
-DagPB does not support the full ["IPLD Data Model v1."](../IPLD-Data-Model-v1.md)
+DagPB does not support the full ["IPLD Data Model."](../../data-model-layer/data-model.md)
 
 ## Format
 
--- a/Codecs/DAG-PB.md
+++ b/Codecs/DAG-PB.md
@@ -32,7 +32,7 @@ message PBNode {
 ```
 
 The objects link names are specified in the 'Name' field of the PBLink object.
-All links in an object must either be blank or unique within the object.
+All link names in an object must either be blank or unique within the object.
 
 ## Pathing
 
--- a/Codecs/DAG-PB.md
+++ b/Codecs/DAG-PB.md
@@ -10,10 +10,10 @@ The DAG-PB IPLD format is a legacy format implemented with a single protobuf.
 // An IPFS MerkleDAG Link
 message PBLink {
 
-  // multihash of the target object
+  // binary CID (with no multibase prefix) of the target object
   optional bytes Hash = 1;
 
-  // utf string name. should be unique per object
+  // UTF-8 string name.
   optional string Name = 2;
 
   // cumulative size of target object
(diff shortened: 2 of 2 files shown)


block-layer/codecs/dag-pb.md
```diff
@@ -0,0 +1,59 @@
...
+  optional bytes Data = 1;
+}
+```
+
+The objects link names are specified in the 'Name' field of the PBLink object.
+All link names in an object must either be blank or unique within the object.
+
+## Pathing
+
+There is some overlap between the Go and JavaScript implementation of DagPB. Both support pathing with link names: `/<name1>/<name2>/…`.
```
Stebalien: Can we note (agree) that this is a bug?

Stebalien (approved):

mikeal (approved):

block-layer/codecs/dag-pb.md
```diff
@@ -0,0 +1,59 @@
+# DagPB Spec
+
```
rvagg, line 2: I think this needs to be inserted here: `**Status: Descriptive - Final**`

block-layer/codecs/dag-pb.md
rvagg: Probably worth linking to https://github.com/ipld/specs/issues/55 as a "discussion of this issue and how it might be resolved".

rvagg (approved):
:thumbsup: if you at least insert the spec status line at the top, will leave you to figure out how to handle the JS/Go discussion bit.

block-layer/codecs/dag-pb.md
```diff
@@ -0,0 +1,59 @@
...
+## Format
+
+The DagPB IPLD format is a legacy format implemented with a single protobuf.
+
+```protobuf
+// An IPFS MerkleDAG Link
+message PBLink {
+
+  // binary CID (with no multibase prefix) of the target object
```
achingbrain, line 15: Is this correct?  You can have v1 CIDs as DAGLinks which can contain a multibase prefix.

block-layer/codecs/dag-pb.md
```diff
@@ -0,0 +1,59 @@
...
+  // refs to other objects
+  repeated PBLink Links = 2;
+
+  // opaque user data
+  optional bytes Data = 1;
+}
+```
+
+The objects link names are specified in the 'Name' field of the PBLink object.
+All link names in an object must either be blank or unique within the object.
```
achingbrain: Since the `Name` field is optional, maybe say it must either be omitted or unique within the object?

block-layer/codecs/dag-pb.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Pathing
```
achingbrain, line 39: I don't think discussing differences between implementations makes for a good spec.  If you were implementing this, what would you implement here?

Maybe just standardise on the JS version since it lets you navigate non-named links and the go version doesn't?

block-layer/codecs/dag-pb.md
```diff
@@ -0,0 +1,59 @@
+# DagPB Spec
+
+DagPB does not support the full ["IPLD Data Model."](../../data-model-layer/data-model.md)
+
+## Format
+
+The DagPB IPLD format is a legacy format implemented with a single protobuf.
```
achingbrain: I don't think it's fair to call this legacy (yet) as it's what pretty much all IPFS data is encoded with and until `UnixFSv2` arrives there's no way to use anything else.

block-layer/codecs/dag-pb.md
vmx: @Stebalien We can totally agree on that. Let's discuss it at #55 (and I'll add a link to it in here).

block-layer/codecs/dag-pb.md
vmx: As per https://github.com/ipld/specs/pull/201/files/d7f528343fa4b051a7ab1042013a4830a8f090a5#r330630459, I'll add a link to the bug. But I think it's better to know how things are currently implemented, rather than not having this documented at all.

vmx:
A new version is up. I made the spec "Descriptive: **Draft**" as we still have issue #55 open (in a final spec I would expect that nothing will change). For those who have already approved that spec, as thumbs up on this comment would be nice. If you can't be bothered, I'll merge it after 2 working days after @achingbrain approved it.

block-layer/codecs/dag-pb.md
vmx: I haven't addressed this as I also don't know (@Stebalien probably does).

rvagg: https://github.com/ipld/js-ipld-dag-pb/blob/c4b83cbb69821ca6da12190105b0e78d27688630/src/serialize.js#L23
https://github.com/ipfs/go-merkledag/blob/2c1a891f6f1b2bbb6442b9da62e5ece591fbdad9/pb/merkledag.pb.go#L425

these both looks like raw CID byte arrays

rvagg (approved):
still lgtm, but that "with no multibase prefix" thing seems wrong, would be good to get an authoritative answer on that, the code looks to me like there's no trimming of the raw CID byte array

block-layer/codecs/dag-pb.md
vmx: CIDs don't contain a multibase prefix.

I created a script to verify this (yes, I am using a sledgehammer to crack a nut, but it gave me the chance to finally use the new js-ipld-dag API as well as playing a bit with go-ipld). The output is:

```console
$ node --experimental-modules combined.js
(node:985) ExperimentalWarning: The ESM module loader is experimental.
actual hash:   89b1d9957f247ee0fbaa60726be4e138fe9983c5919dec1a1b8a43e7ed6509c8
cidv0:     122089b1d9957f247ee0fbaa60726be4e138fe9983c5919dec1a1b8a43e7ed6509c8
cidv1: 0170122089b1d9957f247ee0fbaa60726be4e138fe9983c5919dec1a1b8a43e7ed6509c8
```

And here's the script in case someone wants to run it locally:

```js
import dagPb from 'ipld-dag-pb'
const { DAGLink, DAGNode} = dagPb
import Pbf from 'pbf'

// Creating the data

const node = new DAGNode('some data')

// base58btc - cidv0 - dag-pb - sha2-256-256-89b1d9957f247ee0fbaa60726be4e138fe9983c5919dec1a1b8a43e7ed6509c8
const cidv0 = 'QmXc9raDM1M5G5fpBnVyQ71vR4gbnskwnB9iMEzBuLgvoZ'
const linkv0 = new DAGLink('cidv0link', 10, cidv0)
node.addLink(linkv0)

// base32 - cidv1 - dag-pb - sha2-256-256-89b1d9957f247ee0fbaa60726be4e138fe9983c5919dec1a1b8a43e7ed6509c8
const cidv1 = 'bafybeiejwhmzk7zep3qpxktaojv6jyjy72myhrmrtxwbug4kipt62zijza'
const linkv1 = new DAGLink('cidv1link', 11, cidv1)
node.addLink(linkv1)

//console.log(node)
const serialized = node.serialize()
//console.log(serialized)


// Reading the data (generated by pbf)

// PBLink ========================================

const PBLink = {};

PBLink.read = function (pbf, end) {
    return pbf.readFields(PBLink._readField, {Hash: null, Name: "", Tsize: 0}, end);
};
PBLink._readField = function (tag, obj, pbf) {
    if (tag === 1) obj.Hash = pbf.readBytes();
    else if (tag === 2) obj.Name = pbf.readString();
    else if (tag === 3) obj.Tsize = pbf.readVarint();
};

// PBNode ========================================

const PBNode = {};

PBNode.read = function (pbf, end) {
    return pbf.readFields(PBNode._readField, {Links: [], Data: null}, end);
};
PBNode._readField = function (tag, obj, pbf) {
    if (tag === 2) obj.Links.push(PBLink.read(pbf, pbf.readVarint() + pbf.pos));
    else if (tag === 1) obj.Data = pbf.readBytes();
};


const pbf = new Pbf(serialized)
const data = PBNode.read(pbf)
console.log('actual hash:   89b1d9957f247ee0fbaa60726be4e138fe9983c5919dec1a1b8a43e7ed6509c8')
const cidv0Read = data.Links[0].Hash;
console.log('cidv0:    ', cidv0Read.toString('hex'))
const cidv1Read = data.Links[1].Hash;
console.log('cidv1:', cidv1Read.toString('hex'))
```

achingbrain: https://github.com/multiformats/cid#how-does-it-work

```
<multibase-prefix><cid-version><multicodec-content-type><multihash-content-address>
```

[CIDv0](http://127.0.0.1:8080/ipns/cid.ipfs.io/#QmXc9raDM1M5G5fpBnVyQ71vR4gbnskwnB9iMEzBuLgvoZ) does not contain a multibase prefix so it's assumed to be `base58btc`, [CIDv1](bafybeiejwhmzk7zep3qpxktaojv6jyjy72myhrmrtxwbug4kipt62zijza) does.

vmx: @achingbrain yes, and it also mentions there:

> NOTE: Binary (not text-based) protocols and formats may omit the multibase prefix when the encoding is unambiguous.

And this is what dag-pb is doing. The CID (no matter which one) is stored in binary without the multibase prefix (as the code above shows).

achingbrain: Good point, well presented.

You only need to include the multibase if the encoding is ambiguous - e.g. a CID encoded as a string.  A buffer or similar octet stream is not ambiguous so it can be omitted.

Since that's the case, you could probably remove the `(with no multibase prefix)` altogether.

achingbrain (approved):

block-layer/codecs/dag-pb.md
vmx: > Since that's the case, you could probably remove the `(with no multibase prefix)` altogether.

It's not strictly needed, but I think it adds clarity.

vmx:
There were only minor changes since some of the approvals, so if you feel like they weren't good, please open an issue/PR. I'll merge this now :)

PR #202, opened 2019-10-02
selectors: improve UnixfsV1 example

vmx:
The old example was hard to understand if you didn't really know how
UnixFSv1 is structured. The second selector example was removed as
it actually wasn't accurate and solely based on some prototyping work
I did some time ago.

I'm really way more happy with this example now :)

### DIFF
--- a/selectors/example-selectors.md
+++ b/selectors/example-selectors.md
@@ -196,12 +196,12 @@ The recursion comes into play with the `cid` field that points to another block 
         "name": "somedata.txt",
         "cid": {
           "data": "there's data in here",
-          "Links": []
+          "links": []
         }
       }]
     }
   },{
-    "Name": "subdir2",
+    "name": "subdir2",
     "cid": …
   }]
 }
--- a/selectors/example-selectors.md
+++ b/selectors/example-selectors.md
@@ -170,17 +170,18 @@ Selector:
 
 [UnixFSv1] is a good case of a recursive data structure. Any number of links can be used to create deeply nested structures. The following example is inspired by UnixFSv1, but uses a simplified structure, so that we don't get lost in UnixFSv1 specific implementation details.
 
-Our basic structure looks like this:
+The basic structure described by an [IPLD Schema]:
 
-```json
-{
-  "data": "<only-nodes-without-further-links-have-this-set>",
-  "links": [{
-    "name": "<name-of-the-link>",
-    "cid": "<a-cid-that-points-to-another-block>"
-  },
-  …
-  ]
+```ipldsch
+type FileSystem struct {
+  # Only leaf nodes have data
+  data optional Bytes
+  links [Link]
+}
+
+type Link struct {
+  name String
+  cid &FileSystem
 }
 ```
 
@@ -195,7 +196,7 @@ The recursion comes into play with the `cid` field that points to another block 
       "links": [{
         "name": "somedata.txt",
         "cid": {
-          "data": "there's data in here",
+          "data": "74686572652773206461746120696e2068657265",
           "links": []
         }
       }]
@@ -231,3 +232,4 @@ Selector:
 ```
 
 [UnixFSv1]: https://github.com/ipfs/specs/tree/master/unixfs
+[IPLD Schema]: ../schemas
--- a/selectors/example-selectors.md
+++ b/selectors/example-selectors.md
@@ -168,9 +168,33 @@ Selector:
 
 ### Getting a full sub-DAG
 
-For getting a full file from [UnixFSv1] you need to retrieve a full sub-DAG.
+For getting a full file from [UnixFSv1] you need to retrieve a full sub-DAG. A JSON representation of the structure we are querying would look like this:
 
-An example selector to get the full sub-DAG rooted at a certain CID:
+```json
+{
+  "Links": [{
+    "Name": "subdir1",
+    "Tsize": 87,
+    "Hash": {
+      "Links": [{
+        "Name": "somedata.txt",
+        "Tize": 30,
+        "Hash": {
+          "Data": "there's data in here",
+          "Links": []
+        }
+      }]
+    }
+  },{
+    "Name": "subdir2",
+    "TSize": 74,
+    "Hash": …
+  }]
+}
+
+```
+
+An example selector to traverse a full sub-DAG following the `Links` recursively and matching the `Name` and `Data` fields.
 
 
 ```yaml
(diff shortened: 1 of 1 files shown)


selectors/example-selectors.md
```diff
@@ -168,9 +168,33 @@ Selector:
...
+    "Tsize": 87,
+    "Hash": {
+      "Links": [{
+        "Name": "somedata.txt",
+        "Tize": 30,
+        "Hash": {
+          "Data": "there's data in here",
+          "Links": []
+        }
+      }]
```
rvagg: Is this what happens for an inlined unixfs block or are we blurring block boundaries here and just inserting a linked block where the CID would normally go?
Is it worth clarifying this? Maybe this isn't the best place to expand on UnixFSv1 but I don't feel well equipped to grok what this example is doing because of this gap in my UnixFSv1 knowledge.

rvagg (approved):
fine by me I suppose, although I feel least authoritative on this particular intersection of topics

selectors/example-selectors.md
```diff
@@ -180,39 +204,19 @@ Selector:
...
               next:
                 ExploreFields:
                   fields:
-                    "multihash":
+                    "Hash":
                       ExploreRecursiveEdge
 ```
 
-
-If it's a file in some directory, you can also start at a deeper level:
```
warpfork, line 193: Aw I kinda like this one :/  It's a cool easy-to-grasp concrete use case.

It *is* problematic to make it both A) accurate to unixfsv1 and all its baroque details, and yet B) terse enough to be a good example, though.  I think it's a choose-one.  (Or, indeed, choose neither, I suppose, yeah.)

selectors/example-selectors.md
```diff
@@ -168,9 +168,33 @@ Selector:
...
-An example selector to get the full sub-DAG rooted at a certain CID:
+```json
+{
+  "Links": [{
+    "Name": "subdir1",
+    "Tsize": 87,
+    "Hash": {
+      "Links": [{
+        "Name": "somedata.txt",
+        "Tize": 30,
```
warpfork: You got `Tize`, `Tsize`, and `TSize` in here somehow :)

selectors/example-selectors.md
```diff
@@ -180,39 +204,19 @@ Selector:
     sequence:
       ExploreFields:
         fields:
-          "Links"
+          "Name":
+            Matcher:
```
warpfork: So these additional matcher leafs are kinda interesting, and I wonder if we shouldn't add some more comment on them?  And also this has some interaction with Rod's comments about how much we want to blur block boundaries in the example data...

If one is just worried about "retrieve a full sub-DAG", as you set it up in the first line of the section... well, "retrieve" sort of invokes concepts that are halfway to something higher, like graphsync.  "Retrieve" is a concept entangled with _blocks_.

In practice, one will "retrieve" a bunch more _nodes_ than the ones explicitly matched or traversed searching for matches, because they'll ride along _in the block_ of anything that was explicitly matched or traversed.

Probably we want to be careful to use words about "match"/"visit"/"yield"(?) (I haven't been completely consistent on these to date -- Mea Cupla, and <3 help trying to improve this) most of the time.  Or we could introduce these details about blocks-vs-nodes and those impliciations during these examples too, but I think maybe that might be a little much (and if we do that, maybe we should do it in a non-unixfs example just to reduce the unrelated potential cognitive load for readers...).

warpfork:
Good examples are hard to refine, thank you so much for driving this :)

selectors/example-selectors.md
vmx: You are right, I'm blurring block boundaries to make this example more approachable.

I was trying to say with "A JSON representation of the structure we are querying would look like this:" that it isn't the actual structure (as it also isn't JSON). But I agree it's not clear enough. I'll rephrase that.

selectors/example-selectors.md
vmx: That example was actually wrong. Directories are not represented as named keys (i.e. not as `{"mydir": {"subdir": …}}`, but with the names within the link itself. So you would need to have a conditional matcher which hasn't been specified yet. I tried to add it to the new version of the example I just pushed, but then the whole example didn't really make sense anymore (why would you have named links at all?), hence I left it out.

vmx:
I thought I've left a comment, but apparently I didn't.

@rvagg, @warpfork: Could you please review it again? I changed the whole example, so the approval is not accurate anymore.

selectors/example-selectors.md
```diff
@@ -166,35 +166,49 @@ Selector:
 ```
 
 
-### Getting a full sub-DAG
+### Retrieving data recursively
+
+[UnixFSv1] is a good case of a recursive data structure. Any number of links can be used to create deeply nested structures. The following example is inspired by UnixFSv1, but uses a simplified structure, so that we don't get lost in UnixFSv1 specific implementation details.
```
warpfork, line 171: Did you mean to put a link here?

warpfork:
Updated example LGTM :+1:

selectors/example-selectors.md
```diff
@@ -166,11 +166,49 @@ Selector:
...
+```json
+{
+  "links": [{
+    "name": "subdir1",
+    "cid": {
+      "links": [{
+        "name": "somedata.txt",
+        "cid": {
+          "data": "there's data in here",
+          "Links": []
```
rvagg: s/Links/links

selectors/example-selectors.md
```diff
@@ -166,11 +166,49 @@ Selector:
...
+        }
+      }]
+    }
+  },{
+    "Name": "subdir2",
```
rvagg: s/Name/name

rvagg (approved):

rvagg:
Here's a bonus optional extra, when introducing the example data structure you could also do it with IPLD schemas:

```ipldsch
type Example struct {
  data Any
  links [Link]
}

type Link struct {
  name String
  cid &Example
}
```

(that `data Any` isn't technically right, but it's directionally right and I think we might end up with such a construct sometime, soon?).

selectors/example-selectors.md
vmx: It's linked at the bottom of the file.

vmx:
@rvagg Using IPLD Schemas for describing the structure is an excellent idea. Time for another quick review :)

selectors/example-selectors.md
warpfork: !!!  I learned a new feature of markdown on github today...

vmx: Whaaaat? That's so useful especially if you have so many moving parts as we have on the specs repo. It makes updating links so much easier (and also easier to review).

rvagg: Very -1 on these personally, I prefer not to use these, I hate having them separated. But I'm also fine being consistent when I'm editing a doc that someone else has started with this style.

rvagg (approved):
nice!

selectors/example-selectors.md
vmx: For me it's about being easier to the eye if you read the plain text file. You probably don't care about the exact links.

PR #203, opened 2019-10-14
Update schema field modifier docs

warpfork:
Update schema field modifier docs.

Especially around implicits -- those sections were quite old, and even missing the modern name (previously it still refered to "defaults", which is a term we sunset some time ago due to misleading associations).

Also introduced some longer discussion around considerations that might make one prefer using an `optional` vs and `implicit` when authoring a schema.  That gets a tad long, but I don't know if there's any better place for it at the moment.  (All this could surely be polished better yet -- this commit just gets us closer, and fixes the most glaring terminological desyncs.)

### DIFF
--- a/schemas/schema-kinds.md
+++ b/schemas/schema-kinds.md
@@ -115,6 +115,11 @@ names.
 ## Value Type Modifiers
 
 Values and fields in recursive types can have modifiers.
+These modifiers are:
+
+- `nullable` (for map values, list values, and struct fields)
+- `optional` (for struct fields)
+` `implicit` (for struct fields -- in some representations)
 
 ### Nullable Values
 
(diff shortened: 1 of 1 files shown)


schemas/schema-kinds.md
```diff
@@ -115,6 +115,11 @@ names.
 ## Value Type Modifiers
 
 Values and fields in recursive types can have modifiers.
+These modifiers are:
+
+- `nullable` (for map values, list values, and struct fields)
+- `optional` (for struct fields)
+` `implicit` (for struct fields -- in some representations)
```
rvagg: fix list character at the start of this line

schemas/schema-kinds.md
```diff
@@ -115,6 +115,11 @@ names.
 ## Value Type Modifiers
 
 Values and fields in recursive types can have modifiers.
+These modifiers are:
+
+- `nullable` (for map values, list values, and struct fields)
```
rvagg: I don't think these need parens, style choice and it's up to you tho

schemas/schema-kinds.md
```diff
@@ -143,25 +148,67 @@ valid members by one -- see the "Cardinality Examples" table below.
...
+
+### Choosing between Optional and Implicit
+
+Use `implicit` when the absense of a value should be treated as another
+value that fits unambiguously into your domain; use `optional` when there's
+no such in-domain value.
 
-### Combining Nullable, Optional, and Default
+For example, if you have some integer field, and if `0` is distinctive from
+from the field being absent, use`optional`.
```
rvagg: space between 'use' and `optional`

schemas/schema-kinds.md
```diff
@@ -143,25 +148,67 @@ valid members by one -- see the "Cardinality Examples" table below.
...
+If `0` (or some other value of your choice like `-1`) can be used as a
+entinel value in your application logic, while not being serialized as such,
```
rvagg: "sentinel"?

schemas/schema-kinds.md
```diff
@@ -143,25 +148,67 @@ valid members by one -- see the "Cardinality Examples" table below.
...
+then you may wish to consider using `implicit`.
 
-The `nullable` and `default` modifiers may be freely combined without issue.
+One usage pattern to particularly note: if writing a schema for definiting
```
rvagg: "defining"?

schemas/schema-kinds.md
```diff
@@ -143,25 +148,67 @@ valid members by one -- see the "Cardinality Examples" table below.
...
+configuration that a user may provide, and it has some concept of "defaults",
+and your application also needs to be able to *remember* whether a value was
+the default or explicitly user-provided (e.g. to re-emit the config without
+altering this)
```
rvagg: missing `.` but also maybe this thought needs to be extended slightly with advice for this situation? something like this?

> In this case, `implicit` will cause the provided value to be lost at the application-layer, whereas `optional` will expose and preserve its absence.

rvagg (changes requested):
very good, sans minor nits and comment about the incomplete-feeling final edit

schemas/schema-kinds.md
warpfork: Your skill with prose never ceases to humble me.  Taking that verbatim, thanks.

warpfork:
Took all of those edits -- thank you for your fine toothed comb

rvagg (approved):

PR #205, opened 2019-10-16
Graphsync: Extensions

hannahhoward:
# Goals

Support several additions to the Graphsync Protocol without constantly iterating on spec

# Implementation

- Consolidates #204 #111 into a single proposal for Graphsync Extensions that operate similar to TLS Extensions
- Adds a list of well known extensions:
1. The previous metadata proposal from #111 
2. A proposal for sending excluded CIDs with the request that the responder could use to narrow down the blocks it sends back (by not including any blocks that match the CIDs in the set.

### DIFF
--- a/block-layer/graphsync/known_extensions.md
+++ b/block-layer/graphsync/known_extensions.md
@@ -1,8 +1,8 @@
 # Graphsync: Known Extensions
 
-### Excluded CIDs
+### Do Not Send CIDs
 
-Extension Name: `graphsync/exclude-cids`
+Extension Name: `graphsync/do-not-send-cids`
 
 What it does:
 
@@ -18,10 +18,10 @@ When a requestor sends a request, it should send a CBOR encoded IPLD Node that i
 The IPLD schema is as follows:
 
 ```ipldsch
-type ExcludedCids [Cid]
+type DoNotSendCids [Cid]
 ```
 
-The responder node, if it supports the extension, will not send those blocks back in the response. It does not send a value back in the response for this extension.
+The responder node will execute the selector query as it would normally. However, if it supports the extension, when the selector query passes over any blocks that have a cid from the DoNotSend list, the responder will not send that block back, knowing ahead of time the requestor already has it. The responder does not send a value back in the response for this extension.
 
 ### Response Metadata
 
@@ -29,7 +29,7 @@ Extension Name: `graphsync/response-metadata`
 
 What it does:
 
-Response metadata provides information about the response to help the requestor more efficiently verify the blocks sent back from the responder are valid for the requested IPLD selector. It contains information about the CIDs the responder traversed, in order, during the course of performing the selector query and whether or not the corresponding block was present in its local block store.
+Response metadata provides information about the response to help the requestor more efficiently verify that the blocks sent back from the responder are valid for the requested IPLD selector. It contains information about the CIDs the responder traversed, in order, during the course of performing the selector query, and whether or not the corresponding block was present in its local block store. Telling the requestor immediately that the query passed over a block the responder did not have allows the requestor to advance its local query, and return a seperate error for that particular block.
 
 How it works:
 
--- a/block-layer/graphsync/known_extensions.md
+++ b/block-layer/graphsync/known_extensions.md
@@ -1,0 +1,63 @@
+# Graphsync: Known Extensions
+
+### Excluded CIDs
+
+Extension Name: `graphsync/exclude-cids`
+
+What it does:
+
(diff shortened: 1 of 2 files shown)


hannahhoward:
@anorth fyi this would also allow us to extend to support full piggy backing for data transfer and potentially a Graphsync push mode through an unspecified data transfer extension.

vmx (commented):
Sounds OK to to me, but I leave it to @Stebalien to approve as he knows more about graphsync and its vision/future directions.

warpfork (approved):
LGTM as well (and nice docs!)

block-layer/graphsync/known_extensions.md
```diff
@@ -0,0 +1,63 @@
+# Graphsync: Known Extensions
+
+### Excluded CIDs
```
Stebalien: * Does this exclude entire sub-trees (i.e., stop traversal) or just the node in question? IMO, this is the most _useful_ mode of operation.
* Can I _update_ the request with additional exclusions? If not, this feature feels like it belongs inside selectors themselves. @warpfork?

block-layer/graphsync/known_extensions.md
```diff
@@ -0,0 +1,63 @@
...
+
+The IPLD schema is as follows:
+
+```ipldsch
+type ExcludedCids [Cid]
+```
+
+The responder node, if it supports the extension, will not send those blocks back in the response. It does not send a value back in the response for this extension.
+
+### Response Metadata
```
Stebalien, line 26: * Why do we need this extension? I _should_ be able to derive this information locally at no cost (just execute the selector). I can't _not_ execute the selector for security reasons.
* Are we planning on _relying_ on this extension? If so, we _really_ should be baking it into the protocol.

Stebalien: > The proposal for the initial go implementation, in order to optimize memory usage, is to instead have the server send blocks as soon as it finds them, as well as information about what blocks are not available as soon as it knows. Then the client will verify and return blocks to the graphsync caller as soon as it receives them.

(from https://github.com/ipld/specs/pull/111#issue-269750451)

This worries me. We _can't_ return blocks to the caller until _we've_ verified that they're a part of the query. We can't just verify that the blocks match the CIDs reported by the remote side.

Stebalien (commented):
The extension format looks good to me but I'm confused about the proposed extensions.

block-layer/graphsync/known_extensions.md
hannahhoward: To be clear, the response metadata is already part of the go-graphsync implementation.

The comment above is not clear, and expresses intent poorly. The word "verify" in this case means verify that they are part of the query.

The most useful thing that response metadata does is tell the client about CIDs that it doesn't have. By knowing that CID's aren't present, and won't be incoming, the client can know immediately it can progress a it's local selector query and return an error for that part of the tree.

I can really work on rewriting this.

block-layer/graphsync/known_extensions.md
hannahhoward: So, to be clear, excluded cids -- and maybe it needs another name -- do not affect the execution of the query itself. It does not mean "stop at". Instead, it means "when you pass over this block during traversal, there's no need to send it back to me, cause I already have it". It's purely for network efficiency -- the main use case is I execute a graphsync query that had a network interruption half way through and I want to redo the request without sending the first half of the data.

My belief is this kind of data-efficiency exclusion is pretty tough to express in a selector query  -- there is an add-on coming to selectors that specifically includes the kind of "stopAt" in a recursive selector that you're talking about.

Stebalien: Ah, and we can't bundle it into a _selector_ because we need to send back the entire selector trace, even if a block doesn't match. Makes sense.

block-layer/graphsync/known_extensions.md
Stebalien: > The word "verify" in this case means verify that they are part of the query.

The client can't actually do that without the blocks, right?

> The most useful thing that response metadata does is tell the client about CIDs that it doesn't have. 

_That_ actually sounds pretty important. Should that not be a mandatory field? I guess it's _technically_ not required for correctness but I can't imagine an DoS-proof graphsync implementation that doesn't need this. That is, I'd expect a secure graphsync implementation to disconnect from peers that send over N (100?) blocks while we can't make progress in the query.

Stebalien (approved):
Got it. These extensions LGTM (although, IMO, we might need to require the second extension).

hannahhoward:
@Stebalien I'm going to merge this -- cause it's more information that currently exists in the spec, and file a seperate issue to move metadata out of an extension to a core part of the protocol.

PR #206, opened 2019-10-17
fix: minor schema-schema json formatting mistakes

rvagg:
was manually edited prior to tooling implementation, this is with tooling so it has proper ordering and spacing

seems trivial, I know, but I'm using the raw text of this to test my jsonifiers

### DIFF
--- a/schemas/schema-schema.ipldsch
+++ b/schemas/schema-schema.ipldsch
@@ -3,6 +3,7 @@
 ## Yes, it's self-describing! :)
 ## -----
 
+##
 ## Type names are a simple alias of string.
 ##
 ## There are some additional rules that should be applied:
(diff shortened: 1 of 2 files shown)


rvagg:
added another commit that does some minor cleanup on schema-schema.ipldsch, there was a stray space indent from the enum ints change, I've removed some redundant struct `representation map`s and done some minor cleanup on comments. It's now in a state that I can `cat schema-schema.ipldsch | grep -v '^#'` and get exactly what `ipld-schema to-schema` should produce.

@warpfork would you mind casting a quick eye over these changes?

schemas/schema-schema.ipldsch
```diff
@@ -2,7 +2,7 @@
 ## This is the schema describing the schema declarations for IPLD Schemas.
 ## Yes, it's self-describing! :)
 ## -----
-
+##
```
warpfork, line 6: Nei, above block does not belong attached to `TypeName`'s docs.

warpfork (approved):
LGTM with one tiniest nit, heh.

(If you poked at that comment because of a whitespace diff that breaks your tests, I'm extremely impressed with your tests.  But I guess would probably add a whitespace normalizing filter to them.)

schemas/schema-schema.ipldsch
warpfork: Admittedly, it's possible I'm a monster for writing a doc snippet that's not attached to any types or other structure at all.  (I've not done a lot of attending to how comments would be roundtripped through the "AST", should we choose to make a goal of doing so.)

rvagg: yes, you're a monster, I already had to special case this particular line: https://github.com/ipld/js-ipld-schema/blob/e1f80bd8a384d3075f3406b3a4ea222cc18f189e/test/_split-schema-schema.js#L31-L35
otherwise the pattern you established is quite nice

PR #207, opened 2019-10-17
RFC: Support recursive selectors that have no maxDepth for full DAG traversal

hannahhoward:
# Goal

In several discussions surrounding Graphsync, a common assumption is that you can ask Graphsync to "sync a DAG" -- as in, starting at a CID, traverse the node and anything under it, sending back all blocks that are linked from the root CID. 

However, we don't currently have a selector that actually does this. We can approximate the behavior with a selector that looks like this:

```ipldsch
ExploreRecursive {
   maxDepth 9999999
   sequence {
      ExploreAll {
          next {
             ExploreRecursiveEdge {}
          }
      }
   }
```

For those unfamiliar with the Selector DSL, this means "for each node, traverse all its fields, and then recursively apply the same traversal to the fields, up to a depth level of 9999999 (or other appropriately large number)"

The issue here is that we have to awkwardly supply a constant number of recursions and hope the DAG we're traversing isn't large enough to exceed maxDepth.

# Implementation

The proposal simply would make max depth optional, in which case recursion would continue until it stopped encountering ExploreRecursiveEdge{}. (this would normally occur cause it ran out of feels that could be traversed -- only map types and list types can be traversed (plus links, trasparently), so it would therefore stop as long as the dag had limits to it.

# For Discussion

There are various potential issues with working with selectors with high-recursion maxDepths or no maxDepth at all.

1. Memory requirements - currently selector implementation implementation uses a stack for recursive selectors, which means potential memory issues if recursive continues infinitely

2. Potential DOS vectors in an untrusted network context like GraphSync - if there is no limit, multiple requests for a very large DAG could potentially take down a node

At the same time, there is a convincing argument that all of these concerns should be left to implementation, NOT what is permissible a the selector itself.

For example, on the network DOS level, it seems particularly important to allow a Graphsync node to limit what selectors are accepted beyond the overall selector specification significantly (for example, a node might only accept maxDepth < 100).

Similarly a selector traversal algorithm should probably contain logic to prevent a traversal likely to case an out of memory limit.

In any case, there is a common user expectation that a selector + graphsync can be used to quickly "sync a dag" and therefore, it seems prudent to insure there is a good story there. And while "sync a dag" without limits might not make sense in an untrusted context like the IPFS network, it might make sense in several situations where one wants to do point-to-point transmission between two peers that have reason to trust each other (such as a pre-agreed upon transfer).

This PR is meant just to start a discussion where several interested parties can comment. I've somewhat paraphrased things others have told me for or against something like this, so feel free to explain your position in more detail since I am probably not doing a great job doing so.

Also, it is perhaps a start off for a larger discussion of "what kinds of concerns should determine what goes in the specification of selectors vs what is a concern of implementation or other libraries that use them"
@warpfork @anorth @jbenet @whyrusleeping

### DIFF
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -117,9 +117,14 @@ type ExploreRange struct {
 ## Be careful when using ExploreRecursive with a large maxDepth parameter;
 ## it can easily cause very large traversals (especially if used in combination
 ## with selectors like ExploreAll inside the sequence).
+##
+## If no maxDepth is specified, recursion will continue as long as
+## the selection continues to traverse ExploreRecursiveEdge - be very careful
+## with this
+
 type ExploreRecursive struct {
 	sequence Selector (rename ":>")
-	maxDepth Int (rename "d")
+	maxDepth optional Int (rename "d")
 	stopAt optional Condition (rename "!") # if a node matches, we won't match it nor explore its children.
 }
 
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -100,8 +100,8 @@ type ExploreRange struct {
 ## In implementation, whenever evaluation reaches an ExploreRecursiveEdge marker
 ## in the recursion sequence's Selector tree, the implementation logically
 ## produces another new Selector which is a copy of the original
-## ExploreRecursive selector, but with a decremented maxDepth parameter, and
-## continues evaluation thusly.
+## ExploreRecursive selector, but with a decremented depth parameter for limit
+## (if limit is of type depth), and continues evaluation thusly.
 ##
 ## It is not valid for an ExploreRecursive selector's sequence to contain
 ## no instances of ExploreRecursiveEdge; it *is* valid for it to contain
@@ -114,19 +114,28 @@ type ExploreRange struct {
 ## be thought of like the 'continue' statement, or end of a for-loop body;
 ## it is *not* a 'goto' statement).
 ##
-## Be careful when using ExploreRecursive with a large maxDepth parameter;
+## Be careful when using ExploreRecursive with a large depth limit parameter;
 ## it can easily cause very large traversals (especially if used in combination
 ## with selectors like ExploreAll inside the sequence).
 ##
-## If no maxDepth is specified, it is up to the implementation library using 
-## selectors to identify an appropriate max depth as neccesary so that recursion
-## is not infinite
+## limit is a union type -- it can have an integer depth value (key "depth") or
+## no value (key "none"). If limit has no value it is up to the 
+## implementation library using selectors to identify an appropriate max depth
+## as neccesary so that recursion is not infinite
 
 type ExploreRecursive struct {
 	sequence Selector (rename ":>")
-	maxDepth optional Int (rename "d")
+	limit RecursionLimit (rename "l")
 	stopAt optional Condition (rename "!") # if a node matches, we won't match it nor explore its children.
 }
+
+type RecursionLimit union {
+	| RecursionLimit_None "none"
+	| RecursionLimit_Depth "depth"
+} representation keyed
+
+type RecursionLimit_None struct {}
+type RecursionLimit_Depth int
 
 ## ExploreRecursiveEdge is a special sentinel value which is used to mark
 ## the end of a sequence started by an ExploreRecursive selector: the recursion
--- a/selectors/selectors.md
+++ b/selectors/selectors.md
@@ -118,9 +118,9 @@ type ExploreRange struct {
 ## it can easily cause very large traversals (especially if used in combination
 ## with selectors like ExploreAll inside the sequence).
 ##
-## If no maxDepth is specified, recursion will continue as long as
-## the selection continues to traverse ExploreRecursiveEdge - be very careful
-## with this
+## If no maxDepth is specified, it is up to the implementation library using 
+## selectors to identify an appropriate max depth as neccesary so that recursion
+## is not infinite
 
 type ExploreRecursive struct {
 	sequence Selector (rename ":>")


rvagg:
Instead of making it optional, how about introducing an explicit signal that must be set so the cost isn't entirely lost on whoever is writing such a selector. It's easier to drop an item than to have to set it to a signal value.
`-1` might be an obvious choice as an "unlimited", so would `"Inf"` but that would then require a union on `maxDepth` to allow for a string so not as great.

hannahhoward:
@rvagg I don't personally have a problem with that. there is some danger with optional meaning "I forgot to set this and now I have an unbounded selector"

anorth:
Please accept this change. A selector is an expression of some DAG subgraph/query. Being unable to even express a conceptually simple idea like "all of it" is very strange. Selectors should not be opinionated about how they are used - it's impossible to imagine all possible use cases and inevitable that there are some perfectly reasonable and safe uses of the "everything" selector.

Selectors are like SQL for DAGs, and this is like the SQL syntax standard requiring that all queries specify `LIMIT` or be syntactically invalid. 

The operational concerns of whether some module will accept and process an unbounded selector should not lie within the scope of this piece of syntax.

mikeal:
LGTM. I don’t see a huge argument for requiring maxDepth at the protocol level. I can see why you would want it optionally, but requiring it in the protocol doesn’t really buy you much since you’re going to end up setting a maxDepth in the implementation as a guard anyway. As you’ve already pointed out, if you need protection you’re going to have to fail on something much smaller than `999999` which is what people are already doing.

selectors/selectors.md
```diff
@@ -117,9 +117,14 @@ type ExploreRange struct {
 ## Be careful when using ExploreRecursive with a large maxDepth parameter;
 ## it can easily cause very large traversals (especially if used in combination
 ## with selectors like ExploreAll inside the sequence).
+##
+## If no maxDepth is specified, recursion will continue as long as
+## the selection continues to traverse ExploreRecursiveEdge - be very careful
```
mikeal: In the case that the implementation has its own maxDepth for protection, the semantics wouldn’t be that not setting it traverses indefinitely, it would be that it just fails once it hits a specific depth.

This implies that, if the implementation has its own maxDepth, it should fail when no maxDepth is sent since it will not support an infinite traversal. I don’t think that’s the intended behavior though, I think the intended behavior is probably for recursion to go until it reaches the maxDepth it set as a default.

vmx:
I first though "there would be some limit anyway, so why making it optional?". But I agree with @anorth point that those things shouldn't be enforced on a protocol level. So I'm in favour making `maxDepth` optional.

Thinking about Selector limits in general made me open https://github.com/ipld/specs/issues/210.

mikeal (approved):

warpfork:
I have mixed feelings about this.  For posterity, some of my comments from another channel of discussion before we raised it to github:

---

While "large arbitrary constants are bad" is usually solid rule of thumb...

In this case, the choice of constant is also a discussion-forcer.  And I'm inclined to suspect that the utility of the discussion-forcer is probably larger than the weight the rule of thumb about constants can counterbalance.

This kind of discussion-forcer is something I came to appreciate from Facebook's GraphQL.  It actually doesn't support recursion... *at all*.  The closest thing they have are making snippets (segments? fragments? I forget what they called them) and manually creating a tree of `snippet{snippet{snippet}}`.

At first this made me blink rapidly.  But after seeing it in use in a variety of teams... I came to see some virtue of this.  By making recursion a syntactic burden... It's a discussion-forcer.  And I actually saw it force a bunch of good discussions: it stopped several bad designs with unbounded recursions, as well as several designs that would've been "N+1" query problems, sometimes in situations that would have made O(n^2) or O(n^3) scenarios overall, from reaching production.  It was still code review that was the first and most important junction that made these things get caught; but the presence of the syntactic load made a clear focus where these particular issues could be *seen* and where the code review could *latch on*.  It helped.

The mapping of this concept from GraphQL doesn't 100% apply to our own IPLD Schemas, because recursion in GraphQL is entangled with a whole bunch of other interesting semantics in-practice that _our_ recursion isn't.  But the general anecdote is still a good lesson, I feel.

warpfork:
Other weighings:

I'm also a little dubious that limits coming from the outside of the selector are a better idea than keeping them in the selector declarations.  It's a bit "six of one, a half dozen the other", but I suspect that for a library to choose a constant above which to not accept selectors will tend to end up being perceived as capricious...  whereas forcing the visible existence of the constant in the selector declaration makes it less capricious (or at least homes the capriciousness with the right party), as well as giving it that good shot at being encountered in code review (even for junior folks who aren't used to looking for problems like this).

I am pretty swayed by the comparison to SQL query "LIMIT" clauses.  It depends on the use case, though.  If thinking about it in terms of graphsync, it doesn't map super well.  If thinking about it in terms of how the go-ipld-prime library internally does it -- a visitor callback, one step at a time -- it maps better.  The important detail is that SQL tends to be interacted with via some kind of iterator or visitor, which means partial processing is always possible even when the selection is huge.  That question of how partial progress works is the key bit, I think.  When selectors are used in a similar way, the importance of up-limits is similarly dampened.  I'm not sure how to strongly to weight this comparison overall, though.  How many of our use cases carry the partial progress mechanisms to the next layer up of consumer?

mikeal:
@warpfork I think I agree entirely on the constraint you want to impose, I’m just having a hard time understanding why this is the right place to enforce the constraint.

I agree that *every implementation should have a max-recursion* but a client needs a way of saying “query this to the max-recursion you will allow” since there’s no fixed max-recursion in the spec and will vary based on the implementation and configuration.

I don’t actually see the utility in requiring the client to put an arbitrarily large recursion depth because the semantics are either “try with a large recursion depth and we’ll ignore it if the implementation has a lower guard set” which means forcing the client to include it was actually pointless and semantically identical to just not setting it in this proposal, or it means that the client will immediately receive an error given a recursion depth that is too large, which would mean that every client request to a new implementation is suffering a round trip to discover that implementation’s maximum depth.

Also, recursion depth isn’t the only guard we should be encouraging implementations to have. If an implementation can limit the load a client causes based on timeout, CPU usage, memory usage, and other vectors it definitely should. Essentially, *any* request should assume to be limited by any number of constraints the implementation decides to enforce and we should create a robust enough error handling system that these constraints can be surfaced to the client when they occur.

I spend a lot of time dealing with GitHub’s GraphQL API and I will say that in this dimension they are severely lacking and not something I would use an example of best practice. The protocol includes features for rate limiting but, in addition to those limits, imposes several other undocumented rate limits that cause undocumented errors in the HTTP and socket layer. I’d hate to replicate this pattern by focusing entirely on one dimension of this problem, enforcing it arbitrarily in the client protocol, and then leaving the rest to occur in difficult to diagnose error conditions.

warpfork:
Other questions (that I don't necessarily have answers for, nor know if we need answers for, but nonetheless came to mind):

-  Do other potential user stories -- like pinning for example, which might like to use selectors in the futre -- offer us any insight into what's likely to be desired?
  - e.g. would you let anyone generate a pin in a tool with some UI if they didn't already have a manifest estimating size and providing accurate depth info?
- How many user stories involve working with wildly mixed-depth data in the first place other than blockchain?
 - ... Which is probably the one place where you should absolutely *least* use unlimited recursion, as the oh-shoot-the-whole-chain-is-a-pretty-big-download problem in all block chains beyond minimal adoption have learned the hard way?

warpfork:
I like @rvagg 's proposal a lot.  How do others feel about that?  What if we go all-in on that and do indeed replace `maxDepth` with a union?

I'm thinking `type RecursionLimit union` with initial members None and Depth, like so:

```
type RecursionLimit union {
	| RecursionLimit_None "none"
	| RecursionLimit_Depth "depth"
} representation keyed

type RecursionLimit_None struct {}

type RecursionLimit_Depth int

type ExploreRecursive struct {
	sequence Selector (rename ":>")
	limit RecursionLimit (rename "l")
	stopAt optional Condition (rename "!")
}
```

hannahhoward:
I am personally fine with @warpfork  + @rvagg and like the union type. Selectors are decoded at parse time so there's minimal performance penalty to a union

hannahhoward:
and the existing builder interface in IPLD Prime will make specifying the union type not all that hard.

mikeal:
You had me at “union.” 🥰

vmx:
I do like @warpfork's argument about code review and making things easy to spot they are expensive. Hence I'm also in favour of using a union.

hannahhoward:
@warpfork @rvagg @vmx @rvagg updated to use union type, merging unless anyone has objections in the next 24 hours :)

selectors/selectors.md
```diff
@@ -117,12 +117,26 @@ type ExploreRange struct {
...
+##
+## maxDepth is a union type -- it can have an integer value (key "depth") or
+## no value (key "none"). If maxDepth has no value it is up to the 
+## implementation library using selectors to identify an appropriate max depth
+## as neccesary so that recursion is not infinite
+
 type ExploreRecursive struct {
 	sequence Selector (rename ":>")
-	maxDepth Int (rename "d")
+	maxDepth RecursionLimit (rename "l")
```
warpfork: Could also change the field name to something else at this point, but that's just varnish.

warpfork (approved):
Semantics of this now LGTM :+1:

rvagg (approved):

PR #213, opened 2019-10-23
Feat: Move metadata to core protocol

hannahhoward:
# Goals

Just a follow-up to #205 -- @Stebalien made the valid point that response metadata is actually necessary to protect against DOS vectors from a malicious responder, and therefore should not be an optional extension. This makes sense especially given `go-graphsync` implements metadata as part of its protocol already.

# Implementation

Move metadata into core protobuf, not an extension (I don't want get into required vs optional extensions for now -- all extensions are optional and are not required to be supported in the protocol ATM)

### DIFF
--- a/block-layer/graphsync/graphsync.md
+++ b/block-layer/graphsync/graphsync.md
@@ -74,7 +74,8 @@ message GraphsyncMessage {
   message Response {
     int32 id = 1;     // the request id
     int32 status = 2; // a status code.
-    map<string, bytes> extensions = 3;    // side channel information
+    bytes metadata = 3; // metadata about response
+    map<string, bytes> extensions = 4;    // side channel information
   }
 
   message Block {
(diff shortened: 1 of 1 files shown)


block-layer/graphsync/graphsync.md
```diff
@@ -90,6 +91,40 @@ message GraphsyncMessage {
...
+```ipldsch
+type LinkMetadata struct {
+  link Cid
+  blockPresent Bool
+}
+
+type ResponseMetadata [LinkMetadata]
+```
+
+This field is required as well to help protect from DOS attacks by malicious responders.
```
Stebalien, line 127: So, as much as I like IPLD/CBOR, wouldn't it be easier to just have a "skipped" list?

```
  repeated Block    data      = 4; // Blocks related to the responses
  repeated bytes    skipped = 5; // CIDs of blocks related to the response that have been skipped.
```

(or something like that)

hannahhoward: honestly my inclination is to go with what is already implemented, and also, IMHO there might be useful aspects to having the ones that are present ass well, even though I somewhat struggle to come up with them. Keep in mind due to deduplication you might have metadata that is not tied to an actual block in the response. As it's written, the symbolic meaning of the metadata in the response is: these are all the block traversals you ought to have covered with what I've sent you (including this message)

hannahhoward: anyway, unless there's a strong objection, I'd rather table this for the future

Stebalien: So, the idea is that we could include arbitrary per-node metadata? As in, this metadata could include other information about the query (e.g., why the block was selected)? I really can't think of anything the requester should _trust_ from responder other than "I'm not going to send you this block". Everything else should be figured out locally.

>  these are all the block traversals you ought to have covered with what I've sent you (including this message)

Won't including blocks we've sent have a pretty high overhead in terms of space?

> anyway, unless there's a strong objection, I'd rather table this for the future

If we include this feature, we'll never revisit this discussion. We could leave this as "it doesn't really matter, either way works" but I'd like to get on the same page before we do that.

---

This _mostly_ rubs me the wrong way because we're:

1. Making a general-purpose thing with only one use-case in mind. This rarely works out well in practice.
2. From what I can tell, breaking the clean "protobuf for networking, cbor for IPLD" divide. Unless I'm missing something, this _feels_ like something at the graphsync layer, not the selector/query layer.

hannahhoward: I'm happy to table this for now. It's in the spec as an extension as is, I don't feel the need to move this to metadata. I will just adapt the implementation to match what's in the spec currently, which is for it to be in an extension.

Stebalien: That's probably the cleanest solution. Changing the format in the core protocol is probably trickier than simply adding it.

mikeal:
Closing for now based on Hannah’s last comment.

PR #214, opened 2019-10-24
move data-structures to top-level

rvagg:
This is annoying me now that `schemas/` was moved to the top level. I think we all knew this was inevitable once that decision was made. As I write docs and cruise around the specs repo, this stands out as out-of-place, weird and unlikely to be a source of permalinks.

Unless there are better ideas? I'm still afraid of another major reorg because it still doesn't _feel_ like we have things in a proper ordering with a reasonable hierarchy.


vmx (approved):
I don't really have an opinion on that one.

mikeal (approved):

warpfork:
:+1:

PR #216, opened 2019-10-28
schemas: user-facing docs

rvagg:
I'm playing around with creating docs that are a bit more new-user friendly. So far it's wrapping up into a single doc with links to existing docs but I think I'm going to be extending it out into those docs and cleaning them up a little so it's all more connected and we cover enough of the bases (the DSL is the biggest hole atm so I need to get stuck into that too).

I'm not sure yet if this should be README.md or introduction.md. It feels more GitBook-style and might lend itself to a README that's just a big TOC. There's overlap here with overview.md too which I think I might merge in. But It also might end up big enough that this gets split into smaller docs. We'll see I'll keep on writing and massaging and see where it ends up. Feedback welcome wrt technical accuracy, but it's still early, so don't invest too much time reviewing yet unless you want to join in the fun too.


rvagg:
Working through Schema DSL docs at the moment. That's the last major missing piece before this is grokkable by external folks I think. When the content is organised and TOCs pulled up to the top level it turns into quite a "book", see: https://github.com/ipld/specs/blob/rvagg/schema-docs/schemas/README.md

rvagg:
The "Schema DSL" doc is expanding beyond a simple DSL primer and turning into something a bit different. https://github.com/ipld/specs/blob/rvagg/schema-docs/schemas/schema-dsl.md
I think I'm going to change its name to "Authoring IPLD Schemas" because that's the perspective it's trying to tackle.

The process is highlighting for me all the edge cases that we have to deal with at some point in Schema parsing and handling (mainly wrt schema validation—is the schema you've supplied even valid, can you even do that with a schema and our tools?). I can't decide whether, or how, I should be taking notes about these things as I encounter them.

rvagg:
The DSL doc has been renamed to "Authoring IPLD Schemas" and is a fairly comprehensive introductory document to writing Schemas, covering the DSL and a lot of the additional pieces you need to understand. The other documents provide detail and background.
I've also redone the Representations doc, added some that were missing and made all of the entries consistently detailed.

This is as much as I wanted to achieve in this pass. Next I'll do my own proof pass but I'd like some quick technical sanity checking @warpfork @vmx &&/|| @mikeal if you can spare some time to look over it. I think the best way to consume it, given that it's such a major overhaul, is to read the rendered Markdown @ https://github.com/ipld/specs/blob/rvagg/schema-docs/schemas/README.md and then come back here to comment inline on specifics. I've touched most of the material even if just for formatting.

Beyond that, what are the chances of finding someone with the time and inclination from outside of this team to read over it and see if it gives them a solid understanding of what we're trying to do with Schemas? That would be a good test of whether these changes are worthwhile.

schemas/README.md
```diff
@@ -3,19 +3,76 @@ IPLD Schemas
...
+    * [Schema Kinds](./schema-kinds.md#Schema-Kinds)
+  * [Value Type Modifiers](./schema-kinds.md#Value-Type-Modifiers)
+    * [Nullable Values](./schema-kinds.md#Nullable-Values)
+    * [Optional Fields](./schema-kinds.md#Optional-Fields)
+    * [Fields with Implicit Values](./schema-kinds.md#Fields-with-Implicit-Values)
+    * [Combining Nullable, Optional, and Implicit](./schema-kinds.md#Combining-Nullable-Optional-and-Implicit)
+    * [Choosing between Optional and Implicit](./schema-kinds.md#Choosing-between-Optional-and-Implicit)
+  * [Understanding Cardinality](./schema-kinds.md#Understanding-Cardinality)
+    * [Cardinality Examples](./schema-kinds.md#Cardinality-Examples)
+* **[Representations of IPLD Schema Kinds](./representations.md)**
```
warpfork, line 54: The boldings of the root level entries is a subtle but nice touch.  Makes it perceptibly easier to scan down the list.   :+1:

schemas/feature-summary.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+**Hooks for transparently transforming data with programmatic logic**
+
+```ipldsch
+advanced ROT13
+type EncryptedString string representation advanced ROT13
+```
+
+All of this maps over the IPLD Data Model:
```
warpfork, line 77: The transition from a bunch of bold headings above to here seems to sort of skip a beat.  Some other separator hereabouts seems desirable?

schemas/goals.md
```diff
@@ -0,0 +1,10 @@
+# IPLD Schemas Goals
+
+1. Provide a reasonable, easy to use, general type system for declaring useful properties about data.
+2. Compose nicely over the [IPLD Data Model](../data-model-layer/data-model.md): Schemas should be usable with any data format you can build an [IPLD Codec](../block-layer/codecs/README.md) for.
+3. Be efficient to verify: predictable; roughly linear in the size of the data and Schema; and absolutely not turing complete.
+4. Be language-agnostic: many compatible implementations of the Schema tooling should exist, as well as bindings for many language.
```
warpfork: This... might actually be an ancient typo of _mine_, haha, but since I just saw it... missing 's'

schemas/introduction.md
```diff
@@ -0,0 +1,81 @@
...
+
+IPLD’s Data Model is useful for building an abstraction at the data layer. It pushes concerns about encoding and decoding to its own dedicated domain where such concerns only have to deal with a limited set of data types and how to convert to and from those data types. But it also makes room for the development of an ecosystem of libraries and tools that are able to build on top of this model and can therefore operate according to shared assumptions about how data is presented. Such an ecosystem should empower significantly more sharing of algorithms, tools, techniques and code in the distributed web community, rather than siloing these assets in codebases that are concerned entirely use-case-specific stacks.
```
warpfork, line 9: I have no comment except that these paragraphs are fabulous and we've much needed them.

warpfork (commented):
I'm only a fraction of a way through this but I'm already super happy with everything going on here.

schemas/feature-summary.md
rvagg: Yeah, good point. I've introduced two secondary headings, at the top: "Constructs for Describing Common Data Forms" and above this bit "Mapped to the Data Model".

schemas/feature-summary.md
```diff
@@ -0,0 +1,91 @@
...
+  | "Foo"
+  | "Bar"
+  | "Baz"
+}
+```
+
+**A simple syntax for 'maybe' data elements**
+
+```ipldsch
+type Foo struct { f nullable Bar }
```
vmx: No need to change this, but for me it would be easier to parse when written as:

```iplsch
type Foo struct {
  f nullable Bar
}
```

schemas/feature-summary.md
```diff
@@ -0,0 +1,91 @@
...
+type Bar [nullable Foo]
+```
+
+**A clear syntax for non-required fields**
+
+```ipldsch
+type Foo struct { f optional Bar }
```
vmx: No need to change this, but for me it would be easier to parse when written as:

```iplsch
type Foo struct {
  f optional Bar
}
```

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+* Link: The `&` token prefixing a type is used as a shorthand for links. A generic link to an untyped resource uses the special `&Any`, while a link where there is an expected type to be found uses that type name as a hinting mechanism, `&Foo`. See below and [Links and IPLD Schemas](./links.md) for more information.
+* Union: Appears as `union` following `type` and the Union's type name.
+* Struct: Appears as `struct` following `type` and the Struct's type name.
+* Enum: Appears as `enum` following `type` and the Enum's type name.
+* Copy: Uses the shorthand `=` to indicate a copy type, as in `type Foo = Bar`. The token "Copy" does not directly appear in the Schema DSL.
+
+
+## Naming Types
+
+By convention, type names should begin with a capital letter although this requirement is not strict. Type names must only contain contain printable ASCII characters excluding space and punctuation other than underscores and the first character should be a letter. A strict regular expression for type names would be: `[a-zA-Z][a-zA-Z0-9_]*`. A regular expression following convention would be: `[A-Z][a-zA-Z0-9_]*`.
```
vmx: > Type names must only contain contain printable ASCII characters excluding space and punctuation other than underscores…

It may be simpler to say:

Type names must only contain alphanumeric ASCII characters including underscores…

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+  data MapOfIntLists
+}
+```
+
+is equivalent to:
+
+```ipldsch
+type Foo {
+  id Int
+  data [{String:Float}]
```
vmx: That's not correct. It should be:

```ipldsch
  data {String: [Int]}
```

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+}
+```
+
+As with typedef'd scalar kinds, this has implications for codegen and other API interactions with Schema types. Rather than having an explicit name, `FloatMap`, a generated name may be applied to the List type found at `Foo->data` and the type of the Map nodes found within that list.
```
vmx: `FloatMap` then also doesn't really work anymore (if it is a map if int lists).

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+```
+# This is not valid IPLD Schema but is presented to illustrate the additional verbosity being avoided
+
+type Foo struct {
+	fieldOne nullable String
+	fieldTwo Bool
+} representation map {
+  fields {
+  	fieldOne rename "one"
+	  fieldTwo rename "two" implicit "false"
```
vmx: Indentation is off.

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+In this case we are mapping Int values at in the serialized form to the three Enum values. Note also that the values are again quoted, but will be interpreted appropriately as integers because the context makes that clear.
+
+## Structs
+
+The basic DSL form of a Struct has the following structure:
+
+```
+type TypeName struct {
+  Field1Name Field1Type
+  Field2Name Field2Type
```
vmx: I would find it clearer when `Field1Name` and `Field2Name` would start with a lowercase character.

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+
+Structs must always have a body, enclosed by `{`, `}`. Fields must new-line delimited and should be indented for clarity.
+
+The `representation` strategy for Structs is `map` by default, so may be omitted. Additional representation strategies See [Representations of IPLD Schema Kinds](./representations.md) for more details on these representation strategies.
+
+Field representation parameters are presented in parens when present and representations requiring additional general parameters is presented in a separate `representation` block enclosed by `{`, `}`. For example, a Struct laden with both field representation parameters and general representation parameters:
```
vmx: Just want to note that "laden" wasn't even part of my passive vocabulary, this might not me mean as mine is quite small, but I thought I bring it up. I would say in my words "a Struct containing both…", but feel free to leave it in or find less dull sounding version.

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+
+```ipldsch
+type Foo struct {
+	fieldOne nullable String (rename "one")
+	fieldTwo Bool (rename "two" implicit "false")
+} representation stringpairs {
+  innerDelim "="
```
vmx, line 374: Indentation is off.

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+	entryDelim ","
+}
+```
+
```
vmx, line 378: I would expect seeing an example of a serialized representation of it (using implicitly the data from above):

    one="This is field one of Foo"

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+{
+  "msg": "All good",
+  "payload": {
+    "percent": 0.6,
+    "last": "61626378797a"
+  }
+}
+```
+
+In this example, we have a Map that can be represented as a Struct since it has only two fields, but the `progress` field dosn't have a stable kind so we can't use any of the existing Schema types to represent the field type. Instead, we can introduce a Union and can take different forms depending on the different acceptable forms.
```
vmx: `progress` -> `payload`.

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+
+type ProgressOrPing struct {
+  percent optional Float
+  last optional String
+  ts optional Int
+  nonce optional String
+}
+```
+
+Now the user of such a Schema must do their own field inspection to determine whether a `ProgressOrPing` is a progress message or a ping. Additionally, the burden of ensuring that both `percent` and `last` are present _or_ `ts` and `nonce` are present is left to the user, the Schema layer can't help here. The trade-off present in this scenario regards validation of a node by inspection of its child nodes. Type type of data is common in the real world but IPLD Schemas encourage better data shape design to allow for fast validation through clear discrimination where such variance exists.
```
vmx: Typo: "Type type".

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+```ipldsch
+type Message struct {
+  msg String
+  payload Payload
+}
+
+type Payload union {
+	| Error "error"
+	| Progress "progress"
+  | Ping "ping"
```
vmx: Indentation is off. Though I would say the `Error` and `Progress` one is off (also in the other examples above).

I', going to stop mentioning indention within examples here. You could just check all examples :)

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+}
+```
+
+```json
+{
+  "msg": "All good",
+  "tag": "progress",
+  "percent": 0.6,
+  "last": "61626378797a"
+  }
```
vmx: Stray bracket.

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+The interface presented by this Schema is adjusted in comparison to the previous Unions as `Error` is now a Struct with a `message` field.
+
+### Byteprefix Unions for Bytes
+
+A special case union exists for handling Bytes kinds. Where a node contains a byte array (Bytes kind), we may want to discriminate between two different uses of that byte array at the application layer. For example, consider two different encoding schemes where we store a "key" field that is distinct for the each encoding scheme. For practical purposes they are both byte arrays, but at the application layer it helps to have them separated into distinct forms, perhaps so we can make simple assertions about getting the expected key type for the given encoding scheme. There are additional documentation clarity benefits for extracting distinct forms and naming them in a Schema that may factor in to such a decision.
+
+```ipldsch
+type Authorization struct {
+  key PublicKey
+  keyType String
```
vmx: I find the field named `keyType` a bit confusing. Why would you add the type as string when you discriminate based on the byte prefix. Perhaps the additional field could be something like `keySize Int`?

schemas/authoring-guide.md
```diff
@@ -0,0 +1,829 @@
...
+  nonce String
+}
+
+type Pong struct {
+  ts Int
+  nonce String
+}
+```
+
+The Schema tooling and the reified form of the Schema retains a `copy` kind marker, but tooling that consumes Schemas is expected to treat this marker as an indirection to the named type being copied and copy the entirety of that type's definition to the new name.
```
vmx, line 776: "reified"? In this context does it mean in the "processed form of the Schema"?

PS: Fun fact, the German translation is "re­i­fi­zie­ren" which I've also never heard of. According to the major German dictionary (Duden) it's not in the top 100000 of most used words in the past 25 years.

vmx (changes requested):
Wow, that is a really nice read. Good job @rvagg.

I'm not done with the review yet, I haven't looked at `introduction` and `representations` yet, but will do so early next week.

schemas/feature-summary.md
rvagg: 👍 I agree, I've been intentionally changing these one-line structs when I find them, even though @warpfork explicitly wants the option, I find the clarity of consistency important in docs.

This still makes me chuckle when I see it so I haven't removed it yet (c/o Jeromy): https://github.com/ipld/go-ipld-schema/blob/4b713b12045f9cfc22451e8350d7664efeaad86c/parser/parsing.go#L80

schemas/authoring-guide.md
rvagg: 👍 changing to:

> Type names must only contain alphanumeric ASCII characters and underscores, and the first character should be a letter.

We're making rules here, but I'm more than comfortable with being restrictive. schema-schema states it as:

```
## There are some additional rules that should be applied:
##   - Type names should by convention begin with a capital letter;
##   - Type names must be all printable characters (no whitespace);
##   - Type names must not contain punctuation other than underscores
##     (dashes, dots, etc.).
```

"all printable characters" is pretty expansive, but I'm happy restricting it down to just alphanumeric.

@warpfork FYI this is a spec change.

schemas/authoring-guide.md
rvagg: going with this:

> As with typedef'd scalar kinds, this has implications for codegen and other API interactions with Schema types. Rather than having a explicit names, `MapOfIntLists` and `IntList`, auto-generated names may be applied to `Foo->data` and the type of the List nodes found within that Map. (e.g. perhaps `Foo__dataType`, `Foo__data__valueType`).

schemas/authoring-guide.md
rvagg: removed, probably an unnecessary word

schemas/authoring-guide.md
rvagg: it's the tab vs space thing, going with spaces in here, it's my doc so my rules, tabs be damned!

warpfork: "Tabs for indentation, spaces for alignment" is a rule I favor because it works consistently in all environments and preferences.

(Though docs and examples occasionally get a special pass just because screen realestate premiums in docs can be unusually high...)

schemas/authoring-guide.md
rvagg: eh, I don't know where to do here, we need a name for this that's concise and specific and "processed form" is neither. We're using "reified" in a few other places, including introduction.md, but it's a word both @warpfork and I have had trouble with I think, perhaps it's this problem of not being a common word.
It arguably is the right word though, https://en.wikipedia.org/wiki/Reification_(computer_science)

If we come up with a better word then I'd be happy to use it, I'm kind of sick of being vague about what this thing is that we stick `.json` after when we spit it out.

schemas/authoring-guide.md
warpfork: I'm thumbs up to changing it to "alphanumeric".  Wider tolerance for fascinating characters will pretty much only ever result to more frequent headaches in codegen and etc tooling which interacts with other systems (which have their own limitations, thus potentially requiring translation rules or override supports, and so on), as far as I can tell... and that brings no one joy.

warpfork: While we're here... I've been thinking it might be nice to reserve double underscore in names as well.

In codegeneration, it often comes up that other symbols are needed which are a concatenation of the name of the type in the schema, plus some other internal munging, and having character sequences that's typically-available-in-host-languages but reserved from schema names seems like a nice simplifier.

schemas/authoring-guide.md
rvagg: Tere's my rule: tabs where the context demands and where my editor makes it easy, spaces otherwise.
Almost everything I do is tabless so my editors all do 2-spaces when I press `tab`. In these docs where I need a tab I have to go copy it from somewhere and paste it in! Go and Makefiles are the special cases but thankfully my editors can deal with that.

schemas/authoring-guide.md
warpfork: > begin with a capital letter

^ was intentional, btw.  If we have non-capital starter characters in golang, they'll end up needing to be capitalized, since that's semantic in golang.

And while I wouldn't generally want to take a foible of one programming language as a strong argument for anything, capitalized class names are pretty common across many languages (even if they're not always enforced), so it seems not a stretch to just formalize that requirement.

rvagg: OK, latest commit clears it up here and in schema-schema:

> Type names _must_ only contain alphanumeric ASCII characters and underscores. The first character _must_ be a capital letter. Multiple connected underscores _should_ be avoided (they should be reserved for codegen purposes)

making the dunder rule a "should"

warpfork: Sounds fine to me.

schemas/authoring-guide.md
vmx: Oh, I didn't know it's a computer science-y word. I'm fine with it (which can well be, my domain expertise is database indexing and some geo stuff). I tried a normal dictionary as I didn't identify it as a technical term. If I would've used Wikipedia, it would've been easy.

schemas/representations.md
```diff
@@ -149,102 +244,241 @@ type Fizzlebop struct {
...
+
+Some data matching the `Foo` Struct (shown as JSON) is:
+
+```json
+[["fieldOne", "this is field one"], ["fieldTwo", true]]
+```
+
+Each entry of the parent List is a List containing exactly two entries, where the first is the exact field name and the second is the value.
+
+No field-specific parameters are available for the `listpairs` Struct representation strategy (i.e. `implicit` and `rename` which are available for `map`), although these may be introduced in the future since they are technically possible with this strategy. No general parameters are available.
```
vmx: It is true of `stringpairs`, `stringjoin` (`implicit` only) as well. So perhaps this paragraph could be put in the **Representation Kind: Map** section that field-specific parameters are only to the map-representation and not for others, but it might be added in the future.

vmx (changes requested):
(minor thing) the indentation in the examples in schemas/representations.md are also off.

I've read through all of it now. It's really nice. It's also now even clearer how much work went into the whole Schema stuff. It's an amazing system.

schemas/authoring-guide.md
warpfork: Yeah, it's a term I'm somewhat fond of in comp sci literature.

I'd also agree it would be great to find another term for the thing we're refering to here.  "Reified foo" or "reify the frob" is more like a contextual modifier or general verb... it's not great for a title of a significantly important point in the lifecycle of some data.  It works here, but we have this question for this particular piece of data come up in a bunch of places, so a more proper noun for it would be really nice to come up with.

rvagg: AST being a word we often reach for, but it's not a syntax tree. "abstract data tree" was suggested but it doesn't seem to pin anything down - also I don't think "abstract" is even the right word, the thing we're talking about is _more_ Schema than the DSL is.
A closer term is IR, "intermediate representation", that's much more accurate than AST if we're leaning on compile metaphors. "internal representation" is another one I've reached for but I don't mind the simplicity of "intermediate representation". What think you?

schemas/representations.md
rvagg: yeah, it gets iffy when you move outside of the key=value representations, but still possible. Will move it like you suggest though.

schemas/authoring-guide.md
warpfork: I see the resonance with "IR" too, yeah.

Some phrasing that emphasizes that this format Is The Thing -- and the DSL by contrast is just window-dressing -- would be ideal...

vmx (approved):

mikeal:
This looks great!

I’m going to need to cut out some time for a more thorough review eventually, but given that we have close to zero documentation currently I’d say we should fast track this to get it merged and continue to iterate on it.

rvagg:
ok, so re "reified" or "intermediate representation", let's deal with that as a separate PR so we can isolate that discussion and see all the pieces that touches.

I'm just playing with transforming these into pretty html so maybe we can publish them @ specs.ipld.io/schemas, will hold this open because I suspect it might need a little reorg, authoring-guide.md might be too big and need to be broken up.

PR #217, opened 2019-10-28
[WIP] UnixFSv2 Schema

mikeal:
This is the schema that is currently implemented in [`js-unixfsv2`](https://github.com/ipld/js-unixfsv2).

A lot of kinks in the structure of the file, directory and data layout have had the kinks worked out in the process of implementation. However, the file metadata hasn’t been used yet and the tradeoffs we’ll need to discuss about that aren’t the kinds of things implementation along will resolve and we should use this PR to work them out.

### DIFF
--- a/data-structures/unixfsv2.md
+++ b/data-structures/unixfsv2.md
@@ -1,0 +1,68 @@
+# UnixFSv2 Schema
+
+This doc contains the File and Directory schemas for UnixFSv2.
+
+This schema depends on some Advanced Layouts and types defined
+in other schemas:
+
+* [Data](https://github.com/ipld/specs/pull/211)
+* [Hashmap](https://github.com/ipld/specs/blob/master/data-structures/hashmap.md)
+
+```sh
+type DirData union {
+  | EntryMap "map"
+  | Hamt "hamt"
+} representation keyed
+
+type EntryMap {String:EntryUnion}
+
+type EntryUnion union {
+  | File "file"
+  | &File "fileLink"
+  | Directory "dir"
+  | &Directory "dirLink"
+} representation keyed
+
+# WIP: Hashmap does not yet have an advanced layout definition.
+# type Hamt {String:FileUnion}<HAMT>
+
+# advanced HAMT {
+#   implementation "IPLD/experimental/HAMT/v1"
+# }
+
+type Directory struct {
+  name optional String
+  size optional Int
+  data DirData
+}
+
+type Permissions struct {
+  uid Int
+  gid Int
+  posix Int # The standard 0777 bitpacking masks
+
+  sticky Bool (implicit "false")
+  setuid Bool (implicit "false")
+  setgid Bool (implicit "false")
+}
+
+type Attributes struct {
+  mtime optional Int
+  atime optional Int
+  ctime optional Int
+  mtime64 optional Int
+  atime64 optional Int
+  ctime64 optional Int
+
+  permissions optional Permissions
+
+  devMajor optional Int
+  devMinor optional Int
+}
+
+type File struct {
+  name optional String
+  data optional Data
+  size optional Int
+}
+```


data-structures/unixfsv2.md
```diff
@@ -0,0 +1,68 @@
...
+This schema depends on some Advanced Layouts and types defined
+in other schemas:
+
+* [Data](https://github.com/ipld/specs/pull/211)
+* [Hashmap](https://github.com/ipld/specs/blob/master/data-structures/hashmap.md)
+
+```sh
+type DirData union {
+  | EntryMap "map"
+  | Hamt "hamt"
```
rvagg, line 14: root node becomes inlined here, maybe this should be `&Hamt`? (also still hate using the specific algorithm name but know I'm alone on this so won't push).

data-structures/unixfsv2.md
```diff
@@ -0,0 +1,68 @@
...
+type EntryMap {String:EntryUnion}
+
+type EntryUnion union {
+  | File "file"
+  | &File "fileLink"
+  | Directory "dir"
+  | &Directory "dirLink"
+} representation keyed
+
+# WIP: Hashmap does not yet have an advanced layout definition.
```
rvagg, line 26: The intention of the current schema spec is that you can just do this:

```
advanced HAMT
type Hamt {String:FileUnion} representation advanced HAMT
```

Although I believe `s/FileUnion/EntryUnion` in the current iteration.


What else do we need to proceed with baking this in?

data-structures/unixfsv2.md
```diff
@@ -0,0 +1,68 @@
...
+  atime64 optional Int
+  ctime64 optional Int
+
+  permissions optional Permissions
+
+  devMajor optional Int
+  devMinor optional Int
+}
+
+type File struct {
```
rvagg, line 63: Is it intended that there will be an additional constraint in the code implementing this that `File` should have _at least `name` or `data`_? What's the use case for optional `data`?

data-structures/unixfsv2.md
mikeal: isn’t the root node usually pretty small? why wouldn’t we want to inline it?

data-structures/unixfsv2.md
mikeal: i had assumed that the library you publish would already have `advanced HAMT` in it because you need something to attach the implementation of the advanced layout to before handing it off to a consumer. i would then consume that w/ `type DirHamt {String:FileUnion} representation advanced HAMT`

data-structures/unixfsv2.md
mikeal: at some point we’re going to need to bike shed on how to represent sym links and a few other fancy types. *my* preference would be to actually have them be `File` types with special properties and no `data`, but there will no doubt be a longer thread about that. `name` is always optional because the name is usually provided by the directory key you read the file out of.

data-structures/unixfsv2.md
rvagg: root node will end up being about the same size as the intermediate node and could be roughly in line with the average leaf node size depending on various factors. The root node is just a special intermediate node that's at depth=0.

data-structures/unixfsv2.md
rvagg: Yeah, that would work too, this is a really fuzzy area (as partly demonstrated by #208). I've been imagining `advanced HAMT` as a signalling mechanism that "this thing exists somewhere and will need to be hooked up" and that comes from the thinking that also included `advanced HAMT { implementation "something" }`. It bothers me that we don't have a good answer to where these things sit and what's required where.

So I suppose to be maximally minimal you could omit `advanced HAMT`.

When I was doing up the Filecoin spec docs (have since deleted that work so I can't show it), I found that having `advanced HAMT; advanced Sharray` were useful documentation tools, saying "you'll encounter these things as you work through the rest of the docs, when you do, come back here and you'll understand that we're pointing to some adjacent space that has an implementation of these things. Like a named break-point in the schema that can be coupled with its own documentation. Not quite as helpful here because you only have one instance and could put the docs in that one place to explain what that is.

data-structures/unixfsv2.md
mikeal: ya, i think we should just inline it then. the rest of the directory metadata is so minimal that it doesn’t make a ton of sense to have it in a separate block.

mikeal:
Closing, working on very different looking unixfsv2 now.

PR #222, opened 2019-11-13
gh-pages publishing

rvagg:
Sitting on top of #216 so most of the changes to schemas/ aren't relevant (there are some minor changes dealing with `<pre>` tags).

This does two things, found in the two commits at the head of this branch:

* Use [Vuepress](https://vuepress.vuejs.org) for compiling a static site. You can now `npm install && npm run build` here to generate a static site (which ends up in html/.vuepress/dist). (fwiw I started playing with Hugo cause that's what others in PL are using but the yak shaving was too much, vuewpress is actually really nice, and quite comfortable for a JS-native).
* Use GitHub Actions to publish that static content to the `gh-pages` branch. This'll need a deploy key and a secret key set in this repo's settings for it to work.

For now the root, `/specs/` has a simple index that says what this is just points to `/specs/schemas/`, thats in .vuepress/README.md. It manually copies in the /schemas/ directory before compiling so the other spec docs are ignored but we could add them in over time and expand this.

It adds syntax highlighting to `ipldsch` code blocks, which is really nice to look at.

"Edit this page on GitHub" is on most pages and jumps you straight into the GitHub editing window.

I had to add a hack for .ipldsch and .json files: during the build step they get converted to .md files with code fences around them, so they render nicely. But that means rewriting links that point to them and also makes the "Edit this page on GitHub" too difficult for them.

Preview available at: https://rvagg.github.io/specs/


.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
...
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+    - run: npm install
+    - run: npm run build
+
+    - name: GitHub Pages action
+      uses: peaceiris/actions-gh-pages@v2.5.1
+      env:
+        ACTIONS_DEPLOY_KEY: ${{ secrets.ACTIONS_DEPLOY_KEY }}
```
mikeal, line 29: What is the deploy key for? Does it just do the git checkin/push?

Every action gets a new GitHub token minted that can use to do the checkin. See https://github.com/mikeal/publish-to-github-action/blob/master/entrypoint.sh

Update: just to be clear, you still need a line like this to env exposing the secret, but if you use `secrets.GITHUB_TOKEN` you don’t actually need to put anything in the secret store, this will be created for you by Actions and is nicely limited w/ a quick expiration.

rvagg: just blindly following instructions on this one https://github.com/peaceiris/actions-gh-pages#1-add-ssh-deploy-key
will try it on my account

mikeal: Oh I see what this is, https://github.com/marketplace/actions/github-pages-action#%EF%B8%8F-github_token

Funny enough, I was chatting with someone from GitHub about this today. I think you should just be able to use their GITHUB_TOKEN feature instead. The limitation they are running into is that pushes to master **from other GitHub Actions** won’t trigger push actions, which isn’t something we’re worried about.

I’ll comment in the other thread about the fix for their issue, but I think we can probably just switch to GITHUB_TOKEN.

rvagg: ah, there's an ongoing issue with `GITHUB_TOKEN`: https://github.com/peaceiris/actions-gh-pages/issues/9

the problem is with an action triggering an action, using `GITHUB_TOKEN` won't trigger a gh-pages build on their end, but a custom deploy key will. It works when your repo is private apparently but not public.

https://github.community/t5/GitHub-Actions/Github-action-not-triggering-gh-pages-upon-push/td-p/26869

Perhaps something they'll address, but for now, it seems this is necessary.

mikeal: i think i actually interpreted the issue wrong. github’s own internal infra that triggers the ghpages build won’t fire. that is *super* annoying ;)

rvagg:
#216 merged, this should be easier to review now, only 2 commits

.github/workflows/main.yml
mikeal: ok, i setup this deploy key and secret in the repo, it should be ready to go.

vmx (commented):
I prefer local paths over absolute paths for static websites, so that you can easily run them locally (often you even won't need a webserver). I don't know if that's possible with vuepress. I at least couldn't get it work locally (I also tried starting an http server within `html/.vuepress/dist`).

mikeal:
go ahead and add a CNAME file for `specs.ipld.io`, should work https://github.com/protocol/infra/issues/503

rvagg:
@vmx yeah, I'd much prefer relative URLs too for the flexibility but vuepress doesn't allow this as far as I can make out. I meant to add a `serve` target so I've done that now in the latest addition, `npm run serve` and you'll get a web server running to view the pages.

rvagg:
Following the rabbit hole through https://github.com/vuejs/vuepress/issues/796 and on to various other issues in vue itself and html-webpack-plugin, it looks like (a) it's currently impossible to do relative URLs and (b) it might be possible in the not too distant future once dependency updates flow down. There's a [change in vue](https://github.com/vuejs/vue/pull/8177) that hasn't been published that should help a lot. So long-term, we should make this portable. Short-term let's just make it work and get it up.

For now, `npm run serve`.

I've added a CNAME file with `specs.ipld.io` and changed the publish base to `/`

I've also added another change that replaces all of the `../` URLs in /schemas/ docs to absolute URLs back to this repo. Since we are only publishing /schemas/ for now, those don't work on the web. It can be undone when we publish more stuff.

vmx (approved):
Thanks, now it works locally. That surely is good enough, I just wanted some easy way to view it locally.

warpfork (approved):

PR #229, opened 2019-12-20
stub for car file format

mishmosh:
- Started a markdown file in the block header section (is that the correct folder?)
- Added diagram that @vmx wants to refer to
- Explicitly added "open problems" and "header format" sections

Also note: did _not_ add to README list yet.


rvagg:
Hey @mishmosh I've been [referring](https://github.com/rvagg/js-datastore-car) to CAR as "Content ARchive" or "Content Addressable aRchive" and https://github.com/ipfs/go-car has "A content addressible archive utility" as its description. I'm happy to switch to "Certified Archive" but is this a consensus name that's used by others? And if not, should I use these docs to try and drive a consensus on the name (I'd probably choose "Content Addressable aRchive")?

jbenet:
quick thoughts on name (thanks @mishmosh  for the ping):

- I think Content Addressed aRchive works (second name), and gets part of the way there to explaining the properties 
  - unfortunately not all content-addressing is secure hash linked addressing, or it would imply most/all. maybe some day most people will only think hash-linking when thinking content-addressing, but we're not there yet, as some communities use content-addressing modes which are not hash-linked
- I think Content ARchive does not work, as content does not imply content addressing, hash-linking, or security of any kind, and does not distinguish it from other archive formats, which also archive content. a tarball is an archive of content as any other. Content ARchive is only novel/different and not a very useful name (this is likely to be a shorthand name though, so we'll have to refer to it somewhere)
- I think Certified ARchive (first name) works too, and also gets part of the way there, it follows prior cryptography and crypto-systems nomenclature. 
  - especially if the CAR header (as originally hoped for) has room to carry an optional digital signature over the header hash (plus the associated public key chain up to some root (eg root CA cert))
  - alternative name previously considered: SAR -- Secure ARchive, but discarded in light of the car pun :D, and the inevitable self-fulfilling prophesy of naming something "Secure X" :S
- i'm fine with either Content Addressed Archive and Certified ARchive as a name
  - i like both a lot :), slight pref to "Content Addressed aRchive"
  - i would encourage us to keep the other name in the readme, as an "alternate name" as some  utilities sometimes do

PR #232, opened 2020-01-06
Mention prior art

fulldecent:


rvagg:
Hey, these are really interesting links (although I'm not sure if p2p sip is as relevant). I don't know if there's a clear enough limiting principle on listing prior art for such a section not to accumulate a continuous list of links for anything that comes before IPLD. If there's an explicit, or even implicit, claim that IPLD is somehow blank-slate work that invents new ideas without leaning on prior art then we should certainly fix that! Otherwise, I'd take it as implied that IPLD builds on many decades of prior art, just like most other things in our industry. So I think I'll 👎 this for now, although I'm appreciative of the links to docs that I haven't personally read before.

PR #239, opened 2020-02-06
Add initial specification for selector syntax.

creationix:
This is a proposal for a selector syntax that closely models the semantics already in the IPLD format of selectors.  The rational and constraints for the design are included in the documents as well as many examples and hopefully enough description of lexing/parsing behavior to make it unambiguous.

I would love feedback on what you like about this, what drives you crazy, and hopefully find out if this is a good direction.

### DIFF
--- a/selectors/selector-syntax.md
+++ b/selectors/selector-syntax.md
@@ -96,18 +96,6 @@ This is the default style for maximum terseness.  It minifies everything possibl
 f'characters'f'kathryn-janeway'f'birthday'f'year'.
 ```
 
-#### Using multi-segment-path extension.
-
-We can add some syntax sugar extensions as macros for common use cases as they come up.
-
-```ipldsel
-p'characters/kathryn-janeway/birthday/year'.
-```
-
-This example is sugar for a deeply nested path using a new `"paths"`/`"p"` keyword.  It's a version of `"f"` that splits its input on `/` and creates nested selectors.
-
-This can be done as a macro-like sugar at this level.  Another option is for the selector spec to add `ExplorePaths` as a new selector with optimized representation and execution.
-
 ### Getting a certain number of parent blocks in a blockchain
 
 This is based on [this sample](example-selectors.md#getting-a-certain-number-of-parent-blocks-in-a-blockchain).
@@ -132,18 +120,6 @@ fields('parent'(
 
 # Short Form
 f'parent'f'parent'f'parent'f'parent'f'parent'.
-```
-
-Shorter using paths extension:
-
-```ipldsel
-# Long Form
-paths('parent/parent/parent/parent/parent'(
-  match
-))
-
-# Short Form
-p'parent/parent/parent/parent/parent'.
 ```
 
 #### Parents Using Recursion
@@ -277,16 +253,11 @@ Some of the selectors accept literal values as parameters.  These are currently 
 
 #### Integers
 
-Integers can be encoded using common integer literals found in programming languages:
+Positive integers can be encoded using base 110:
 
 ```
 123 # Decimal
-0xfeed # Hex
-0o644 # Octal
-0b1001010 # Binary
 ```
-
-Note these are case insensitive for both the hex digits as well as the base prefixes.
 
 #### Strings
 
@@ -300,8 +271,6 @@ strings'
 'two'_'strings'
 ```
 
-Technically, any character possible in the selector itself (likely unicode) is allowed in here.  Also this means we cannot have two strings appearing next to eachother without a `_` separator between them.  Currently this never happens since no selector accepts two strings in a row.
-
 #### Maps
 
 We need to be able to encode the keys for the `fields` selector.  This is done using multiple string literals followed by nested contents.
@@ -312,21 +281,6 @@ fields(
   'bar'(...)
 )
 ```
-
-#### Bytes
-
-Selectors don't yet use bytes, but should ExploreFields be modified to allow binary keys, we are specifying a way to encode bytes.
-
-This is done using hex pairs surrounded by `!` characters.
-
-```
-!48 65 6c 6c 6f!
-!48656c6c6f!
-!48656
- c6c6f!
-```
-
-Notice that whitespace is ignored.
 
 ### Whitespace and Comments
 
--- a/selectors/selector-syntax.md
+++ b/selectors/selector-syntax.md
@@ -204,14 +204,16 @@ This mode-less flexibility, combined with tools to automatically translate in bu
 
 Whitespace is completely ignored by the parser except for inside quoted strings.
 
-When extending this in the future, be aware that whitespace cannot be used as keyword boundaries (`"ab cd"` is identical to `"a bc d"`).
-We should have enough space for dozens of long and short names, but will want to write a tool to automatically look for ambiguities as well as improve developer experience with auto formatters and smart highlighters.
+Line comments are also ignored even if they contain things that look like quoted strings
 
 ### Parentheses are Usually Optional
 
 Parentheses annotate structure and are sometimes required for ambigious cases such as unions which contain an arbitrary number of selectors or selectors with optional parameters of conflicting types.
 
-However the parser can usually infer the structure without them because most selectors have a fixed or semi-fixed arity and certain types are only allowed at certain places.
+For example `union(union match match)` is interpreted as `union(union(match match))` and not `union(union(match) match)` because in the first, the last match will be part of the inner union and not the outer union.  When minimizing, some parentheses might be kept to preserve semantics.
+
+- `uu..` -> `{ selector: { '|': [ { '|': [ { '.': {} }, { '.': {} } ] } ] } }`
+- `uu(.).` -> `{ selector: { '|': [ { '|': [ { '.': {} } ] }, { '.': {} } ] } }`
 
 The best practice (and what the default formatting styles will enforce) is for human readable selectors to use parentheses liberally while URL embedding style will only contain the required ones.
 
@@ -244,31 +246,40 @@ Best practice is to annotate `limit`, but not `sequence` for human readable, and
 # Human Readable
 recursive(limit=5 ...)
 # URL Embeddable
-R(5...)
+R5...
 ```
 
 ### Literal Values
 
-Some of the selectors accept literal values as parameters.  These are currently `String`, `{String:Selector}`, and `int`.
+Some of the selectors accept literal values as parameters.  These are currently `String`, `{String:Selector}`, and `Int`.
 
 #### Integers
 
-Positive integers can be encoded using base 110:
+Integers can be encoded using base 10 with optional leadin sign:
 
 ```
 123 # Decimal
+-123 # Negative decimal
 ```
 
 #### Strings
 
-Strings are quoted using single quote, they can be escaped using double single quote.  You can include non url-safe characters between the quotes, but will need to escape the entire selector properly when embedding in a URL.
+Strings are quoted using single quote.  They can contain any characters including newlines and unicode characters.  The following characters can be escaped using backslash (`\`) followed by a special character.  If the backslash is followed by a character not in the list, it's considered a syntax error.
+
+- \b  Backspace (ascii code 08)
+- \f  Form feed (ascii code 0C)
+- \n  New line
+- \r  Carriage return
+- \t  Tab
+- \'  Single quote
+- \\  Backslash character
 
 ```
 'Hello World'
-'It''s a lovely day'
+'It\'s a lovely day'
 'Multiline
 strings'
-'two'_'strings'
+'Multiline\nstrings'
 ```
 
 #### Maps
@@ -291,27 +302,26 @@ A comment starts at `#` and ends at end of line.
 Parser Specification
 --------------------
 
-### String and Comment Modes
+# Initial Stripping
 
-The selector text is normally treated initially as a stream of characters.  For purposes of parsing, strings and comments create modal changes to the rules.
-
-- When in normal mode:
-  - Finding `"'"` changes to string mode.
-  - Finding `"#"` changes to comment mode. (Also discard it).
-  - Discard whitespace, defined as `"\r"`, `"\n"`, `"\t"`, and `" "`.
-- When in string mode:
-  - Finding `"'"` changes back to normal mode.
-  - Preserve all characters.
-- When in comment mode:
-  - Finding newline changes back to normal mode.
-  - Discard all characters.
-
-If comments and strings overlap, whichever comes first is the correct mode:
+The parser must act as is there was an initial pass that removed all whitespace not inside strings and all line comments.
 
 ```ipldsch
 # This is a comment 'this is not a string'
 'This # is # a string' this is normal
 this is also normal
+
+# Keywords get merged
+hello world
+helloworld
+
+# Comments get stripped
+'a string' # and comment
+'a string'
+
+# Strings inside comments are still comments
+empty # a comment with a 'string'
+empty
 ```
 
 ### Identifier Tokenization
@@ -321,27 +331,6 @@ The parser knows a fixed set of built-ins to look for.  This is the long and sho
 ```ipldsel
 # This will match `fields` first and not even try `f`.
 fields...
-```
-
-### Number Tokenization
-
-Numbers are tokenized similar to the identifier method. If a single zero is followed by `x`, `o`, or `b` and then one or more digits belonging to that base, it will be tokenized as that base.  Otherwise it will be a zero.  Normal decimal numbers are also parsed greedily.
-
-For example:
-
-```ipldsel
-123   # this is 123
-0xdeg # this is 0xde or 222 with `g` leftover to tokenize.
-0123  # this is 123
-```
-
-### String Tokenization
-
-Strings are tokenized simply by switching modes based on the presense of `"'"` characters.  We enable quote escaping with a rule that whenever two string literals are next to eachother, they are combined into a single string with a single quote inserted between them.
-
-```ipldsel
-'I am a string'      # "I am a string"
-'I''m a string too'  # "I'm a string too"
 ```
 
 ### Parentheses and Parse Order
@@ -363,7 +352,7 @@ We could have added parentheses to this, but they were not needed since the defa
 fields('fieldname'(match))
 ```
 
-When parentheses are added, it sets constraints on what level tokens live on.  It goes up with every `"("` and down with every `")"`.  All parameters to a single consumer must have the same nesting level or they don't match.
+When parentheses are added, they can override the default greedy behavior in some otherwise ambigious cases.  Again, the example ``union union match match`` is not `union(union(match) match)` but is `union(union(match match))` because the innermost union gets to greedily match first.  Extra parentheses can be added as `union union(match) match)`, or in short form`uu(.).` vs `uu..`
 
 Known issues
 ------------
--- a/selectors/selector-syntax.md
+++ b/selectors/selector-syntax.md
@@ -37,7 +37,7 @@ For example, this selector simulates a git shallow clone by recursively walking 
 
 ```ipldsel
 # Starting at the commit block.
-R5f'tree'R*~'parents'*~
+R5f'tree'Rn*~'parents'*~
 ```
 
 #### Human Friendly
@@ -54,7 +54,7 @@ The exact same selector for git shallow clone from above can also be written in 
 recursive(limit=5
   fields(
     'tree'(
-      recursive(
+      recursive(limit=none
         all(recurse)
       )
     )
--- a/selectors/selector-syntax.md
+++ b/selectors/selector-syntax.md
@@ -10,7 +10,7 @@ Introduction
 
 *Prerequisites: [Selectors](https://github.com/ipld/specs/blob/master/selectors/selectors.md).
 
-IPLD Selectors are represented as IPLD data nodes.  This is great for embedding them in a structured way, but authoring them or viewing them in this format isn't the easiest.  This Syntax provides a textual DSL for reading/writing selectors in a more text friendly format.
+IPLD Selectors are represented as IPLD data nodes.  This is great for embedding them in a structured way, but authoring them or viewing them in this format isn't the easiest.  This syntax provides a textual DSL for reading/writing selectors in a more text friendly format.
 
 Tooling can be used to convert between formats and even various styles optimized for the use-case at hand.
 
@@ -48,7 +48,7 @@ This means it should be terser than the JSON or YAML representations of the IPLD
 
 This means it should allow flexibility with whitespace as well as allowing optional symbols and annotations to make structure easier to see visually.
 
-The exact same selector for gt shallow clone from above can also we written in the following form using this exact same syntax: (This is not another mode, it's the same syntax):
+The exact same selector for git shallow clone from above can also we written in the following form using this exact same syntax: (This is not another mode, it's the same syntax):
 
 ```ipldsel
 recursive(limit=5
@@ -330,7 +330,7 @@ Notice that whitespace is ignored.
 
 ### Whitespace and Comments
 
-Comments are allowed in this syntax and will be preserved by auto-formatters wen possible, but will be stripped when converting to URL style and are not included in the IPLD representation of the selector.
+Comments are allowed in this syntax and will be preserved by auto-formatters when possible, but will be stripped when converting to URL style and are not included in the IPLD representation of the selector.
 
 A comment starts at `#` and ends at end of line.
 
--- a/selectors/selector-syntax.md
+++ b/selectors/selector-syntax.md
@@ -1,0 +1,430 @@
+Specification: IPLD Selectors Syntax
+=============================
+
+**Status: Prescriptive - Draft**
+
+Introduction
+------------
+
+### Motivation - What is Selectors Syntax
+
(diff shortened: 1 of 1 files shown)


selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+**Status: Prescriptive - Draft**
+
+Introduction
+------------
+
+### Motivation - What is Selectors Syntax
+
+*Prerequisites: [Selectors](https://github.com/ipld/specs/blob/master/selectors/selectors.md).
+
+IPLD Selectors are represented as IPLD data nodes.  This is great for embedding them in a structured way, but authoring them or viewing them in this format isn't the easiest.  This Syntax provides a textual DSL for reading/writing selectors in a more text friendly format.
```
vmx: typo: "This Syntax" -> "This syntax"

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+
+#### Human Friendly
+
+Selector syntax should be easy to read/author by humans.
+
+This means it should be terser than the JSON or YAML representations of the IPLD data, but still verbose enough to have meaningful structure and keywords/symbols.
+
+This means it should allow flexibility with whitespace as well as allowing optional symbols and annotations to make structure easier to see visually.
+
+The exact same selector for gt shallow clone from above can also we written in the following form using this exact same syntax: (This is not another mode, it's the same syntax):
```
vmx: typo: "gt" -> "git"

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+!48656c6c6f!
+!48656
+ c6c6f!
+```
+
+Notice that whitespace is ignored.
+
+### Whitespace and Comments
+
+Comments are allowed in this syntax and will be preserved by auto-formatters wen possible, but will be stripped when converting to URL style and are not included in the IPLD representation of the selector.
```
vmx: typo: "wen" -> "when"

vmx (commented):
Wow, that is really a good read. My comments are only in regards to typos the rest sounds great.

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+```js
+[ '!', "'", '(', ')', '*', '-', '.', '0', '1',
+  '2', '3', '4', '5', '6', '7', '8', '9', 'A',
+  'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J',
+  'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S',
+  'T', 'U', 'V', 'W', 'X', 'Y', 'Z', '_', 'a',
+  'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j',
+  'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's',
+  't', 'u', 'v', 'w', 'x', 'y', 'z', '~']
+```
```
ribasushi, line 32: "Easy embedding inside URLs" implies "easy visual skimming" ( perhaps with some initial training needed, just like e.g. regular expressions ). Assuming a person reading this is proficient: are we comfortable with a case sensitive, visually-collidable character set? 

I am not particularly leaning one way or the other, but rather am bringing the point up for discussion .

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+When extending this in the future, be aware that whitespace cannot be used as keyword boundaries (`"ab cd"` is identical to `"a bc d"`).
+We should have enough space for dozens of long and short names, but will want to write a tool to automatically look for ambiguities as well as improve developer experience with auto formatters and smart highlighters.
+
+### Parentheses are Usually Optional
+
+Parentheses annotate structure and are sometimes required for ambigious cases such as unions which contain an arbitrary number of selectors or selectors with optional parameters of conflicting types.
+
+However the parser can usually infer the structure without them because most selectors have a fixed or semi-fixed arity and certain types are only allowed at certain places.
+
+The best practice (and what the default formatting styles will enforce) is for human readable selectors to use parentheses liberally while URL embedding style will only contain the required ones.
```
ribasushi, line 218: I am not sure I fully understand this / the motivation for `R`...`~` and other scope-pairs is not entirely clear. Could you elaborate  "even harder" why you didn't go with a more traditional balanced pair of characters e.g. `(` `)` ? Especially jarring is the discrepancy of doing couple paragraphs down:
```
# URL Embeddable
R(5...)
```

ribasushi (changes requested):
Marking "request changes" as stand-in for "request discussion". Will d another pass over this once the first two pieces are clarified

Awesome work as a whole!

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+Whitespace is completely ignored by the parser except for inside quoted strings.
+
+When extending this in the future, be aware that whitespace cannot be used as keyword boundaries (`"ab cd"` is identical to `"a bc d"`).
+We should have enough space for dozens of long and short names, but will want to write a tool to automatically look for ambiguities as well as improve developer experience with auto formatters and smart highlighters.
+
+### Parentheses are Usually Optional
+
+Parentheses annotate structure and are sometimes required for ambigious cases such as unions which contain an arbitrary number of selectors or selectors with optional parameters of conflicting types.
+
+However the parser can usually infer the structure without them because most selectors have a fixed or semi-fixed arity and certain types are only allowed at certain places.
```
warpfork: I'm a _little_ a'twitch about this part about "fixed or semi-fixed arity".

It's true that in the Schema for selectors, most things have fixed "arity" and any optionals are clearly demarcated.  However, it's also the case that we have some idea of how migrations within the Schema system will work if we add more fields or selector types.  Here, it seems it might be slightly less clear; migration rules that work for the Schema won't automatically translate to this DSL.

Maybe that's fine.  I suppose unless the DSL is exactly an IPLD codec (... in which case it's hardly delivering the kind of increased terseness that makes something earn the term "DSL" at all!), it's necessarily going to have different migration rules.  And maybe there's no major problem with that.

Just thinking outloud.

warpfork (commented):
I have no objections to this, I think :)

I'm also not really reviewing for ergonomics though, as I feel ill suited to do so without an application in my mind's eye, which I'm pretty sparse on.  And thus, "no objections" is about the greenest light I'm likely to give, if that makes sense. :P

selectors/selector-syntax.md
creationix: Agreed, I'm not settled on this part yet either.  I'm currently going through and implementing a full parser.  I feel I'll get a much better sense of where it's ambiguous and be able to more clearly describe the grammar.

selectors/selector-syntax.md
creationix: I'm not sure what you mean by pairing up `R` and `~`.  They are not necessarily balanced; there could technically be multiple recursive edges inside a single recursive node.  It's not a scope pair, but rather two distinct node types in the selector AST that happen to relate to eachother.

Also I use `(` and `)` already in the syntax to help with forcing parameters to the correct nodes.

selectors/selector-syntax.md
creationix: This is certainly something we can use as an added constraint to consider when choosing the characters used for short form.  Currently, it only uses `f`, `i`, `r`, `u`, `c`, `F`, `*`, `.`, and `~`.

The listing of url safe characters is more of a technical constraint about what ASCII characters can be embedded in url components without needing to be escaped.

creationix:
After spending a day implementing a full parser, I've discovered the white-space rules are more interesting than I initially thought.  The design goal was to make white-space 100% irrelevant to the parser.  Most programming languages claim to not have significant white-space, but that's not entirely true for any real language.

The first exception obviously is white-space within strings needs to be preserved.  This is easy enough, I simply turn off white-space stripping when within quoted sections.

The second case is one you typically don't realize.  Most languages use white-space as token separators.  For example:
```
# Escaped quotes
'This is ''one'' string with escaped quotes.'
# Escaped quotes or two strings?
'is this one string?' 'or two strings?`
```

The parsing of the second example *depends* on that space being between the two string literals.  If we really ignore white-space, then it should parse the same if the space is removed.

Currently, this syntax really means it when it says no white-space, and the second will be a single string.

Another example is identifiers.  In most languages `foo bar` is clearly two identifiers while `foobar` is one.  In this language, we ignore the space entirely and both look identical to the parser. It needs other clues to know where one identifier/selector starts and where another ends.  This is accomplished by the language having a fixed set of identifiers and careful design to make sure that there is never ambiguity.

For example we should avoid having a `foo` selector with short form `f` if there is also another selector who's short form is `o`.  A developer may write `foo` and we might interpret it as `f`, `o`, `o`, or they might write `f o o` and we interpret it as `foo`. 

This spec does specify which it should be (it would be always `foo` because it's longer).  But we also want to avoid these situations whenever possible since humans are not compilers and may have different expectations since most/all existing languages don't really ignore white-space.

selectors/selector-syntax.md
ribasushi: > They are not necessarily balanced; there could technically be multiple recursive edges inside a single recursive node. 

This might **very** well be my lack of understanding of the problem domain. Let's cover this during the ipld meet if time permits.

creationix:
@warpfork One solution to the potential confusion with merged identifiers would be to preserve the white-space as tokens and tell the parser about them.  But the problem with this is it would require those spaces to be preserved in compact mode.  This could almost double the length of minimized version and include lots of spaces which can sometimes be problematic in URLs.

I'll consider it further though.

selectors/selector-syntax.md
ribasushi: We chatted about this a bit more - I now understand what was meant by the above quote. I withdraw the question about balanced parens ;)

warpfork:
Definitely meant my remark on that as a "2 cents".  You've probably already thought about it much more than I have.

creationix:
The initial lexer implementation is now done.  The description in the spec for parsing out identifiers seems to be working. https://github.com/creationix/sel-parse-zig/blob/887c3628e11b4ff751e68a9ade4c18a9bf4daf25/src/lexer.zig

In particular, there is a hard coded list of identifiers expected here:
```zig
// Sorted by longest first, then lexical within same length.
const identifiers = .{
    "condition",
    "recursive",
    "recurse",
    "fields",
    "index",
    "match",
    "range",
    "union",
    "all",
    ".",
    "*",
    "~",
    "c",
    "f",
    "i",
    "r",
    "R",
    "u",
};
```
And then parsing those ends up being quite straightforward.
```zig
// Tokenize Identifiers
inline for (identifiers) |ident| {
    var matched = true;
    var i: u32 = 0;
    while (i < ident.len) {
        if (i >= input.len or ident[i] != input[i]) {
            matched = false;
            break;
        }
        i += 1;
    }
    if (matched) return Token{ .id = .Identifier, .slice = input[0..i] };
}
```

creationix:
To get an idea of what the lexer tokens look like, this is the output of the following:

```ipldsel
recursive(limit=5
  fields(
    'tree'(
      recursive(
        all(recurse)
      )
    )
    'parents'(
      all(recurse)
    )
  )
)
```

```zig
0       Id.Identifier   9B      `recursive`
9       Id.Open 1B      `(`
10      Id.Unknown      1B      `l`
11      Id.Identifier   1B      `i`
12      Id.Unknown      1B      `m`
13      Id.Identifier   1B      `i`
14      Id.Unknown      2B      `t=`
16      Id.Decimal      1B      `5`
17      Id.Whitespace   3B      `
  `
20      Id.Identifier   6B      `fields`
26      Id.Open 1B      `(`
27      Id.Whitespace   5B      `
    `
32      Id.String       6B      `'tree'`
38      Id.Open 1B      `(`
39      Id.Whitespace   7B      `
      `
46      Id.Identifier   9B      `recursive`
55      Id.Open 1B      `(`
56      Id.Whitespace   9B      `
        `
65      Id.Identifier   3B      `all`
68      Id.Open 1B      `(`
69      Id.Identifier   7B      `recurse`
76      Id.Close        1B      `)`
77      Id.Whitespace   7B      `
      `
84      Id.Close        1B      `)`
85      Id.Whitespace   5B      `
    `
90      Id.Close        1B      `)`
91      Id.Whitespace   5B      `
    `
96      Id.String       9B      `'parents'`
105     Id.Open 1B      `(`
106     Id.Whitespace   7B      `
      `
113     Id.Identifier   3B      `all`
116     Id.Open 1B      `(`
117     Id.Identifier   7B      `recurse`
124     Id.Close        1B      `)`
125     Id.Whitespace   5B      `
    `
130     Id.Close        1B      `)`
131     Id.Whitespace   3B      `
  `
134     Id.Close        1B      `)`
135     Id.Whitespace   1B      `
`
136     Id.Close        1B      `)`
```

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+```ipldsel
+# Starting at the commit block.
+R5f'tree'R*~'parents'*~
+```
+
+#### Human Friendly
+
+Selector syntax should be easy to read/author by humans.
+
+This means it should be terser than the JSON or YAML representations of the IPLD data, but still verbose enough to have meaningful structure and keywords/symbols.
```
rvagg, line 47: minor style suggestion: collapse these two paragraphs into a bullet-point list hanging off the first sentence:

> ... easy to read/author by humans. This means it should:
> * Be terser ...
> * Allow flexibility ...

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+  )
+)
+```
+
+Examples
+--------
+
+### Deeply Nested Path
+
+Based on [this example](https://github.com/creationix/specs/blob/selector_text_syntax/selectors/example-selectors.md#deeply-nested-path).
```
rvagg: remove leading part of this URL, keep it relative, either `./example-selectors.md` or just `example-selectors.md`.

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+
+We can add some syntax sugar extensions as macros for common use cases as they come up.
+
+```ipldsel
+p'characters/kathryn-janeway/birthday/year'.
+```
+
+This example is sugar for a deeply nested path using a new `"paths"`/`"p"` keyword.  It's a version of `"f"` that splits its input on `/` and creates nested selectors.
+
+This can be done as a macro-like sugar at this level.  Another option is for the selector spec to add `ExplorePaths` as a new selector with optimized representation and execution.
```
rvagg: > "Another option is for ..."

I'm not sure that it's appropriate to put enhancement suggestions inline in this way; implementers and users reading this aren't going to find this helpful. So maybe find a way to split it out into a separate doc detailing possible future enhancements or just file an issue or even open a PR to add the change if you think it's important enough?

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+Written verbosely with parentheses, named parameters, and whitespace, it looks like this:
+
+```ipldsel
+recursive(
+    limit=5
+    sequence=...
+)
+```
+
+Depending on the context, we could omit the parentheses because the optional `stopAt` parameter is of type `Condition` and the parser likely expects something else after this node.
```
rvagg, line 239: so there aren't any selector forms with possibly ambiguous lengths so predicting a `stopAt` in the shortened selector syntax should be straightforward?
I see `Matcher` in the selector schema also has two optional fields, `onlyIf` and `label`, could these get in the way?

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+Integers can be encoded using common integer literals found in programming languages:
+
+```
+123 # Decimal
+0xfeed # Hex
+0o644 # Octal
+0b1001010 # Binary
+```
+
+Note these are case insensitive for both the hex digits as well as the base prefixes.
```
rvagg: I _think_ that only positive integer literals are used by selectors, so is it worth calling this fact out because `-` doesn't need to be handled. Also, the ints in the selectors schema that I see are only things like counts/limits and array indexes, does this make the non-base10 forms unnecessary complexity?

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+
+#### Strings
+
+Strings are quoted using single quote, they can be escaped using double single quote.  You can include non url-safe characters between the quotes, but will need to escape the entire selector properly when embedding in a URL.
```
rvagg: Is `'` going to be robust enough if we wanted to start escaping more? Say we wanted to handle `/` characters in fields but not have a selector confuse it with a path, `'` won't help us here I think, it's the kind of thing you'd use `\` for: `parent/parent/a\/b` where `"a/b"` is a map key at some node.

selectors/selector-syntax.md
```diff
@@ -0,0 +1,430 @@
...
+```ipldsel
+fields(
+  'foo'(...)
+  'bar'(...)
+)
+```
+
+#### Bytes
+
+Selectors don't yet use bytes, but should ExploreFields be modified to allow binary keys, we are specifying a way to encode bytes.
```
rvagg: because we only allow string map keys, so there should never be bytes fields, so unless there are conditions that look into field values, or we really want to leave the door open to non-string map keys, then I think this might be unnecessary?

rvagg:
I really like how the two forms are really just one form, with shortenings (and no comments for URL form), but we're trading that against ease of implementation though, so every environment that needs to be able to use these would need to have a custom parser written. I doesn't _seem_ difficult, but it's worth noting as we expand our language support (Go, JS and now Rust, but also Filecoin is using selectors and being implemented in C++ and maybe others too and this might be something they want at some point) that we're making that exchange.

selectors/selector-syntax.md
creationix: It's hard to say.  This is why I'm implementing syntax parsers.  So far I've not come across any concrete use cases where the parentheses are actually needed.

selectors/selector-syntax.md
creationix: Agreed, I'll remove it.  In another conversation we discovered that the current short form without the enhancement is just as terse and doesn't depend on semi-url-unsafe characters like `/`.

selectors/selector-syntax.md
creationix: I don't understand the limitation.  It can escape any string you want by simply doubling any `'` characters found in the original string.  If you want to include `\/` in the string, it's fine.

selectors/selector-syntax.md
creationix: My understanding was map keys didn't have to be strings.  But since current selectors only support strings for fields, I can remove this for now to simplify the initial version.

creationix:
@rvagg FWIW, I'm working on two concurrent implementations.  One in vanilla JS and one using zig-lang which can be compiled to webassembly or a C ABI library for use in virtually any language.  I plan to tweaking this spec with my findings from the two parsers to ensure it's not more difficult than it needs to be.

selectors/selector-syntax.md
creationix: If you're talking about embedding the selector in a url component, then the forward slash is already escaped at the url embedding layer.  It's not included in the safe character set because `encodeURIComponent("foo/bar")` is `foo%2Fbar`  This syntax assumes you'll use either `encodeURI` or `encodeURIComponent` when embedding a selector in such a situation.  We avoid the characters that would be escaped to reduce size bloat, but otherwise they don't pose any issues.

rvagg: I'm thinking about the case where we want _special_ characters, not just raw characters, in the same way that `\n` is a special character with many `\` escaping schemes. My point about `/` is hypothetical—imagine a case where we say that `'foo/bar/baz'` is parsed according to the rules you have here, it's a path shorthand. Right now, if you happen to have data with keys that contain the `/` character then paths aren't going to work nicely for you (bad luck, choose better keys buddy). But say we decide that it's important that we expand the range of characters that paths work for to include `/` characters _inside_ of keys, so we want to be able to switch between `/` for path separators and `/` as a character inside a key. A logical way of doing this is to escape it, turn it into something special that the parser has to treat differently ("oh, that's not a path separator because it's escaped, it must be part of the key string for this path segment"). How would you escape anything other than `'` when `'` is both your escaping character and your string terminator? If you have a key `b/a/z` at the end of your path, you can't: `'foo/bar/b'/a'/z'` because you've got termination confusion. If a character other than `'` was your escaping character then it's clearer: `'foo/bar/b\/a\/r'`.

creationix: I see, you want the ability to embed things like newlines and tabs in the string without using actual newlines and tabs.  In the current scheme, it's quite possible to include them, but it's going to look weird.
```
'this is a string
with a newline'
```
vs
```
'this is a string\nwith a newline'
```

creationix:
So while implementing multiple versions of the parser, I'm getting more and more convinced this spec needs to be more automated and formalized.  I'm now spiking on generating a syntax grammer that can be automatically derived from the selector schema directly.  The design will still be similar to what's proposed here, but it should be more consistent and less hand-crafted to make tooling across the board easier.

creationix:
Ok, the JS parser can now correctly compile all the sample selectors in this spec.  The new approach worked well.  Basically, I load the existing IPLD Schema for selectors and generate a parser from that.  In order to match the proposed syntax, the parser generator accepts a list of "aliases" for various types.  For example, this is the line in JS that generates the selector syntax parser:
```js
const schema = schemaParse(readFileSync(`${__dirname}/selectors.ipldsch`, 'utf8'))

const parseSelectorEnvelope = makeParser(schema, "SelectorEnvelope", {
  Matcher: ['match', '.'],
  ExploreAll: ['all', '*'],
  ExploreFields: ['fields', 'f'],
  ExploreIndex: ['index', 'i'],
  ExploreRange: ['range', 'r'],
  ExploreRecursive: ['recursive', 'R'],
  ExploreUnion: ['union', 'u'],
  ExploreConditional: ['condition', 'c'],
  ExploreRecursiveEdge: ['recurse', '~'],
  RecursionLimit_None: ['none', 'n'],
})
```

The types mentioned are existing in the schema, I'm creating a semi-automated DSL by specifying the entry point and long and short keywords for some types.

Note that this library could be used to create a DSL for *any* IPLD data structure that has a schema.

creationix:
I found a case where the parentheses are significant and can't always be removed when converting to short form.  Consider the following selector:
```sh
union(
  union(
    match
  )
  match
)
```

Since`union` (aka  `ExploreUnion`) contains a list (aka `[Selector]`), it consumes an arbitrary number of selectors.  If the parentheses are removed, the two matches will be put under the inner union.  Therefore the short form of this is:

```sh
# Correct short form
uu(.).

# Wrong short form
uu..
```

We could keep it simple and say the minimizer always preserves parentheses when encoding a list.  I don't even know of any real world use cases that use `ExploreUnion` and would love to see the actual use cases.  A smarter minimizer could do some basic analysis to know when the parentheses are required and only include them then.

creationix:
Another case where they are required is labeled matchers inside of fields.
```sh
# Fields with labels
fields(
  'with-label'(
    match('label')
  )
  'without-label'(
    match
  )
)

# Properly Minimized
f'with-label'(.'label')'without-label'.

# Another Properly Minimized
f'with-label'.('label')'without-label'.

# Broken minimized
f'with-label'.'label''without-label'.
```

This brings up another question about normalization of minimized form.  Are we OK with there being multiple correct short forms?  Does it matter?

creationix:
@rvagg, you were right!

Strings are problematic too.  The less common escaping method used for strings (two single quotes) works, but it also introduces cases where multiple token are merged.  For example, consider the following:

```sh
fields
  'foo'
    match 
      label='blue'
  'bar'
    match
```

This currently breaks because the `'blue'` label and the `'bar'` field are merged into a single label containing `"blue'bar"`.

```sh
Sample:

    fields
      'with-label'
        match 
          label='my-label'
      'another-field'
        match


SyntaxError: "fields'foo'matchlabel='blue''bar'match"
                                               ^ Unexpected extra syntax at 33
```

I propose we switch to a more traditional string syntax with backslash escaping.  It will add some bloat when url encoding, but overall should be an improvement.

selectors/selector-syntax.md
creationix: I went ahead and switched to a more traditional string literal syntax.  It's now single quote only, but otherwise very similar to JSON with backslash escaping for whitespace, backslash and single quote.  I'm leaving unicode escaping out of this since unicode values can simply be included as-is and the JSON form of `\uxxxx` is only 16 bit.

rvagg:
> This brings up another question about normalization of minimized form. Are we OK with there being multiple correct short forms? Does it matter?

I don't know specifically for this but we keep on finding cases elsewhere where we don't have one-single-way and this being a problem. I don't know how that would show up here, but maybe if someone chose to encode a selector string rather than a full selector as per schema then having more than one way to say the same thing might be a problem. There's a lot of byte-shavers around that will look at this work and look at the full selector schema and opt for a short string form.

Would it be a big deal to make a _must be simplest accurate representation_ rule that would give us one-single-way?

Re `ExploreUnion`, it's a selection strategy that is only useful when you have more than one selector to use isn't it? For practical purposes you should always have >1 item in the list, making parens always necessary and your example overly simple. There are going to be cases of poorly crafted selectors that misuse it, but that shouldn't be the normal case.

mikeal:
We’ve learned a lot from this but we’re not quite sure how we want to handle simplified string representations for selectors and paths. Closing for now.

PR #240, opened 2020-02-06
Add ExplorePath selector proposal.

creationix:
Add a new selector type that is optimized for the common use case of deeply nested field access.

This allows for a more compact serialization of the selector as well as giving implementations a place to optimize the runtime for this use case.


vmx:
I surely like the idea. But at the moment I'd like to keep Selectors as minimal as possible to make it easy to implement and also iterate in case we want to change things in the future (Selectors have gone through so many iterations already). Hence I'd prefer not merging this one.

creationix:
Fine with me.

PR #241, opened 2020-02-19
Library design recommendations.

warpfork:
Starting a new folder for this.  The content I'm imagining gathering
here isn't *exactly* specs, per se... but it's definitely content
that's worth gathering somewhere more centrally than in the comments
in the source code of any single implementing library and language.

First up: some remarks on Node and Kind, and how and why regarding Node
as an interface becomes systemically important.

I felt this is particularly important to write about because it's
quite non-obvious from the first couple of things a new library
author is likely to encounter first; it doesn't show up until you
start trying to implement some of the more advanced features...
but by then it may be too late to address without a painful refactor.
So knowing about it earlier is likely to save a great deal of work.

There's probably quite a lot of other content we can gather which is
useful recommendations for implementers, but not exactly "spec",
which can later flesh out this folder with more content;
this nodes-and-kinds doc is just the first thing I come up with.

### DIFF
--- a/design/libraries/README.md
+++ b/design/libraries/README.md
@@ -1,0 +1,10 @@
+Library Design Guidance
+=======================
+
+This directory contains some documentation of recommendations for
+library authors who want to make IPLD libraries in a new language
+(or, perhaps for readers who want to understand an existing library better).
+
+Some of the information expressed here comes down to opinions moreso than specification;
+what is good ergonomics may vary wildly per language, so take these as
+recommendations rather than strictures.
--- a/design/libraries/nodes-and-kinds.md
+++ b/design/libraries/nodes-and-kinds.md
@@ -1,0 +1,152 @@
+Nodes and Kinds
+===============
+
+Your IPLD library should have two key types:
+
+1. `Node`;
+2. `Kind`.
+
+`Node` should be an interface -- the membership should be open
+(aka, it should be possible for other packages to implement it).
+
+`Kind` should be an enumeration -- a fixed set of named members,
+which should not be extendable.
+
+
+Kind
+----
+
+`Kind` maps very directly onto the definition of
+[Data Model Kinds](../../data-model-layer/data-model.md#kinds).
+
+`Kind` does not include the Schema layer's concept of "struct", etc.
+
+`Kind` must be an enum, **and not a sum type**.  Attempting to implement
+kind as a sum type conflates it with `Node`.
+(This may be tempting to try to combine `Kind` and `Node` into a single
+sum type definition if you're only looking at the Data Model layer,
+but it is a mistake: both Schema types and Advanced Layouts require
+the ability to add more implementations of `Node`, so this conflation
+will cause cataclysmic problems and force a painful refactor
+as soon as you get to implementing those systems.
+See the [different implementors of Node](#different-implementors-of-node)
+section, later in this document, for more information on this.)
+
+
+Node
+----
+
+`Node` is a monomorphized interface for handling data -- in other words,
+we make all data look and act like a `Node`, so that we can write all of our
+functions against the `Node` interface, and have that work for any sort of data.
+
+`Node` has functions for examining any of the
+[Data Model Kinds](../../data-model-layer/data-model.md#kinds).
+For example, this means `Node` must be able to
+do a key lookup for a map kind,
+provide an iterator for a list kind,
+or be convertible to a primitive if it's a integer kind.
+
+`Node` is generally implemented by making an interface with the superset of all
+these methods needed for the various kinds of data.
+Some programming languages may also have a pattern-matching faculty which
+may make this nicer; feel free to use it (but mind the caveats issued in the
+[Kind](#kind) section above, and the
+[different implementors of Node](#different-implementors-of-node) section below:
+the membership of `Node` must remain *open*;
+you do *not* want to use a sum type with a closed list of concrete members here,
+or it will cause other roadblocks later that *will* force a redesign).
+For languages where this is most straightforwardly implemented by a single
+interface containing the superset of all necessary methods, many of the methods
+will error if the `Node` refers to information of the wrong kind for that method;
+this is fine.
+
+`Node` should be clear about what sets of methods are valid for acting on it.
+Typically, this is done by a `Node.Kind()` method, which should return
+a member of the [Kind](#kind) enum.
+This information is useful for anyone writing functions which use the `Node`
+interface, because it's much more pleasant (and fast) to check the Kind and
+know which methods can be expected to work than it is to have to probe every
+method individually for failure.
+(Again, programming languages with pattern-matching faculties may find
+a cleverer way for their compiler and type system to support this.)
+
+### different implementors of Node
+
+Though the methods on the `Node` interface are defined as those necessary for
+examining data of the [Data Model Kinds](../../data-model-layer/data-model.md#kinds),
+**`Node` is not only implemented by the Data Model**:
+
+- Yes, `Node` is implemented by types that just hold basic Data Model info;
+- `Node` is also implemented by [Advanced Data Layouts](../../schemas/advanced-layouts.md) --
+	- consider a HAMT that spans many separately-seralized chunks of data; it should still be usable as if it's a regular map.
+- `Node` is also implemented by [Schema-typed Nodes](../../schemas/) --
+	- Both if implemented by a single implementation that evalates rules at runtime (so, finite count of implementing types and known at core library compile time)...
+	- or if handled by codegen/macros (unknown count / open set of implementors of `Node`; not known at core library compile time; may be created in other packages that import the core, rather than core importing them!).
+
+Even further, some libraries may choose to make even more various
+implementations of `Node` for optimizing performance of specific tasks:
+for example, a `Node` which implements basic Data Model "map" semantics,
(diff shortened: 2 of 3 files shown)


design/libraries/nodes-and-kinds.md
```diff
@@ -0,0 +1,152 @@
...
+
+Kind
+----
+
+`Kind` maps very directly onto the definition of
+[Data Model Kinds](../../data-model-layer/data-model.md#kinds).
+
+`Kind` does not include the Schema layer's concept of "struct", etc.
+
+`Kind` must be an enum, **and not a sum type**.  Attempting to implement
```
vmx, line 63: Based on previous conversation I assume you mean with "enum" something that can be extended and "sum type" something fixed. Then it doesn't make sense to me. I can see the `Node` should be open, but `Kind` is a fixed thing.

To me `Node` is an interface, `Kind` is a fixed list of items that describe the Data Model.

warpfork: No, other way around.  Enum is fixed.  Sumtypes aren't.

Enum has a cardinality of the count of its members.

Sumtype has a cardinality of the sum of the cardinality of whatever its members are... so it can easily become "countable infinity" if any one of the members is.

So,

> To me Node is an interface, Kind is a fixed list of items that describe the Data Model.

^ this statement you conclude with is indeed correct.

warpfork: (It **seriously** drives me nuts that the Rust syntax conflates these two things into one name.  They have such very, very, categorically different properties... :imp: :angry: :fire: )

mikeal:
Should I add a section on what a good `Block` interface should be?

design/libraries/nodes-and-kinds.md
vmx: I really need to brush my programming language/type theory vocab and write it down for 5 year olds. Or perhaps here it doesn't really matter. `Kind` is a fixed set of things (whatever we call it).

warpfork:
> Should I add a section on what a good Block interface should be?

I'd put it in another file.  I kinda have one theme that I wanna hammer the crap out of in this file, which is that Node really needs to be an interface or there will be regrets.

design/libraries/nodes-and-kinds.md
warpfork: ... Reading again in the light of the morning, I'm thinking "fixed" is probably not quite the word to disambiguate those two either.  Enum and sum types both have fixed membership in terms of types; it's just that the cardinality of member *values* can easily be infinite in a sum type (it happens as soon as a single thing like 'int' or 'string' is part of one of the sum's member types).

Whereas what the Node type should be is an interface, because that can have a non-fixed and not-known-in-advance number of member *types*.

warpfork: Maybe we actually **should** hammer out a quick type systems vocab primer/glossary for ourselves, here in this repo, just for clarity of reference and ease of linking?  Good idea @vmx !

Perhaps something in the 'concepts' directory.  The 'cardinality' doc over there already kind of gets close to flirting with this.  Expanding on it to talk about sum types versus enums versus product types versus countably-infinite scalars etc could be really useful.

warpfork:
There's a good idea for further followup in @vmx's [comment thread](https://github.com/ipld/specs/pull/241#discussion_r381555564) -- but I don't wanna do it today / in this PR :sweat_smile: 

Any objections on moving this one towards a merge?

design/libraries/nodes-and-kinds.md
vmx: Having some proper vocab/glossary would be great. I think I slowly grok what you mean with cardinality in regards to enums and sum types. Thinking of "enums" as what the name suggests: *enumerable* items, hence finite. A sum type can have recursion (that's probably not the right word)/wrap "invitite" (as in all ints) values.

warpfork: Yeah, you've... got me on a roll with this kick in the shorts, actually, so thanks for that :D  I'll probably hoist another PR with some content about this in a day or two.

rvagg: > I really need to brush my programming language/type theory vocab and write it down for 5 year olds.

Don't feel bad, I need to engage a compsci translation layer when trying to absorb some of @warpfork's thoughts. This just argues for needing to be more clear when things get overly academic so we can talk to a broad enough audience. It's probably on us when we see things like this to highlight them as needing further carification.

design/libraries/nodes-and-kinds.md
```diff
@@ -0,0 +1,152 @@
...
+----
+
+`Node` is a monomorphized interface for handling data -- in other words,
+we make all data look and act like a `Node`, so that we can write all of our
+functions against the `Node` interface, and have that work for any sort of data.
+
+`Node` has functions for examining any of the
+[Data Model Kinds](../../data-model-layer/data-model.md#kinds).
+For example, this means `Node` must be able to
+do a key lookup for a map kind,
```
rvagg, line 85: linebreaking pattern is getting pretty weird here, in most of the doc it kind of looks like you're breaking close to 80 chars on the nearest punctuation, but it breaks down in this para.

design/libraries/README.md
```diff
@@ -0,0 +1,10 @@
+Library Design Guidance
+=======================
+
+This directory contains some documentation of recommendations for
+library authors who want to make IPLD libraries in a new language
+(or, perhaps for readers who want to understand an existing library better).
+
+Some of the information expressed here comes down to opinions moreso than specification;
+what is good ergonomics may vary wildly per language, so take these as
+recommendations rather than strictures.
```
rvagg, line 10: I think a preface here about the perspective this is written from would be appropriate. Some of the language below applies very specifically to a narrow set of languages (the para on `Node` as an interface for example). But that's OK if we make it clear up-front where this is coming from.

Something like:

> These design guidelines are primarily based on the experience of building, and rebuilding IPLD libraries in Go and reflecting on the limitations of implementations that have existed in both Go and JavaScript and the implications of those limitations on the potential and feature-set of IPLD. The language used in these guidelines are reflective of a Go programming perspective but apply broadly to most strongly typed languages. Loosely and untyped languages will need to interpret these guidelines appropriately while extracting the key concepts.

design/libraries/nodes-and-kinds.md
```diff
@@ -0,0 +1,152 @@
...
+Almost all features should be implemented to take `Node` arguments,
+and return `Node` values.
+
+Traversals and walks can be implemented in this way.
+
+Selectors can be implemented in this way.
+
+Transformations can be implemented in this way.
+
+Codecs themselves can be implemented this way.
```
rvagg: Can you expand on this one? I can't even figure out what it would mean for a codec to take a `Node` and return a `Node`.

rvagg:
Some small suggestions in my comments. I'm OK with merging this after addressing those.

The most valuable thing from the discussion you and I had about this @warpfork from the JavaScript perspective was the ideas around porcelain vs plumbing which we've totally messed up in JS which leads to a radically different design of ipld-prime vs the APIs we currently have available in JS.

This is a higher-level conern that what you've addressed here so maybe it doesn't belong in this file—or it could go into an intro section. It'd be something like "why we need a Node abstraction around IPLD data" and discuss metadata and state maintenance. That sometimes you just want a native instantiation of some blob of IPLD data (`AsObject()`) but in doing so, you lose valuable data that you may need for certain workflows and you miss out on efficiency and performance opportunities.

rvagg:
this also deserves an entry in the main README.md I reckon

design/libraries/nodes-and-kinds.md
warpfork: Added some psuedocode to explore this.  I'm worried it might be distracting line noise if not read tolernately, but I guess it's probably better than nothing.

design/libraries/README.md
warpfork: Added a large block of caveats to the top of the doc.

I almost wonder if we won't end up making a habit of repeating chunks of these at the top of most of our documents.  Benedictions on one page or another don't really seem to carry over to even their most proximate siblings when readers jump into one doc or other without following any path to get there that we anticipated.

design/libraries/nodes-and-kinds.md
warpfork: Yeah, I blithely follow my heart.  I try to aim for "eighty-_ish_", but heavy on the _-ish_.  I'll also readily break on commas or sentences.

My overall rule is "try to make the diff minimally irritating in future rounds", so the punctuation heuristic helps with that, whereas a character count enforced strictly is irritation-maximalizing.

I don't think we'll get the equivalent of `go fmt` for natural language any time soon, though, so I try not to lose too much sleep over anything.

warpfork:
I addressed a bunch of comments, and this has been out for quite a while, so I'm gonna roll forward in accordance with that last ~"lgtm after suggestions" comment and merge this.

I *didn't* attempt to address that porcelain-vs-plumbing thing @rvagg [brought up](https://github.com/ipld/specs/pull/241#issuecomment-595566573), but I think it'd be really really excellent to do that in the future.

rvagg:
some of my recent work even switched to using the "porcelain" terminology .. it wasn't really part of my CS language (aside from `git`'s use of it) but you've colonised my head on this one

PR #242, opened 2020-02-23
Introduce a type theory glossary.

warpfork:
The goal here is to provide just enough discussion about type theory and the basics of applied state counting that we can establish some language-independent terminology clearly.  This should in turn be useful for clarity in conversation, and as a quick link to drop in many other documents that require some foundational concepts, or need to avoid language-specific jargon.

A lot of this exists in literature and theory already... but gathering it in one place, written in one style, in a single page that can be read top to bottom in one sitting... seems to provide value.(Sending someone off on a quest to "read wikipedia and all the related content around concept $X" is great and all, but, ehm.  It's a little high latency, a little unreliable in outcome, etc.)

This ended up subsuming and expanding upon the already existing cardinality doc, so that file is deleted; its original content and much more is now in the new file.

This was kicked off in large part by discussion over in https://github.com/ipld/specs/pull/241#discussion_r381942147 , so thanks to @vmx for the good idea and the kick in the shorts to start writing.

Also thanks to @bsundsrud and @batmanaod for some polishing and out-of-band feedback on early drafts, and to @Reasonable-Solutions for some review of the categorical bits, all of which was very helpful.


concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
+A quick Primer and Glossary of Type Theory
+==========================================
+
+This document provides a quick overview of "type theory", and specifically,
+how we use it to reason about data and design in IPLD.
+
+We take "type theory" to mean primarily: a way of reasoning about programs
+and data by counting how many possible states they may have.
+We call this counting of possible, inhabitable states counting "cardinality".
```
BatmanAoD, line 9:
```suggestion
We call the number of possible, inhabitable states "cardinality".
```

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+(We acknowledge that there's much more to "type theory" than this in the
+broader literature and history of computer science, but since IPLD is
+about *data*, focusing on this subset of type theory suits our purposes.)
+
+You'll find this document is linked to by the IPLD specs -- both for describing
+key theoretical foundations of the Data Model, and also to better describe some
+of the advanced features in the Schema layer (plus the reasons for them) --
+*and* you'll also find this document linked to by the
+[library design notes](../design/libraries) -- where we use some of this
```
BatmanAoD:
```suggestion
[library design notes](../design/libraries), where we use some of this
```

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+Now, consider a pair of boolean values: how many states can it have?
+Four: `(true, true)`, `(true, false)`, `(false, true)`, or `(false, false)`.
+So, the cardinality of two booleans is `4`.
+
+The exercise gets more interesting as we consider larger accumulations of values.
+
+Consider three boolean values: how many states can it have?
+(You can draw the full table yourself, for this one.)
+When you count them all up, you should reach `8`.
```
BatmanAoD: Readers will probably recognize the exponential growth here, so it's probably better to mention that instead of or in addition to suggesting drawing a full table of possible values.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+But now, let's imagine something new: suppose we've got some sort of values
+that we call "nooleans".  Nooleans can have the value 'nue' or 'fuu'.
```
BatmanAoD: It seems like this example doesn't actually go anywhere? I was expecting you to explicitly note the isomorphism with bools.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+We'll explore more examples of this, and start identifying categories of
+things where we can do predictable kinds of math with cardinality in
+the next section: [Categories of Types](#categories-of-types).
+(Spoiler: the examples above correspond to an enum, then a product type,
+then another bigger product type, and finally a sum type.)
```
BatmanAoD: Did you mean for noolean to be a sum type?

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+Categories of Types
+-------------------
+
+We can identify a few categories of "types" based on how we evaluate their cardinality.
+(These definitions will use names that are popularized in computer science,
+and so they *may* match the name used in your favorite programming language -- but
+whether or not the same is familiar, the *concept* is universal, even across languages!)
+
+### enums
```
BatmanAoD: I would suggest using "enumeration" here rather than the language-keyword abbreviation, which, as you've noted, is prone to causing an incorrect understanding of the underlying concept. (Well, I guess that's only true of Rust. C enums are bad for many reasons, but "naming accuracy" is not one of them.)

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+https://chadaustin.me/2015/07/sum-types/ also contains some useful content
+and examples from a variety of languages.
+
+### product types
+
+Product types have a cardinality that is *the multiplied cardinality of their member types*.
+
+This may sound outlandish, but you've been using them all your life:
+*"product types" are commonly known as "structs"*
+(or, "classes", in more aggressively object-oriented languages).
```
BatmanAoD, line 125: "Classes" are often more likely to be infinite-recursive than structs. Heck, in C#, "capable of being self-recursive" would be a reasonable way to distinguish `struct`s from `class`es!

So I'd maybe just state that classes _can_ be product types, rather than implying that the definition of "class" is equivalent to "product type".

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+known as "(discriminated) unions", "(tagged) variants", or other terminology.
+(We use the "union" term ourselves, in the lexicon of IPLD Schemas!)
+
+You can also consult the Wikipedia page on
+[Sum Types](https://en.wikipedia.org/wiki/Sum_types)
+for broader discussion of the topic.
+https://chadaustin.me/2015/07/sum-types/ also contains some useful content
+and examples from a variety of languages.
+
+### product types
```
BatmanAoD, line 119: In languages with "record" types, are they generally product types? (I haven't used such a language yet.)

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+This may sound outlandish, but you've been using them all your life:
+*"product types" are commonly known as "structs"*
+(or, "classes", in more aggressively object-oriented languages).
+
+Product types usually show up in prose descriptions of data as the word "and".
+
+For example, say: "I have a boolean and another boolean and another boolean" --
+you might write code for this as `struct{ a bool, b bool, c bool }`,
+and its cardinality would be "|2*2*2|" -- "|8|" --
```
BatmanAoD, line 131:
```suggestion
and its cardinality would be "|2*2*2| = |8|" --
```

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+two for each of the booleans, and multiplied because each of them may vary independently,
```
BatmanAoD, line 132:
```suggestion
two for each of the booleans, multiplied because each of them may vary independently,
```

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+and thus each of the booleans adds another "dimension" to the state space described by this combined value.
```
BatmanAoD, line 133: It may be worth noting that product types have a concept of order. People may think of "three booleans" as three indistinguishable coins being tossed, which of course has cardinality 4.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+we could say "I have an Animal, *and* it's on top of a Furniture" --
+and then I could have any one of of *six* states
+("cat on a chair", "cat on desk", "dog on a chair", etc):
+the cardinality is "|3*2|".
+
+### infinite recursives
+
+Infinite recursives, as it says in the name, have infinite cardinality.
+
+Maps and lists are both examples of infinite recursives.
```
BatmanAoD: By "recursive," I was assuming you meant a type that can contain itself, but this sort of depends on what's meant by "map" and "list" here. In general, the statement you're making is true of "containers", whether or not they're recursive. I suppose all fp-style containers can be represented as recursive types, so maybe it doesn't matter? (Is there another category of types that represents the general concept of a "container" as used in e.g. C++?)

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+Since we can keep inserting entries into a map, or appending values to a list,
+and still have a map or list respectively, we can't count their cardinality.
+
+Note that it's still useful to do math around these, though.
+For example, if you have "maybe a map" (see the [maybe types](#maybe-types) section,
+coming later, in the [Types in the Wild](#types-in-the-wild) chapter) you may
+find it useful to declare that the cardinality of your map is "a" (an arbitrary placeholder),
+and then the cardinality of "maybe map" is "|a+1|".
```
BatmanAoD, line 157: For the mathematically inclined, it may be helpful to mention here that this is why we refer to "cardinality" rather than "ordinality" (since `order(a+1) = order(a)` when `a` is infinite).

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+"Unit" types don't come up too often... or at least, they're not often named as such.
+Just like "product types" are quite common once you recognize them, though, so is "unit".
+
+"Null" can be thought of as a "unit" type.
+
+"Maybe"/"Optional" types (further discussion later in the [Types in the Wild](#types-in-the-wild) chapter,
+specifically the [maybe types](#maybe-types) section) can be thought of as "unit plus a thing".
+
+Enums can be considered a sum type of where every type is itself a just another named unit type!
+(Whether or not you find this a *useful* perspective is up to you.  If it's confusing, forget it.)
```
BatmanAoD, line 199: Heh, glad to see this made it into the doc.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+In programming languages, the definition of "features" usually is expressed in
+terms of a set of methods.  This makes sense for programming languages,
+because if you don't know what methods you can call on a thing,
+you can't really do much with the thing!
+
+One could also imagine using a different definition of "features" to describe
+"interfaces" that apply purely to data:
+For example, stating that some map has the features to conform to an interface
+only if some entries of some known names and value types are present in the map.
+(However, this is not often seen, so this document will continue to frequently
```
BatmanAoD, line 231: Maybe it's not often expressed in these terms, but this is exactly how "stringly typed interfaces" play out in practice. JSON APIs are an obvious example.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+### explicit versus structural interfaces
+
+Interfaces as found in programming languages tend to fall into one of two large camps:
+*explicit* interfaces, or *structural* interfaces.
+
+Explicit interfaces require types implementing them to say so.
+
+Structural interfaces state some set of methods (or more generally, features)
+that must be present on a type for it to satisfy the interface;
+then, any type that has those methods implements the interface.
```
BatmanAoD, line 243: I.e. "duck typing"?

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+Both explicit and structural styles of interfaces have "open" membership --
```
BatmanAoD, line 245:
```suggestion
Both explicit and structural styles of interfaces have "open" membership:
```

I'm not sure why "compile" and "link" are in scare quotes below.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+#### booleans
+
+In many languages, the concept of `boolean` can be understood as a particular instance of an enum --
+it's just an enum with two values!
+Sure, it may be implemented as 'primitive' in many languages... but clearly,
+for cardinality counting, we can see a boolean as exactly "|2|".
+
+### sum types in the wild
+
+In Java, *there is no direct support for sum types*.
```
BatmanAoD, line 289: People coming from the JVM world _may_ be familiar with Scala's "sealed traits", which are sum types declared as though they were interface types but with the word `sealed` ensuring external modules can't add new variants.

BatmanAoD: (i.e. a language-supported version of the "create an interface, and promise..." approach)

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+(See https://github.com/BurntSushi/go-sumtype for tooling to help support this;
+the readme also contains good documentation of the concept!)
+
+In Rust, the `enum` keyword (confusingly) also creates sum types.
+Rust's sum types are at least properly closed (unlike Java's and Go's).
+
+#### maybe types
+
+In *many* languages, you may find a type in the standard library called "Maybe",
+or equally frequently, "Optional", or, "Either".
```
BatmanAoD, line 305:
```suggestion
"Optional", or "Either".
```

BatmanAoD: An equivalent example is "nullable types" e.g. in C# and Kotlin.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+These are a sum type!
+They're typically explained as "either **none**, or **some** of a thing":
+so, they take the cardinality of a thing and simply "plus one" to it,
+and colloquially call that additional state the "none" state.
```
BatmanAoD, line 309:
```suggestion
and colloquially call that additional state the "none" or "empty" state.
```

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+For some `Maybe<T>`, the cardinality is "1+|T|".
+
+(This is also a great example for how we can use cardinality counting methods
+to create clear comparisons between differently named but functionally
+identical concepts between programming language communities!)
+
+### product types in the wild
+
+In Java, `class` creates a product type.
```
BatmanAoD, line 331: Or, as mentioned above, an infinitely recursive type.

C# is probably worth mentioning explicitly here: it's the only mainstream language (possibly the only language, period?) where `struct` refers _only_ to product types, and infinite recursive types must be `class`es.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+In Go, `struct` creates a product type.
+
+In Rust, `struct` creates a product type.
+
+### interfaces in the wild
+
+In Java, the `interface` keyword introduces an interface.
+These interfaces are [*explicit*](#explicit-versus-structural-interfaces) -- a class must use the `implements` keyword to explicitly state that it can satisfy that interface.
```
BatmanAoD:
```suggestion
These interfaces are [*explicit*](#explicit-versus-structural-interfaces): a class must use the `implements` keyword to explicitly state that it can satisfy that interface.
```

I love em-dashes, but I really feel like you're making colon jealous

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+(That the class has sufficient methods to match the interface is then also checked at compile time -- but having the methods is not necessary rather than sufficient for the "implements" relationship to be considered true).
```
BatmanAoD, line 341:
```suggestion
(That the class has sufficient methods to match the interface is then also checked at compile time -- but having the methods is necessary rather than sufficient for the "implements" relationship to be considered true).
```

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+Numbers are... interesting.
+
+For the most part in computing, we tend to pretend and build abstractions as if
+numbers are [infinite scalars](#infinite-scalars).
+
+Of course, this is rarely true in practice.
+Integers are often implemented in such a way that values that would require beyond 32 (or 64, etc) bits are unrepresentable;
+arguably, this makes them more like a (very (*very*) large) enum than anything else.
+Floating point numbers implemented in IEE764 style also have their own entire host of foibles.
```
BatmanAoD: Instead of referring to "foibles", I'd perhaps limit it to noting that the range of values is similarly constrained by the number of bits used, but that the specific values representable by a floating point number of a given size are different from the values representable by an integral type of the same size. (In fact, given signed zero and multiple types of NaN, they have the same number of valid values but represent a strictly smaller set of _numbers_, especially if you don't consider infinity to be a number.)

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+### what about numbers?
+
+Numbers are... interesting.
+
+For the most part in computing, we tend to pretend and build abstractions as if
+numbers are [infinite scalars](#infinite-scalars).
+
+Of course, this is rarely true in practice.
+Integers are often implemented in such a way that values that would require beyond 32 (or 64, etc) bits are unrepresentable;
+arguably, this makes them more like a (very (*very*) large) enum than anything else.
```
BatmanAoD, line 359: It may be worth mentioning that most languages will provide an infinite-recursive type in the standard library (or the language itself, e.g. Python) for representing arbitrarily large integers.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+
+Types in IPLD Schemas
+---------------------
+
+It probably won't surprise to you hear that IPLD Schema's core features
+map very closely to the categories of types we've discussed:
+
+- [enums](#enums) are `enum`.
+- [sum types](#sum-types) are `union`.
```
BatmanAoD, line 396: This being the case, it's probably worth mentioning explicitly in the "sum types" section that C/C++ unions _are_ sum types but are...not really usable as such.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+Instead, the *whole Schema* is treated as an interface:
+since you can take some data and take a schema, and attempt to unify them to get typed data if the schema "matches"...
+you can use a series of Schemas in the same way as you'd use a series of interfaces.
+(IPLD Schemas are effectively [*structural*](explicit-versus-structural-interfaces) interfaces.)
+
+
+Different sizes of Infinity
+---------------------------
+
+// todo: not quite sure what to say about this.  the answer isn't "give up", that's for sure.
```
BatmanAoD: Is there much that you want to say here beyond what you've already said above, e.g. with "maybe map"?

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+// but i'm also not sure how to give this a brief treatment in a beginner-friendly way.
+
+// maybe just some admonishments about the dangers of subtracting from infinity -- i.e. "reserved keywords" -- would be useful.
+
+// "infinity plus one" is also still very useful to track opaquely as "a+1" rather than flatten to inf.
+
+// something about shifting/unshifting prefixes/suffixes could be useful -- link to "infinite hotel" metaphor.  folks seem to often find this a bit heady, though, so a more applied story would be nice.
```
BatmanAoD: The infinite hotel, if I recall correctly, is about Cantor-style uncountables, not about enumerable cardinality. I would strongly advise against bringing this up in any way other than "if you've read the infinite hotel story, please don't try to apply that here".

BatmanAoD (commented):
Some thoughts/suggestions from an IPLD outsider

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+members is **not** considered an "enum".
+(We'd call that an [infinite scalar](#infinite-scalar) instead.))
+
+### sum types
+
+Sum types have a cardinality that is *the sum of cardinality of their member types*.
+
+Sum types usually show up in prose descriptions of data as the word "or".
+
+For example, say: "I have a boolean *or* an int" --
```
vmx: If I read "int" I think of "oh is that this architecture sized thing from C?". I would write "integer" instead. This really isn't that important, it's more of a personal note you might agree with or not.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+and still have string, we can't count its cardinality.
+
+Infinite scalars aren't very different from infinite recursives, arguably.
+("A string is just a list of characters" is one such argument.)
+Still, we often regard these things as having separate categories in practice:
+strings are usually treated as a single token in serializers, despite being "infinite";
+numbers are infinite (even if there are some practical problems with this:
+see [what about numbers?](#what-about-numbers) in the [Types in the Wild](#types-in-the-wild) chapter),
+but certainly aren't usually considered to be recursive (though this does sometimes
+see usage in theory -- see (Peano Arithmetic)[https://en.wikipedia.org/wiki/Peano_arithmetic]!);
```
vmx:
```suggestion
see usage in theory -- see [Peano Arithmetic](https://en.wikipedia.org/wiki/Peano_arithmetic)!);
```

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+In Go, *there is no direct support for sum types*.
+The nearest thing possible is to create an interface with an unexported method,
+and make sure only a known number of types in the same package implement it.
+(See https://github.com/BurntSushi/go-sumtype for tooling to help support this;
+the readme also contains good documentation of the concept!)
+
+In Rust, the `enum` keyword (confusingly) also creates sum types.
+Rust's sum types are at least properly closed (unlike Java's and Go's).
+
+#### maybe types
```
vmx, line 302: Why is the "Java, Go, Rust" section missing here? In Rust it is `Option<T>`.

concepts/type-theory-glossary.md
vmx:
```suggestion
Floating point numbers implemented in IEE754 style also have their own entire host of foibles.
```

vmx: +1. (non-native speaker note: I've never came across the word "foible" before).

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+For the most part in computing, we tend to pretend and build abstractions as if
+numbers are [infinite scalars](#infinite-scalars).
+
+Of course, this is rarely true in practice.
+Integers are often implemented in such a way that values that would require beyond 32 (or 64, etc) bits are unrepresentable;
+arguably, this makes them more like a (very (*very*) large) enum than anything else.
+Floating point numbers implemented in IEE764 style also have their own entire host of foibles.
+
+The scope of this document is not to solve this conundrum, nor to try to convince you how you should consider numbers.
```
vmx, line 381: As much as I enjoy  reading elaborate English and learning new words, I think for this docs we should keep it simple :) I've heard "conundrum" for the first time, rephrasing it would be appreciate.

vmx (commented):
Wow, that was such a good read. This document really does what it should. Now those terms sound so simple and straightforward and I feel enabled to talk about them.

You link a lot (which is great!), using the Markdown syntax where you put the links at the bottom and you are linking with just `[this is a link]` might help keeping things in sync/you sane.

concepts/type-theory-glossary.md
rvagg: ^ that. This needs to be expanded. Is it about "interchangeable systems" or something else? When I started reading this I was expecting this to be about booleans-with-null, introducing a cardinality of 3. Either way, this needs to have its dots connected.

concepts/type-theory-glossary.md
rvagg: Maybe just add a clarification, something like:

>  Maps and lists are both examples of infinite recursives in that they can contain a theoretically unbounded number of additional maps and lists.

This terminology ("recursive") has come up in the schema documentation and it's worked quite well there I think.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+Types in the Wild
+-----------------
+
+These different categories of types are present in most programming languages,
+but not all of them are present in every language (!),
+and they often go by different names in different programming communities.
+Here's a few (non-exhaustive) examples in a handful of languages:
+
+### enums in the wild
```
rvagg, line 269: How exhaustive do you want to be with languages here?

JavaScript doesn't strictly have enums but can be [emulated in a variety of ways](https://2ality.com/2020/01/enum-pattern.html), TypeScript has them.

concepts/type-theory-glossary.md
rvagg: Sum types can be emulated in most languages that have some concept of "private" too tho, with public factory methods, right? You don't really even need the private/public side of this if you have strong language conventions.

concepts/type-theory-glossary.md
rvagg: Do languages that support proper `null` (JavaScript, Java, not Go—`nil` is weird) have this in a basic form, "may be concrete or may not exist"?

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+so, they take the cardinality of a thing and simply "plus one" to it,
+and colloquially call that additional state the "none" state.
+
+For some `Maybe<T>`, the cardinality is "1+|T|".
+
+(This is also a great example for how we can use cardinality counting methods
+to create clear comparisons between differently named but functionally
+identical concepts between programming language communities!)
+
+### product types in the wild
```
rvagg, line 329: might be reasonable to say that JavaScript doesn't have any formal definition of a product type? sadly our corollary of classes, prototypes, define methods but not data, so I don't think they help out here.

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+with the caveat that libraries processing data which exceeds the boundaries they can process should error and halt explicitly.
+(In IPLD, we have the good fortune that this definition is unusually easy to stick to:
+since we only carry data around, and don't provide mathematical operations on it,
+it's easy to expect that data which was processable on the way in will remain processable throughout the time we handle it.)
+
+
+Types in IPLD Schemas
+---------------------
+
+It probably won't surprise to you hear that IPLD Schema's core features
```
rvagg: getting a bit chatty here, maybe wind back to a bit more formal language

concepts/type-theory-glossary.md
```diff
@@ -0,0 +1,396 @@
...
+
+IPLD Schemas don't have a concept of [interfaces](#open-vs-closed-membership).
+At least, *not exactly* -- not *within* the schema.
+Instead, the *whole Schema* is treated as an interface:
+since you can take some data and take a schema, and attempt to unify them to get typed data if the schema "matches"...
+you can use a series of Schemas in the same way as you'd use a series of interfaces.
+(IPLD Schemas are effectively [*structural*](explicit-versus-structural-interfaces) interfaces.)
+
+
+Different sizes of Infinity
```
rvagg: delete section if you can't find words for it, I see it adding only marginal benefit which makes it not worth holding up this PR for

rvagg:
Love it, there are some minor things to address in here but let's get it merged.

There's also a section at the bottom of /schemas/schema-kinds.md titled "Understanding Cardinality" that should include a link to here. The pattern in those docs I've taken looks like this:

`See [IPLD Schema Kinds](./schema-kinds.md) for more information on this topic.`

A link here would be just above the examples. The URL will need to be absolute, to this repo (complete with `https://...), until we get this page published to specs.ipld.io.

concepts/type-theory-glossary.md
warpfork: Good idea.  Adding a section for this.

I'm going to include the claim that "these "big" numbers are typically *not interchangable* with the language's built-in number types".

I know that's not *always* true -- there's always an oddball language out there (in this case, iirc, Ruby actually allows bignum shockingly interchangeably with primitives) -- but I think it's *almost* always true, to the point I'm willing to skip any further qualification.  Do you know any other languages that make them interchangable?  Does python?  Are there qualifications required on *how* interchangable?  (If there's a lot of languages that do this, maybe it deserves a note; but if most of those are with a bunch of "well, except $X" hedging, then it tips back to not deserving expansion.)

vmx: > Do you know any other languages that make them interchangable?

IIRC Erlang has arbitrary sized numbers by default. I have no idea how small numbers are represented internally.

warpfork:
This is a response to a couple of the individual comments on this diff,
but I'm gonna post this as a top-level comment for visibility and reference and future discoverability --

I want the "in the wild" sections to be short, pithy triads of examples.
I'm not sure I'll generally be favorable towards adding more.
The whole document is meant for quick, single-sitting(!) reading,
and if the examples start including detailed caveats, variations, opinions, and dozens of languages, then that goal becomes failed.

I picked three languages that give an interesting cross-section.
(E.g.: one has enums; one has sum types; one has neither.)
They might not be a perfect choice, but I think they're an interesting and illustrative choice.

I didn't pick javascript because there's less of a type system there and more a collection of functions and patterns that sometimes(!) have common naming.
I didn't pick lisp for the same reason.
It doesn't mean those languages aren't important; I just found it harder to make illustrative comments using them.

I didn't pick haskell because it's actually too good at these concepts;
anyone who knows haskell or other similar families of language probably doesn't need this primer.
It doesn't mean those languages aren't important; I just found it less useful to make illustrative comments using them.

So, while I appreciate all the comments and reviews and the future PR's I'm sure will come on this topic:
there's a good chance my response will be along the lines of "that's true; perhaps it would be interesting if you made a blog about it, or a comparison table in a larger more detailed document somewhere else :)".
(And it's a true wish: if you make a writeup like that, accumulating links to those in a "More Information" table in the bottom of this primer will be the best of all readability situations!)
Or, if you want to add something to this document: make a PR that adds that language to *each* example.  (If it's coming in as a suggestion for one of the example sets, I don't necessary have enough familiarity to fill in the examples in all the other sets.  And it's possible that in doing so, we'll see that the language doesn't provide interesting new perspective on any but one example, which may also change how we want to write about it.)

concepts/type-theory-glossary.md
warpfork: Yeah, I'm not sure what to say on this.  It's probably reasonable to say js doesn't have any formal or enforced concept that maps to product types.  But in practice, a lot of JS code probably still reasons around the idea that some "type" of objects will only have fields "x", "y", and "z"...

Is that useful to remark on?  I dunno.  I think "ways to map type theory to javascript" is probably its own whole small book, and it might be hard to wrangle in any lesser way.

(Instanceof https://github.com/ipld/specs/pull/242#issuecomment-600560619)

concepts/type-theory-glossary.md
warpfork: Not even remotely exhaustive.  :)

(Instanceof https://github.com/ipld/specs/pull/242#issuecomment-600560619 .)

JS and TypeScript might be nice to comment on because they're pretty large-audience.  (And TypeScript, IFAICT, actually maps *really* closely to a bunch of these concepts... which makes it doubly-fascinating, since javascript so very much doesn't, and yet TypeScript is positioned as being so close to javascript!  mind-:boom:!)  But they're a bit tricky to talk about, so if we want to try, it probably deserves its own PR, at minimum.

(That link to how to emulate them is a great piece of content to include if we do expand this, though.)

concepts/type-theory-glossary.md
warpfork: It's an H4, not an H3.  It's just an example of sum type.

I'm afraid expanding this section to list examples would be terribly dry: "Lang1 calls it Option.  Lang2 calls it Optional.  Lang3 calls it Maybe.  Lang4 calls it Option.  Lang5 calls it Either.", etc.

Yeah, I don't wanna talk about Go's typed nils.

Having a null is somewhat different and somewhat the same.  Null generally means a "+1", which makes it the same.  But when we talk about null, we generally talk about languages that allow null *anywhere*, which turns into a different beast, because it's like *every* Type *always* having a Maybe attached to it, rather than the Maybe being something you attach when appropriate.  Having null be an option *everywhere*, with no way to outlaw it, ends up being contrary to building cardinality-correct representations of data.

warpfork: ... hm.  Maybe "nullable is sorta a sum type" deserves another H4 around here.

Better phrasing than "sorta" wanted, though.  So I'm probably going to leave this for "later".

concepts/type-theory-glossary.md
warpfork: Fair point.  I'll apply `s/int/integer/g` to this whole document.

warpfork: lol, note to future self, never *actually* do `s/int/integer/g`.   integererfaces will result, along with pointegers, and a bunch of other very integereresting misspellings which are definitely not integererchangeable with real words.

vmx: `s/int\b/integer/g`?

concepts/type-theory-glossary.md
warpfork: Nice.

May this go to show in the future that I *definitely* should not be trusted with numbers.  Off-by-ten in the spec name... :sob: The irony!

concepts/type-theory-glossary.md
warpfork: Done.

concepts/type-theory-glossary.md
warpfork: I think it still might apply to some situations?

Consider: I have two types, both of which are strings (so, they have cardinality `|inf|` and `|inf|`).  I want to make a new type which is a sum type of them.  If I still want to represent this as a string, I can actually do this losslessly... by prefixing them with a discriminator.  But it's essential to do that prefix consistently such that it's not subsumed in the infinity the two initial types can already contain or the result is insufficiently defined.


... you're right, it's still overkill, isn't it.  What I wanted to get at here is an admonition against trying to make sum types without discriminators, and especially doing it wrong in the form of sloppy manipulations of infinite scalars.  But that's a different thing and the infinite hotel certainly isn't the greatest tool for explaining this.

Anyway: I dropped this whole section, per other comments on it.  If it can't be finished today, it can be rebooted in some later PR.

concepts/type-theory-glossary.md
warpfork: Instanceof https://github.com/ipld/specs/pull/242#issuecomment-600560619 .

It's true you can produce any of these patterns with enough methods, either compiler-enforced or convention-and-pray, but it becomes increasingly confusing to talk about more ways to do it in this document that just wants to convince people that the concept exists and can be extracted.  I think the Go and Java examples already cover most of the it-isn't-explicitly-supported-so-i-pretended options at a broad level.

concepts/type-theory-glossary.md
warpfork: I tend to think of them as different, but I outright resorted to plugging "structural typing vs duck typing" into a search engine, because this does seem murky and arguable to me.  One of the first things that came up was https://softwareengineering.stackexchange.com/questions/259943/is-there-a-difference-between-duck-typing-and-structural-typing , and I think I more or less agree with the answer there.

"Duck typing" tends to refer to probing a thing (and typically, at _run_ time, rather than compile time) to see if it can do what you want.  I also tend to conceptualize it as probing for *specifically* the function you're about to call -- this is often what it means in practice when discussed in the Ruby community (and if they didn't originate this terminology, they're certainly at least where I'm familiar with it from).

"Structural typing" involves having a whole suite of constraints in a bundle, and checking them all at once.  It also seems to be the phrasing that's used more often when people are discussing compile time checks.

But I'm using a lot of words like "tends" and "seems" and "often".  So I agree they're pretty dang similar.

I'm also just plain biased here because I've seen "duck typing" used to justify a lot of bad code over the years, and for whatever reason, I've seen very little of the same for "structural typing".  Does that seem nonsensical, given that they're describing almost the same thing, when one dissects it?  Yes.  Do I make this observation nonetheless?  Also yes.

So if I have to choose one term to signal-boost: it's gonna be "structural typing".  And the distinction is murky enough I'm not actually sure I want to draw attention to it in this primer.

concepts/type-theory-glossary.md
warpfork: Hm, maybe I should rewrite this paragraph somehow.  I'm just not sure quite how.

The concept of "interfaces but for data" is what I'm trying to get to in the very end, because that's the direction that IPLD Schemas explore into.

But I'm not actually sure if JSON APIs in practice are a great example for this.  Or rather, they are, but with a zillion fascinating caveats that make me not want to give it its own entire space rather than try to treat it in this document.  (Lots of JSON APIs in practice end up using some language's type facilities to define *sort of*/*most of* the rules, but end up suffering from this in at least three different ways: the language semantics intrude (e.g. nullability appearing even if you don't want a +1 cardinality); the language types don't contain *all* the rules; and those facts combining make a nightmare for documentation, clarity, compatibility, etc.)

And "stringly typed" invokes the idea of giving up on compile-time or useful automation around type systems entirely, and writing a bunch of ad-hoc stuff.  Whereas where I'd like to go is towards useful type systems and automation and away from ad-hoc stuff, even while we stay close to the realities of serial data.

concepts/type-theory-glossary.md
BatmanAoD: Python does, yes (though you can observe when the switch from primitive to bignum occurs using `type()`).

concepts/type-theory-glossary.md
warpfork: Made another run at this riffing from @rvagg's text.

concepts/type-theory-glossary.md
warpfork: I guess my question is... does it work *without caveats*?

I'm not even entirely sure how to qualify this question in a language agnostic way.  But for example, in java, I can think of about a dozen ways where if one tries to mix `Integer` and `BigInteger`, the result is Not What You'd Want and the pitfalls are numerous.

BatmanAoD: @warpfork This is probably not helpful for the document, but yes, as far as I can tell, there are no caveats. I don't even know if there's a way to prevent integer arithmetic on the built-in numeric types from producing oversized ints.

concepts/type-theory-glossary.md
warpfork: You're right that the phrasing is stilted, and perhaps this could still be improved... but I think I'm going to leave this one for now, because I really want to hammer the association with the concept and action of _counting_ rather than the existence of _a_ number.

concepts/type-theory-glossary.md
warpfork: Too soon.  This is still the intro.  I'm going to fix this and the half-finished example in the next comment by A) tossing that and B) clarifying this is an intro and exploration continues in subsequent sections.

concepts/type-theory-glossary.md
warpfork: I don't honestly remember what was in my head when I started writing this bit.  I'm going to drop it entirely now in the interest of getting this document finished enough to merge gracefully.

concepts/type-theory-glossary.md
warpfork: I can only conclude neither of you are any fun :P "foibles" is being replaced over my *cough* ~dead~ subsequently-PR'd-but-otherwise-receptive body.  It's simply beyond my capabilities to talk about floating point numbers without getting a little whimsical.

The point about different representability than an integer type with the same bit count, and signed zero, are worthy of remark though.  I'll add a bit about that.

warpfork: ... although now that I've written it, and look at it again, I wonder if providing copious detail here actually makes the section _worse_.  The point of the section was to say that the type theory in this document is _not_ acknowledging the floating point situation.

Ah well.  It's written.  I'll push it.  We can edit further later.  Moving on.

concepts/type-theory-glossary.md
warpfork: I'm going to leave any editing that should result from this discussion as future work.  I don't feel confident enough or have the time to finish researching this deeply enough to make this section better right now.

concepts/type-theory-glossary.md
warpfork: Yeah, okay.  I've been trying to refrain from getting "too many" languages carving out an explicit mention, but this one is present enough in worldwide discourse that it's worth a mention.

warpfork:
Okay... whew...

Rebased and a bunch more things addressed.

Thank you so much for all the comments.

This PR has been daunting me for a while from the sheer volume of them, but I still appreciate it immensely :)  I took most of the changes.  And at this point I think I also have to leave a few behind, and declare bankruptcy on processing much deeper on this in one thrust.  The amount of time that goes into trying to form good, general cross-language statements that can survive examination, much less at arbitrary resolution, is... significant.

Since the general tone feedback seems positive, I'm going to go ahead and merge this now.  (I want to start being able to link to it!  And github's started doing the "$N hidden conversations" thing 😒, so this PR is definitely long enough that it's getting difficult to manage.)  If we've got more refinements to make -- and I'm sure there's plenty of them possible! -- let's do subsequent PRs.

warpfork:
> The URL will need to be absolute, to this repo (complete with `https://...), until we get this page published to specs.ipld.io

Punting on this, btw.  I tried, but I couldn't bring myself to do it -- linking to a path that includes `github` and `tree/master` just zests my giblets a hair over the line.  These links will be broken until we finish renovating the website to fix the whole situation; we should motivate ourselves to do that rather than work around it.

concepts/type-theory-glossary.md
BatmanAoD: Wow, I'm so glad I'm looking through all the resolved comments to find your new responses; I completely missed "integererfaces" two months ago.

PR #247, opened 2020-03-10
dag-pb: express as IPLD Schema

### LINKED ISSUE #245: Describe DagPB also as IPLD Schema
DagPB doesn't support the full IPLD Data Model, nonetheless it can be serialized into the IPLD Data Model. Hence it would be great to have an IPLD Schema for it. Such a Schema already exists in code at https://github.com/ipld/go-ipld-prime-proto/blob/e32bd156a1e52653dbdb8b1696d74205a20f220c/gen/main.go and could be translated into the schema syntax.

rvagg:
Fixes: #245

While I have it open, may as well push something for this. Feel free to edit directly if you have opinions.

### DIFF
--- a/block-layer/codecs/dag-pb.md
+++ b/block-layer/codecs/dag-pb.md
@@ -6,7 +6,7 @@ DAG-PB does not support the full ["IPLD Data Model."](../../data-model-layer/dat
 
 ## Format
 
-The DAG-PB IPLD format is a format implemented with a single protobuf.
+The DAG-PB IPLD format is a format implemented with a single protobuf:
 
 ```protobuf
 // An IPFS MerkleDAG Link
@@ -35,6 +35,21 @@ message PBNode {
 
 The objects link names are specified in the 'Name' field of the PBLink object.
 All link names in an object must either be omitted or unique within the object.
+
+This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
+
+```ipldsch
+type PBLink struct {
+  Hash Bytes
+  Name optional String
+  Tsize optional Int
+}
+
+type PBNode struct {
+  Links optional [PBLink]
+  Data optional Bytes
+}
+```
 
 ## Pathing
 
--- a/block-layer/codecs/dag-pb.md
+++ b/block-layer/codecs/dag-pb.md
@@ -4,9 +4,9 @@
 
 DAG-PB does not support the full ["IPLD Data Model."](../../data-model-layer/data-model.md)
 
-## Format
+## Serial Format
 
-The DAG-PB IPLD format is a format implemented with a single protobuf:
+The DAG-PB IPLD serial format is described with a single protobuf:
 
 ```protobuf
 // An IPFS MerkleDAG Link
@@ -36,6 +36,10 @@ message PBNode {
 The objects link names are specified in the 'Name' field of the PBLink object.
 All link names in an object must either be omitted or unique within the object.
 
+## Logical Format
+
+When we handle DAG-PB content at the Data Model level, we treat these objects as maps.
+
 This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
 
 ```ipldsch
@@ -51,9 +55,21 @@ type PBNode struct {
 }
 ```
 
-## Pathing
+The first node in a block of DAG-PB data will match the `PBNode` type.
 
-The pathing is currently different between implementations. Please see [issue #55] for more information about the harmonization effort. This section describes the current implementations as of September 2019.
+When decoding data with the DAG-PB codec, maps with exactly these fields will result.
+
+When creating data, you can create maps using the standard Data Model concepts,
+and as long as they have exactly these fields, the DAG-PB codec can encode them.
+If additional fields are present, the DAG-PB codec will error, because there is no way to encode them.
+
+## Alternative Pathing
+
+While the [logical format](#logical-format) implicitly describes a set of mechanisms for pathing over and through DAG-PB data,
+DAG-PB also enjoys some other special forms of pathing in addition to the Data Model norms, and these are supported by most major applications that use DAG-PB.
+
+This alternative pathing is covered here as part of this descriptive spec, but was developed independently of the Data Model and is thus not well standardized.
+It currently differs between implementations. Please see [issue #55] for more information about the harmonization effort. This section describes the current implementations as of September 2019.
 
 The Go and JavaScript implementation both support pathing with link names: `/<name1>/<name2>/…`.
 


block-layer/codecs/dag-pb.md
```diff
@@ -36,6 +36,21 @@ message PBNode {
 The objects link names are specified in the 'Name' field of the PBLink object.
 All link names in an object must either be omitted or unique within the object.
 
+This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
+
+```ipldsch
+type PBLink struct {
+  hash optional Bytes
```
warpfork: Should this be `optional` when we describe it with an IPLD Schema?  I think not.

Name apparently *is* optional, the prose above suggests.  Tsize... I think also tends towards being considered technically optional, though I sort of don't want to touch that discussion again.  But the hash itself?  Pretty sure if that's not present, it will quickly result in nope.

I suspect `hash` is only optional in the protobuf because of the weird "everything should be optional" meme in protobuf lands, and I don't think we should carry that into IPLD Schema descriptions because we don't have any of the same problems nor ergonomics around optionality that protobuf does.

block-layer/codecs/dag-pb.md
```diff
@@ -36,6 +36,21 @@ message PBNode {
...
+  name optional String
+  tsize optional Int
+}
+
+type PBNode struct {
+  links [PBLink]
+  data Bytes
```
warpfork: `data` *should* be `optional`?  I think PBNode with only lists of links and no data are not uncommon.

block-layer/codecs/dag-pb.md
```diff
@@ -36,6 +36,21 @@ message PBNode {
...
+
+```ipldsch
+type PBLink struct {
+  hash optional Bytes
+  name optional String
+  tsize optional Int
+}
+
+type PBNode struct {
+  links [PBLink]
```
warpfork: I wonder if `links [PBLink]` should be marked `optional`.  Does protobuf on the wire have distinguishment between an empty and an absent `repeated` element?

rvagg:
reaching the limits of my protobof knowledge, @vmx are you any deeper on this stuff?

warpfork:
I pushed another commit to the branch with a little more prose.  It added section headers to separate description of the serial format and the logical format into their own topics.  While I had it open, I also poked at the introduction of the Pathing section a bit, to try clarify that it's one of several mechanisms, and nonstandard compared to other parts of IPLD.

@rvagg you may want to take a look at that and make sure I didn't make a complete mess of it.

block-layer/codecs/dag-pb.md
```diff
@@ -36,15 +36,48 @@ message PBNode {
...
+and as long as they have exactly these fields, the DagPB codec can encode them.
+If additional fields are present, the DagPB codec will error, because there is no way to encode them.
+
+## Alternative Pathing
+
+While the [logical format](#logical-format) implicitly describes set of mechanisms for pathing over and through DagPB data,
+DagPB also enjoys some other special forms of pathing in addition to the Data Model norms, and these are supported by most major applications that use DagPB.
+
+This alternative pathing is covered here as part of this descriptive spec, but was developed independently of the Data Model and is thus not well standardized.
+It currently differs between implementations. Please see [issue #55] for more information about the harmonization effort. This section describes the current implementations as of September 2019.
```
rvagg, line 72: TIL that `[issue #X]` is a thing in GFMD. But, maybe put this long-form so it's safe for publishing to specs.ipld.io when we get there. `[issue #55](https://github.com/ipld/specs/issues/55)`.

warpfork: It's one of those bottom-link things.

block-layer/codecs/dag-pb.md
```diff
@@ -36,15 +36,48 @@ message PBNode {
 The objects link names are specified in the 'Name' field of the PBLink object.
 All link names in an object must either be omitted or unique within the object.
 
-## Pathing
+## Logical Format
 
-The pathing is currently different between implementations. Please see [issue #55] for more information about the harmonization effort. This section describes the current implementations as of September 2019.
+When we handle DagPB content at the Data Model level, we treat these objects as maps.
+The map keys correspond to the lowercase field names.
```
rvagg: Is this lower-casing true in Go? It's not in [JS](https://github.com/ipld/js-ipld-dag-pb), we have `node.Data` and `node.Links`, then `link.Name`, `link.Tsize` and `link.Hash`.

I wouldn't even say we treat it as a map if I were being precise about the way it's handled in JS, but maybe the rough idea is good enough to state that way.

block-layer/codecs/dag-pb.md
rvagg: oh, right, I'll hold my tongue from further comment lest it get me in trouble

warpfork: (I know -- I did a double-take when editing it too.  I'm really dubious of this feature of markdown.  If it had a different syntax -- `[[` or something -- it wouldn't be so bad; as it is, it looks to me like one just forgot to finish typing.)

block-layer/codecs/dag-pb.md
warpfork: Uuf, you're right: https://github.com/ipld/go-ipld-prime-proto/blob/e32bd156a1e52653dbdb8b1696d74205a20f220c/pb_node_gen.go#L599-L619 it's currently implemented in uppercase in the new stuff as well.

I'm not sure if I wanna fix our doc here or if I want to change the implementation.  This capitalization is distinctly weird.

(It's *definitely* and absolute [a map](https://github.com/ipld/go-ipld-prime-proto/blob/e32bd156a1e52653dbdb8b1696d74205a20f220c/pb_node_gen.go#L1356-L1358) in go, though.)

vmx: I think also in the Data Model it should be the capitalized names, so that it corresponds to the Protobuf names (1:1 mapping). I think a rename is fine here, but we can also use the capitalized names directly (I kind of like the rename thing, to show the power of IPLD Schemas).

block-layer/codecs/dag-pb.md
ribasushi: > Does protobuf on the wire have distinguishment between an empty and an absent repeated element?

No. The "sub-frame" is either there or it isn't. There is no way to say on-wire "there are zero repeated elements of type N"

block-layer/codecs/dag-pb.md
ribasushi: > I think PBNode with only lists of links and no data are not uncommon.

They aren't valid. You need to signify the "sub-codec" for the links:
- is it a bunch of file fragments? ( type 2, formerly type 0 )
- is it a directory with all links being named ( type 1 )
- is it a single named symlink ( type 4 )
- is it a hamted-directory-shard ( type 5 )

Also due to any intermediate file-stream node ( type 2 ) in a graph being a file on its own right, the size of the stream section and the sync-hints go into `data`

TLDR: not optoinal

ribasushi: ( above is written in the sense of "there is virtually no use of dag-pb outside of unixfs", if there is such a beast what I wrote above does not apply )

block-layer/codecs/dag-pb.md
mikeal: we should just make the schema read exactly like the Data Model because when you sit down to write selectors for this stuff it’s going to be against the Data Model anyway. it’s going to be super painful if we have to get everyone to adopt a schema rename in their selector engine in order to handle the queries being written against the renamed keys.

block-layer/codecs/dag-pb.md
Stebalien: Exactly:

* Required = exactly one (`.`).
* Optional = at most one (`.+`).
* Repeated = any number (`.*`).

block-layer/codecs/dag-pb.md
Stebalien: For DagPB, data _is_ optional. However, for unixfs, it's _not_ optional.

And yeah, there are uses outside of unixfs (the webui has a dagpb geolocation database, cluster uses it, the pin datastructure uses it, etc.). They're just not commonly transferred/queried on the network (well, except for the geolocation databse).

block-layer/codecs/dag-pb.md
Stebalien: Therefore, there is no meaningful distinction between "empty" and "not there", so I wouldn't mark it as optional.

block-layer/codecs/dag-pb.md
Stebalien: In terms of the _schema_, I wouldn't distinguish between "there is data but its empty" and "there is no data".

block-layer/codecs/dag-pb.md
Stebalien: From a schema standpoint, I wouldn't treat it as optional. Size/name are definitely optional (both technically and logically from a schema standpoint).

block-layer/codecs/dag-pb.md
rvagg: > I wouldn't distinguish between "there is data but its empty" and "there is no data".

I'm not sure where that leaves us, does an `optional` descriptor help us here at all? I suppose to describe the format, adding an `optional` is telling the reader that there _are_ use-cases where this might not be present. So, I should add an `optional`?

block-layer/codecs/dag-pb.md
rvagg: I've unwound the two mentions of lowercase, and moved the schema to just use uppercase. ptal

rvagg:
I've pushed some edits here: took away `optional` on the Hash and made the names start with uppercase characters. I think I'm synthesizing the conversation here so far but please take a look and let me know if you're unhappy with the state, otherwise I'll assume silence as agreement.

block-layer/codecs/dag-pb.md
warpfork: Seems we're spec'ing DagPB here, not unixfs(v1), so: on this line we should say `data optional Bytes`, yep.

(And continue to **not** say `nullable` -- saying `nullable` would be claiming there's a distinguishment between "there is data but its empty(/null)" and "there is no data", which per @Stebalien 's statement would be incorrect.)

Stebalien: Ah. So you distinguish between "optional" (does not need to be specified in the serialized data) and "nullable" (null doesn't matter)? In that case, everything in dagpb is optional, and nothing is nullable.

rvagg: ok, so `links` is optional too then? or will that just be a zero-length array?

I'm getting a bit mixed up here with what we're trying to represent. We're saying this is the "logical" format but I think this pushes it into the language level where the distinction between `nullable` and `optional` starts to get weird and maybe not so helpful. When we're talking about representation then yes, crisp and clear, but that's not what this is saying. The representation is described by the protobuf spec above this block. I _think_ we should just be aiming for what is most conceptually useful to the reader of this doc trying to understand what data is in a dag-pb block.

Have pushed a new version but I'm dubious, latest is:

```ipldsch
type PBLink struct {
  Hash Bytes
  Name optional String
  Tsize optional Int
}

type PBNode struct {
  Links optional [PBLink]
  Data optional Bytes
}
```

vmx: > I _think_ we should just be aiming for what is most conceptually useful to the reader of this doc trying to understand what data is in a dag-pb block.

The reason I opened this issue is, that go-ipld-prime transforms from the Data Model into the actual DagPB. Rust IPLD does the same and I'd like to make sure the both are actually doing the same thing/all implementers can just look at the schema, implement it and are good.

rvagg: ok, well, 🤷‍♂, I think this might be good to go but it's a bit hard to tell for sure since concepts are not lining up with crystal clarity

vmx:
@rvagg sorry for screwing up this PR's log. I thought I merge master into it and things will be fine. Somehow there was still a conflict. So I ended pushing a proper rebase, hence the force push. I just updated it to use `DAG-PB` instead of `DagPB` to be consistent.

This PR has been open for long and I haven't seen any objections, so I think it's good to go.

PR #256, opened 2020-03-31
tcp congestion control algorithm - cubic vs Westwood

RubenKelevra:


rvagg:
I think it might be best to leave these docs as static archives of older content of the issues they're pulling from. I'm very appreciative of your input both here and in #76 @RubenKelevra and don't want to discourage further exploration on this but maybe the place for that is further continuation in #76 on a casual basis for now, or opening a new issue _if_ there is a concrete proposal.

The problem we're having with this topic area is that it's fairly low down on the priority stack as we try and build a coherent toolchain. We're much more focused on the basics right now for these things to figure very much in the range of concerns that IPLD has right now. That's not to say that it might not make sense to consider it at a different place within the IPFS stack, nor that it will never be a concern here—it will be, we're just getting our head above water on the basics for now!

So I think we can encourage more discussion on these topics and this may all be helpful for background when we end up having enough pieces in place for this to be helpful.

RubenKelevra:
@rvagg 

How about adding this content just into ipfs/go-ipfs/docs as server system optimization documentation?

There are some other tweaks for Linux which can speed up the operation of ipfs as well.

rvagg:
@RubenKelevra you might have to try adding it over there since it's a separate project. I can't speak for the IPFS team at all but I imagine if you craft something so that it fits in well then it should be a good addition.

RubenKelevra:
Thanks for your opinion, will work on an extended document over there :)

PR #259, opened 2020-04-28
DagJSON: Be more specific about Multibase Base64 encoding

vmx:
How CIDs are encoded was underspecified. The binary type is now also using
multibase instead of plain Base64.

Reasons for using Multibase instead of raw Base64:

 - It is clear which variant of Base64 is used
 - CID libraries support it natively
 - Having symmetry between different Base64 encded items makes it easier
   to remember and probably also the implementation easier to follow

### DIFF
--- a/block-layer/codecs/DAG-JSON.md
+++ b/block-layer/codecs/DAG-JSON.md
@@ -42,8 +42,8 @@ The Bytes kind is represented as an object with `"bytes"` as key and a [Multibas
 A link is represented as a base encoded CID. CIDv0 and CIDv1 are encoded differently.
 
  - CIDv1 is represented as a Multibase Base32 encoded string. The Base32 encoding is the one described in [RFC 4648, section 6](https://tools.ietf.org/html/rfc4648#section-6) without padding, hence the Multibase prefix is `b`.
- - CIDv0 is represented in its only possible Base58 encoding. The Base58 encoding is the one described in [Base58 draft](https://tools.ietf.org/html/draft-msporny-base58), hence the Multibase prefix is `Q`.
+ - CIDv0 is represented in its only possible Base58 encoding. The Base58 encoding is the one described in [Base58 draft](https://tools.ietf.org/html/draft-msporny-base58).
 
 ```javascript
-{"/": String /* Multibase Base58 encoded CIDv0 or Multibase Base32 encoded CIDv1 */}
+{"/": String /* Base58 encoded CIDv0 or Multibase Base32 encoded CIDv1 */}
 ```
--- a/block-layer/codecs/DAG-JSON.md
+++ b/block-layer/codecs/DAG-JSON.md
@@ -16,22 +16,34 @@ Codec implementors **MUST** do the following in order to ensure hashes consisten
 This produces the most compact and consistent representation which will ensure that two codecs
 producing the same data end up with matching block hashes.
 
-### Simple Types
+### Natively supported kinds
 
-All simple types except binary are supported natively by JSON.
+All kinds of the IPLD Data Model except Bytes and Link are supported natively by JSON.
 
 Contrary to popular belief, JSON as a format supports Big Integers. It's only
 JavaScript itself that has trouble with them. This means JS implementations
 of `dag-json` can't use the native JSON parser and serializer.
 
-#### Binary Type
+### Other kinds
+
+The non-natively supported kinds are wrapped in an object, where the key is a slash (`"/"`) and the value is the actual kind.
+
+#### Bytes kind
+
+The Bytes kind is represented as an object with `"bytes"` as key and a [Multibase](https://github.com/multiformats/multibase) Base64 encoded string as value. The Base64 encoding is the one described in [RFC 4648, section 4](https://tools.ietf.org/html/rfc4648#section-4) without padding, hence the Multibase prefix is `m`.
+
 
 ```javascript
-{"/": { "base64": String }}
+{"/": { "bytes": String /* Multibase Base64 encoded binary */ }}
 ```
 
-#### Link Type
+#### Link kind
+
+A link is represented as a base encoded CID. CIDv0 and CIDv1 are encoded differently.
+
+ - CIDv1 is represented as a Multibase Base32 encoded string. The Base32 encoding is the one described in [RFC 4648, section 6](https://tools.ietf.org/html/rfc4648#section-6) without padding, hence the Multibase prefix is `b`.
+ - CIDv0 is represented in its only possible Base58 encoding. The Base58 encoding is the one described in [Base58 draft](https://tools.ietf.org/html/draft-msporny-base58), hence the Multibase prefix is `Q`.
 
 ```javascript
-{"/": String /* base encoded CID */}
+{"/": String /* Multibase Base58 encoded CIDv0 or Multibase Base32 encoded CIDv1 */}
 ```


mikeal (approved):

mikeal:
I agree, will need to update the dag-json codec to match

rvagg:
so this would be a breaking change to the spec, we're comfortable with that? because this is not expected to be actually used much in the wild?

vmx:
> so this would be a breaking change to the spec, we're comfortable with that?

The breaking change is in the binary type and that one is AFAIK only implemented in JS, it is not in Go or Rust. So I'm comfortable with that change.

mikeal:
I don’t know of anyone currently using Binary in `dag-json` either. It is, for obvious reasons, the worst part of `dag-json` so whenever you encounter binary it’s better to use links to `raw` blocks instead.

block-layer/codecs/DAG-JSON.md
```diff
@@ -24,14 +24,18 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
...
 
+### Other Types
+
+The IPLD Data Model specifies two types that cannot directly be represented in JSON. They are both represented as [Multibase](https://github.com/multiformats/multibase) Base64 encoded strings. The Base64 encoding is the one described in [RFC 4648, section 4](https://tools.ietf.org/html/rfc4648#section-4) without padding, hence the Multibase prefix is `m`.
+
 #### Binary Type
 
 ```javascript
-{"/": { "base64": String }}
+{"/": { "base64": String /* Multibase Base64 encoded binary */ }}
```
austinabell: isn't the `"base64"` key redundant if you are going to store the base64 prefix in the string? 

If this key is unneeded maybe it should be more compact? Seems like adding the multibase prefix is breaking anyway, so why not update that when doing this?

block-layer/codecs/DAG-JSON.md
```diff
@@ -24,14 +24,18 @@ Contrary to popular belief, JSON as a format supports Big Integers. It's only
...
 ```
 
 #### Link Type
 
 ```javascript
-{"/": String /* base encoded CID */}
+{"/": String /* Multibase Base64 encoded CID */}
```
austinabell: Why specify specifically Base64 for this? If only allowing base64 encoding then why include the multibase prefix in the string encoding?

warpfork:
Can someone review whether this will make it significantly more difficult to process these entities -- specifically, if it will still be possible to do it with a fixed amount of lookahead in terms of token count during parsing?

There's a fairly hairy piece of code in the go codebase for handling this: https://github.com/ipld/go-ipld-prime/blob/7e0619f3a9842f4a84737fc622c8ce56ef0d5d46/codec/dagjson/unmarshal.go#L39-L73 -- and it's based upon a very fortunate coincidence that there's fixed number of tokens we might need to deal with before it becomes clear if we need to engage any special logical branches.  If this becomes untrue, it bodes poorly for any efficient parser implementations.

If that code (or something like it) can still safely work through that decision tree within that fixed amount of memory, I'm happy.  If it can't, we might want to search for alternative ideas that don't present such a problem.

---

Aside from the above concern, can confirm that bytes in dag-json simply haven't been previously supported by the go libraries, so no objections from me on that front.

block-layer/codecs/DAG-JSON.md
mikeal: We still need some kind of key, because it can’t be a bare string as we’re already using that for the `Link` type

block-layer/codecs/DAG-JSON.md
mikeal: Because almost all the code we have to generate a string from a CID returns a multibase encoded string. It would actually be more error prone for this to not be multibase encoded.

mikeal:
FYI, this means we can’t link to CIDv0 from `dag-json` because you have to use `base58btc`. I’m fine with that, but just want to make sure everyone is aware since this is what we’re doing.

block-layer/codecs/DAG-JSON.md
austinabell: yeah I agree, I want it to be multibase encoded, I'm just saying for the spec to say in this comment that it is Base64 specifically might be a bit misleading?

block-layer/codecs/DAG-JSON.md
austinabell: Yeah, my suggestion was just for maybe something more compact since there is no new information given with this key.

Edit: I guess this doesn't matter since it's just JSON, just seems weird to have the duplicate definition, feel free to ignore this

Stebalien: "bytes"? Or "multibase"?

vmx: I also think "base64" isn't such a good name. I like @Stebalien proposition of "bytes", it aligns well with the naming the [IPLD Data Model](https://github.com/ipld/specs/blob/2014318ecba05c3a506d2676d49385905f86d7c4/data-model-layer/data-model.md#kinds).

So it would read:

```javascript
{"/": { "bytes": String /* Multibase Base64 encoded binary */ }}
```

block-layer/codecs/DAG-JSON.md
vmx: @ausinabell The idea is that it must be Base64 encoded CID. So that implementations don't need to support all base encodings. You can also generate Base64 encoded strings in the browser.

vmx:
> FYI, this means we can’t link to CIDv0 from `dag-json` because you have to use `base58btc`. I’m fine with that, but just want to make sure everyone is aware since this is what we’re doing.

I see two thumbs down. Please let's do this and only support CIDv1. Let's keep DagJSON easy to produce and consume without the need to have full library support.

block-layer/codecs/DAG-JSON.md
Stebalien: I get that from a compactness standpoint but it would be nice if the encoding matched the default CID encoding (base32).

Also, what about CIDv0?

Stebalien: > Also, what about CIDv0?

Ah, I see. Below.

Stebalien:
WRT CIDv0, I've floated this idea around a few times: https://github.com/ipld/specs/issues/261

block-layer/codecs/DAG-JSON.md
austinabell: Okay gotcha, didn't know there was a limitation for generating any of the multibases in js. Also yeah the default Cid encoding of base32 lower would be nice

austinabell:
> > FYI, this means we can’t link to CIDv0 from `dag-json` because you have to use `base58btc`. I’m fine with that, but just want to make sure everyone is aware since this is what we’re doing.
> 
> I see two thumbs down. Please let's do this and only support CIDv1. Let's keep DagJSON easy to produce and consume without the need to have full library support.

Oh I guess I was under the assumption that this would just be optional support (ie. if a provider or consumer supported the ambiguous multibase, it would work fine, but wouldn't have to?) I don't see why you couldn't serialize/ deserialize a v0 if you wanted to support. Didn't really mean to downvote, just meant to indicate that you can still technically do that

vmx:
I force pushed a completely new version, based on our conversations during the [weekly call](https://github.com/ipld/team-mgmt/blob/8ce1d9ac063881e2e0a3d5445ddfcd7d6d9bfe79/meeting-notes/2020/2020-05-04--ipld-sync.md).

I restructured it a bit and also changed the language to match other specs, like the Data Model one (talking about "kinds" and not about "types". Please do another full review.

Stebalien (approved):

block-layer/codecs/DAG-JSON.md
```diff
@@ -16,22 +16,34 @@ Codec implementors **MUST** do the following in order to ensure hashes consisten
...
+#### Link kind
+
+A link is represented as a base encoded CID. CIDv0 and CIDv1 are encoded differently.
+
+ - CIDv1 is represented as a Multibase Base32 encoded string. The Base32 encoding is the one described in [RFC 4648, section 6](https://tools.ietf.org/html/rfc4648#section-6) without padding, hence the Multibase prefix is `b`.
+ - CIDv0 is represented in its only possible Base58 encoding. The Base58 encoding is the one described in [Base58 draft](https://tools.ietf.org/html/draft-msporny-base58), hence the Multibase prefix is `Q`.
 
 ```javascript
-{"/": String /* base encoded CID */}
+{"/": String /* Multibase Base58 encoded CIDv0 or Multibase Base32 encoded CIDv1 */}
```
Stebalien:
```suggestion
{"/": String /* Base58 encoded CIDv0 or Multibase Base32 encoded CIDv1 */}
```

CIDv0 isn't multibase.

vmx: Oh wow, I've never noticed that.

vmx:
We talked about this in the last [IPLD Dev Meeting](https://github.com/ipld/team-mgmt/blob/4e66a998f6523c096caf96306a26bcc232aedc97/meeting-notes/2020/2020-05-04--ipld-sync.md) and found and agreement within the team, hence I merge it now.

PR #262, opened 2020-05-11
DagJSON: lowercase spec filename and fix links

vmx:
Now files in that directory are named the same way.


rvagg (approved):
Fine. But I'd like to register that I don't like the `Dag` prefix that gets used, it's an acronym in the same way as JSON and CBOR are, it _should_ be `DAG`. But whatever.

vmx:
> But I'd like to register that I don't like the `Dag` prefix that gets used, it's an acronym in the same way as JSON and CBOR are, it _should_ be `DAG`

Thanks for bringing this up. I've created https://github.com/ipld/specs/pull/263 to fix that.

PR #263, opened 2020-05-13
fix: capitalize DAG correctly

vmx:
It's e.g. DAG-CBOR and not DagCBOR.

I only updated the specs and not the exploration reports as I would
just keep them as they are.

### DIFF
--- a/FOUNDATIONS.md
+++ b/FOUNDATIONS.md
@@ -213,7 +213,7 @@ inline data into large objects (lots of duplication and copying).
 
 ## Primitives
 
-The "recommended" IPLD format (currently DagCBOR) needs to support *at a minimum*:
+The "recommended" IPLD format (currently DAG-CBOR) needs to support *at a minimum*:
 
 * 32/64 bit integers without losing information.
 * 32/64 bit floats without losing information.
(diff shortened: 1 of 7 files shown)


rvagg (approved):
as stated elsewhere, I prefer this because of correctness, but we're up against historical-cultural baggage here that others may not be happy with steamrolling over, so let's tread lightly and not move too quickly on this

warpfork:
I've also alternatingly spelled it `dag-cbor` and (when required for e.g. package names, which have character legality fun) `dagcbor`.

I'm +1 that picking something and standardizing it in the specs docs is Good :tm:, but beyond that I really don't care one way or the other what it is.  Don't consider me blocking for anything, one way or another.

(You'll also probably not easily get me to habitualize to writing in SCREAM-CAPS though, fwiw.  Four capitals in a row seems to empirically be about the max one can get away with before it starts parsing as scream in my brain...  I'll do it in the specs repo if that's our chosen form, but I doubt I'd start doing it in informal writing.)

vmx:
It seemed common to uppercase "JSON" and "CBOR", hence I think it makes sense to also uppercase "DAG" for consistency. I also agree that screaming is not great (I also try to remove that from all filenames). I'm happy to informally say "dag-json". To me key here is, that it's consistent, whatever it is and I think uppercasing all is currently the most common ground.

PR #270, opened 2020-06-12
bitcoin: add bitcoin docs (WIP)

rvagg:
Not complete, but it's big enough and very tedious, that I just want to push _something_. If anyone feels like reviewing as a WIP feedback would be appreciated but I've got a lot more to do to connect the pieces to IPLD. Will call for reviews when I think it's ~finished.

### DIFF
--- a/block-layer/codecs/bitcoin.md
+++ b/block-layer/codecs/bitcoin.md
@@ -28,15 +28,15 @@ When representing Bitcoin block graphs as coherent IPLD DAGs, these trees need t
 
 Navigating using IPLD linking from a Bitcoin block graph header to transactions requires traversal via the "transaction Merkle root" through binary Merkle tree to the desired transaction. Navigating to individual transactions without these internal binary Merkle tree nodes is not possible using IPLD's linking system, which is why they must be present in explicit form when viewed with IPLD tooling.
 
-## Bitcoin Block Graph Format
+## Bitcoin Block Graph Binary Format
 
-The Bitcoin block graph format, available as hexadecimal via the Bitcoin Core CLI using the `getblock <address> 0` command, includes the following elements:
+The Bitcoin block graph binary format, available as hexadecimal via the Bitcoin Core CLI using the `getblock <address> 0` command, includes the following elements:
 
 1. Header - the first 80 bytes
 2. A list of one or more Transactions - for each Transaction:
   2.1 A list of one or more Transaction Ins - detailing historical transactions (and a specific Transaction Out) where coins are being transferred from, can be used to derive the wallet to debit
   2.2 A list of one or more Transaction Outs - detailing wallets where coins are being transferred (credited) to
-3. Transaction witness data - directly after some transactions, since BIP91 (SegWit)
+3. Transaction witness data - directly after some transactions, since [BIP141 (SegWit)](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki)
 
 ### Header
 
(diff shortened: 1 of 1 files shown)


block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+
+This usage of the term "block" does not map directly to the IPLD definition, so this document will add additional clarity when refering to these data structures by referring to them as a **block graph**. This term should be interpreted as consisting of a header, transactions and transaction witness data. The composition of this block graph will be explained below.
+
+### Content digests
+
+The Bitcoin format consistently uses a double-SHA2-256 hash to produce content digests. This algorithm is simply the SHA2-256 digest of a SHA2-256 digest of the raw bytes. These digests are also used publicly when referring to individual transactions and whole block graphs. The Bitcoin Core CLI as well as the many web-based block explorers allow data look-up by these addresses.
+
+When publishing these addresses, they are typically presented as big-endian in hexadecimal. To represent these in byte form on a little-endian system, they therefore need to be reversed and the hexadecimal decoded.
```
ribasushi, line 19: Since endianness is usually defined over a multibyte integer type, I am for real not sure which type of "little endian" is meant here ( and casual googling doesn't help ). If I see the following 128bit long payload on disk:
- `00 11 22 33 44 55 66 77 88 99 aa bb cc dd ee ff`
What is the actual value:
1. `33 22 11 00 77 66 55 44 bb aa 99 88 ff ee dd cc` 
2. `77 66 55 44 33 22 11 00 ff ee dd cc bb aa 99 88`
3. `ff ee dd cc bb aa 99 88 77 66 55 44 33 22 11 00`
4. Something else? 

Alternatively - if the on-disk structures are explicitly defined over >64bit integer types: this needs to be called out early, so folks like me get in the right mindset.

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Status: Descriptive - Final**
+
+This specification describes the method for representing the Bitcoin blockchain with IPLD. This specification _does not_ describe a flexible method for representing arbitrary data and the Bitcoin format _does not_ support the full ["IPLD Data Model."](../../data-model-layer/data-model.md). Rather, this specification is primarily for the purpose of representing Bitcoin data and interacting with it via IPLD tools and linking mechanisms.
+
+## Bitcoin as a Content Addressed Format
+
+### Terminology
+
+By definition, the Bitcoin blockchain in its entirety is a [Merkle DAG](https://en.wikipedia.org/wiki/Merkle_tree), with every "block" linking to the previous "block", until the genesis which does not contain a link. In addition, each Bitcoin "block" contains within itself a coherent Merkle DAG that links from an 80-byte "header" through to the transactions and transaction witness data. The Bitcoin ecosystem uses the term "block" to refer to a complete data structure that includes this header, transactions and transaction witness data. One of these "blocks" is produced roughly every 5 minutes and appended to the "blockchain".
```
ribasushi, line 11: It's 10 minutes not 5

```suggestion
By definition, the Bitcoin blockchain in its entirety is a [Merkle DAG](https://en.wikipedia.org/wiki/Merkle_tree), with every "block" linking to the previous "block", until the genesis which does not contain a link. In addition, each Bitcoin "block" contains within itself a coherent Merkle DAG that links from an 80-byte "header" through to the transactions and transaction witness data. The Bitcoin ecosystem uses the term "block" to refer to a complete data structure that includes this header, transactions and transaction witness data. One of these "blocks" is produced roughly every 10 minutes and appended to the "blockchain".
```

block-layer/codecs/bitcoin.md
ribasushi, line 19: Remove `address` to not confuse with a BTC address ( separate bag of worms )

```suggestion
The Bitcoin format consistently uses a double-SHA2-256 hash to produce content digests. This algorithm is simply the SHA2-256 digest of a SHA2-256 digest of the raw bytes. These digests are also used publicly when referring to individual transactions and whole block graphs. The Bitcoin Core CLI as well as the many web-based block explorers allow data look-up by such content digests.

When show to users such content digests are typically presented as big-endian in hexadecimal. To represent these in byte form on a little-endian system, they therefore need to be reversed and the hexadecimal decoded.
```

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+Navigating using IPLD linking from a Bitcoin block graph header to transactions requires traversal via the "transaction Merkle root" through binary Merkle tree to the desired transaction. Navigating to individual transactions without these internal binary Merkle tree nodes is not possible using IPLD's linking system, which is why they must be present in explicit form when viewed with IPLD tooling.
+
+## Bitcoin Block Graph Binary Format
+
+The Bitcoin block graph binary format, available as hexadecimal via the Bitcoin Core CLI using the `getblock <address> 0` command, includes the following elements:
+
+1. Header - the first 80 bytes
+2. A list of one or more Transactions - for each Transaction:
+  2.1 A list of one or more Transaction Ins - detailing historical transactions (and a specific Transaction Out) where coins are being transferred from, can be used to derive the wallet to debit
+  2.2 A list of one or more Transaction Outs - detailing wallets where coins are being transferred (credited) to
```
ribasushi, line 38: Markdown is dumb that way

```suggestion
  2.1. A list of one or more Transaction Ins - detailing historical transactions (and a specific Transaction Out) where coins are being transferred from, can be used to derive the wallet to debit
  2.2. A list of one or more Transaction Outs - detailing wallets where coins are being transferred (credited) to
```

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+Decorated with links, the basic structure of the Header is as follows:
+
+```ipldsch
+type Header struct {
+  version Int
+  previousblockhash Bytes
+  parent &Header # previousblockhash converted to CID with `bitcoin-block` codec
+  merkleroot Bytes
+  tx &TransactionMerkle # merkleroot converted to CID with `bitcoin-tx` codec
+  time Int
```
ribasushi, line 54: As you note below that's a `Uint32`: do we have a way to express that in `ipldsch` ?

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+```ipldsch
+type Block struct {
+  hash String                       # uint256 BE in hex form, the double-SHA2-256 of the block header
+  version Int
+  versionHex String                 # same as version, but base16
+  previousblockhash optional String # uint256 BE in hex form, will not be present for the genesis block
+  merkleroot String                 # uint256 BE in hex form
+  time Int
+  bits Int
+  difficulty Float                  # derived from `bits`
```
ribasushi, line 80: Ugh... didn't we kinda/sorta agree among ourselves that floats are there but **not** to be used, and discouraged at all costs? :)

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+* Block and transaction identifiers and hashes are always presented in this form as hexadecimal strings representing the big-endian form of the uint256 digests. This means the bytes are reversed from their standard little-endian form.
+
+### Transactions
+
+There are at least one transaction in a Bitcoin block graph. The first transaction is called the "coinbase" and represents the miner rewards. A block graph may _only_ contain a coinbase or it may also also contain a number of transactions representing the movement of coins between wallets. Each transaction contains a list of one or more "Transaction Ins" and a list of one or more "Transaction Outs" representing the flow of coins. The coinbase contains a single Transaction In containing the block reward and the Transaction Outs list represent the destination of the rewards. Non-coinbase transactions contain Transaction Ins representing the source of the coins being transacted, linking to previous transactions, and a list of Transaction Outs containing the details of the destination wallets.
```
ribasushi, line 100: > There are at least one transaction in a Bitcoin block graph.

Technically, past ~2140, when everyone working on this is dead, this may no longer be true ;P

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+
+Navigating using IPLD linking from a Bitcoin block graph header to transactions requires traversal via the "transaction Merkle root" through binary Merkle tree to the desired transaction. Navigating to individual transactions without these internal binary Merkle tree nodes is not possible using IPLD's linking system, which is why they must be present in explicit form when viewed with IPLD tooling.
+
+## Bitcoin Block Graph Binary Format
+
+The Bitcoin block graph binary format, available as hexadecimal via the Bitcoin Core CLI using the `getblock <address> 0` command, includes the following elements:
+
+1. Header - the first 80 bytes
+2. A list of one or more Transactions - for each Transaction:
+  2.1 A list of one or more Transaction Ins - detailing historical transactions (and a specific Transaction Out) where coins are being transferred from, can be used to derive the wallet to debit
```
ribasushi, line 37: I'd try to avoid using `wallet` throughout this document, as it implies... more than what we are dealing with. Suggested edit ( not sure if the best term )

```suggestion
  2.1 A list of one or more Transaction Ins - detailing historical transactions (and a specific Transaction Out) where coins are being transferred from, can be used to derive the source of funds to debit
```

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+}
+
+type TxIn struct {
+  prevOut OutPoint
+  scriptSig Bytes # unbounded vec<char>
+  sequence Int
+}
+
+type OutPoint struct {
+  hash Bytes # 256-bits
```
ribasushi, line 123: This, together with the #int64 below, almost makes one want to say "ipld schema integers are of arbitrary precision", and leave it up to the codecs when to switch the wire-representation, **and** leave it to codecs when to use a language internal `bigint` and when to use a native integer.

This has probably been discussed already, so feel free to ignore with no further discussion.

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+* Each `vec<char>` is prefixed by its size, using the compact size format. There must be exactly that number of bytes. There may be zero or more bytes in the byte array.
+
+Decoding a Transaction into this structure requires reading the following elements in order:
+
+* `version`: a signed 32-bit integer
+* `segwit`: is implicit and `false` for all block graphs prior to the SegWit soft fork, which occurred at a height of 481,824. After this height, the two bytes following `version` are inspected, if they are equal to `[0x0, 0x1]`, the bytes are consumed and `segwit` is `true`. If the bytes are not exactly these values, `segwit` is false, and the two bytes instead form the begining of `vin` (the first byte of `vin` is part of the compact size integer, and as `vin` must contain one or more elements, it cannot be `0x00`, hence the reliability of the `segwit` flag maintaining backward-compatibility).
+* `vin`: one or more elements, prefixed by a compact size int, then, for each element up to the size:
+  * `hash`: an unsigned 256-bit integer / a 32-byte binary string, the OutPoint transaction ID hash identifying the source transaction for the coins
```
ribasushi, line 145: This goes together with the endianness discussion above: being an integer and a string at the same time can't be a thing.

block-layer/codecs/bitcoin.md
```diff
@@ -0,0 +1,220 @@
...
+
+The secondary binary Merkle tree _does not contain the coinbase_ as the coinbase is used to hold the root of this binary Merkle tree. Instead, the first transaction entry (coinbase) is replaced with a "null" hash, i.e. 32-bytes of zeros (`0x00`).
+
+Continuing the IPLD Schema of the Transaction with `MaybePubKey`, we find that, for a transaction with segregated witness data, _one of the coinbase `vout` elements_ will contain a digest of a **WitnessCommitment** (which we turn into a Link with IPLD). This field is the digest of a nonce and the root of the secondary binary Merkle tree containing the witness data. There is no requirement that a particular `vout` element contain the witness commitment digest, rather, that _one of them_ must contain it. The 32-byte digest is prefixed by 6 special bytes: `{ 0x6a, 0x24, 0xaa, 0x21, 0xa9, 0xed }`, forming a 38-byte field. With this encoding, the `vout` list can be scanned for a 32-byte
```
ribasushi, line 195: Unfinished sentence

ribasushi (commented):
Did a first pass over this. Exciting!

block-layer/codecs/bitcoin.md
rvagg: 3, **as if** you read it entirely as a 32-byte unsigned integer, you read it in the reverse than you would if you treated it as LE. "usually defined over a multibyte integer type" is what's being got at here, but it's 32-bytes, not some repeating sub-pattern.

The "as if" makes me think this is leaning too heavily on the "uint256" thing too much. I'm tempted to remove that language entirely and say it's just a byte string and **by convention** it gets byte-reversed and turned into hexadecimal when presented publicly.

rvagg: In truth, I never touch this "uint256" thing myself in any of my code. It treat all of these things as byte arrays and then reverse+hexadecimal whenever I need to present the value. Otherwise they're only useful as byte arrays. So I guess that fact in itself suggests the backing out of this concept. It's really just window dressing to make the zeros go at the start of block addresses.

block-layer/codecs/bitcoin.md
rvagg: see discussion in #272, these are intended to be adjuncts to the schema when they're important, the schema isn't intended to be a feature-complete description format that can be used for encoders & decoders (maybe one day, but I'm pretty sure Eric will resist that until his death)

block-layer/codecs/bitcoin.md
rvagg: OK, I need to work on the language around this more, but the lead-in to this particular snippet of Schema says "the Header **may be represented**". There's no binary involved in `difficulty`, it's purely something that's _only_ present in the Data Model, just like `versionHex`. It's also not critical that an implementation of an IPLD Bitcoin codec include this since a consumer could derive it for themselves. I've done it so I can generate almost exactly matching JSON to compare against the Bitcoin Core RPC's output to validate my data.

There's more of this kind of awkwardness coming up, particularly when it comes to script decoding! When it comes to codecs like Bitcoin, what do we present at the Data Model layer? What does it mean to "decode" Bitcoin data? Do we _only_ present exactly what we find in the bytes? Does that mean we leave out the implicit Links (I'm inserting a `tx` CID to link to transactions and leaving `merkleroot` as a byte array, so it's essentially duplicated but not navigable in the raw form)? There's a bunch of different points along the "decode" spectrum that can be chosen and I think I'll have to write that up below and leave it to implementers to decide how far along they go.

I went the whole way with Bitcoin, I can even fully decode scripts into their string form and present them as the Bitcoin Core RPC does, so my decode/encode validation goes both ways and compares against the authoritative and only needs to discount the few chain-contextual pieces of data (`chainwork` etc.). Even as I'm working on Zcash (again), I think there's parts of it where I'm probably going to stop short and shortcut my validation because there's definitely diminishing value as you get into more detail.

block-layer/codecs/bitcoin.md
rvagg: Do you mean that it's only true if people are transacting on Bitcoin and beyond ~2140 there may no longer be transactions? It's still going to be true as long as someone is mining Bitcoin because there's always a coinbase. There cannot exist a "bitcoin block graph" without at least one transaction!

I'm looking through Zcash right now and it's kind of sad how many coinbase-only transactions there are near the head. It makes it look like it now exists to be mined ...

block-layer/codecs/bitcoin.md
rvagg: good point, I've removed all references to "wallet" because tbh I don't fully understand that side of things anyway!

block-layer/codecs/bitcoin.md
rvagg: ooops, I think this was an unsaved buffer in my editor, ftr it's:

> With this encoding, the `vout` list can be scanned for a `scriptPubKey` exactly 38-bytes long and prefixed with the 6 special bytes. The remaining 32-bytes are taken as the witness commitment digest.

rvagg:
Notes to self arising from discussion so far, for when I do revisions:

 * Probably remove the "uint256" concept, it's a confusing mess. In practice it's just a byte array that is presented to users as a byte-reversed hex string by convention
 * Notes about the various levels on the gradient of what it means to "decode" a bitcoin block and what you present at the data model - from strictly the elements you pull out of the binary format up to decoding all of the things including turning linkable things into CIDs and even decoding the script into its string format (or perhaps something more advanced?).
 * Work on something to clarify how ipldsch is being used—this applies broadly to our docs too, need better language to say "this defines a structure that could be conceived of as a data model thing, but we need more details as an adjunct to talk about specifics of binary representations", which gets tricky still because in these docs I'm even presenting different forms of data model things (see previous point)—the raw decoded pieces vs the more advanced version that can be presented to match the Bitcoin Core RPC (i.e. convention).

mikeal:
what’s the status here?

i’d like to get something up on the specs website that i can link to

rvagg:
status is that each time I sit down to attack this I'm overwhelmed by the size of the task to pull it together into a coherent form that covers everything that it needs to; but it does weigh on me that it's outstanding and I need to get it closed out along with js-multiformats reworks of the codec(s).

It's not in a worthy state to even merge as a draft tbh, so you're out of luck for now but I'll try and get to it asap.

block-layer/codecs/bitcoin.md
deefactorial: It's every 10 minutes not 5

warpfork:
It would be really cool to merge this, even if we want to put some disclaimer texts in somewhere.  This is way more and better information than we have on this topic anywhere else, as far as I know.

rvagg:
Not merge-worthy IMO, it's so far from what it should have been. I think a better approach might be to start from the reverse end, like the Filecoin, and now Ethereum data specs, and work backward. It turned out to be really hard to work forward like I was doing it here.
tbh I'm not sure what to do with this, I don't see any time on the horizon for me to finish this out but it's one of those things that linger in the back of my head, along with the code that backs this work which is also out of date now with the ecosystem it sits within.

PR #273, opened 2020-06-18
fix: raw blocks are bytes, ty @ribasushi

mikeal:


ribasushi (approved):
👍 this really tripped me up earlier :)

rvagg (approved):

PR #277, opened 2020-06-30
proposal: specify hash algorithm by integer code in hashmap

mikeal:
This might be worth doing since we’ve been trying to move away from string identifiers for multiformats. It’s also a little more compact when serialized.

### DIFF
--- a/data-structures/hashmap.md
+++ b/data-structures/hashmap.md
@@ -96,7 +96,7 @@ See [IPLD Schemas](../../schemas) for a definition of this format.
 ```ipldsch
 # Root node layout
 type HashMapRoot struct {
-  hashAlg String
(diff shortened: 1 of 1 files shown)


rvagg:
Line 138 should probably be updated to say "code" rather than "identifier", other than that I guess this is OK.

mvdan:
We discussed this on Slack a couple of weeks ago and agreed it's a good idea. Should we just apply Rod's suggestion and merge? I think GitHub nowadays allows maintainers to edit PRs directly.

rvagg:
> Line 138 should probably be updated to say "code" rather than "identifier", other than that I guess this is OK.

that still stands @mikeal, the text needs to be changed to match

PR #282, opened 2020-07-22
hashmap: document Filecoin HAMT variations

rvagg:
I believe this is correct as of now, but there's some potential modifications that may happen prior to mainnet.

This section will get linked from https://github.com/ipfs/go-hamt-ipld/pull/52 and the Filecoin specs which currently just point to the top of this doc.

### DIFF
--- a/data-structures/hashmap.md
+++ b/data-structures/hashmap.md
@@ -28,6 +28,9 @@
   * [Hash algorithm](#Hash-algorithm)
   * [Buckets](#Buckets)
   * [Security](#Security)
+* [Appendix: Filecoin HAMT Variant](#Appendix-Filecoin-hamt-variant)
+  * [Implicit and fixed parameters](#Implicit-and-fixed-parameters)
+  * [Block layout](#Block-layout)
 
 ## Introduction
 
@@ -44,7 +47,7 @@ The IPLD HashMap is constructed as a [hash array mapped trie (HAMT)](https://en.
 * Peergos [CHAMP](https://github.com/Peergos/Peergos/blob/master/src/peergos/shared/hamt/Champ.java) implementation
 * [IAMap](https://github.com/rvagg/iamap) JavaScript implementation of the algorithm
 * [ipld-hashmap](https://github.com/rvagg/js-ipld-hashmap) JavaScript IPLD frontend to IAMap with a mutable API
-* [go-hamt-ipld](https://github.com/ipfs/go-hamt-ipld) Go implementation, not strictly aligned to this spec
+* [go-hamt-ipld](https://github.com/ipfs/go-hamt-ipld) Filecoin Go HAMT implementation used by the [Lotus](https://lotu.sh/) client. See the appendix for how this implementation differs from this specification.
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


data-structures/hashmap.md
```diff
@@ -295,3 +298,55 @@ The impact on the data structure layout imposed by the use of buckets in the IPL
...
+
+The reference Go implementation for the Filecoin HAMT is used by the [Lotus](https://lotu.sh/) client and is available at <https://github.com/ipfs/go-hamt-ipld>.
+
+### Implicit and fixed parameters
+
+The Filecoin HAMT _does not_ use an explicit root block (`HashMapRoot`) to encode its parameters within the data. Instead it is expected that consumers of the data understand the parameters from a combination of the Filecoin specification and versioning of the blockchain over time. All HAMT nodes take the same form, there is no differentiation for a root node and an implementation must bring implicit parameters when decoding each node.
+
+* `hashAlg`: The hash algorithm used by the Filecoin HAMT is the x64 form of the 64-bit [MurmurHash3](https://github.com/aappleby/smhasher). _(TODO: this may be changed to SHA2-256 prior to Filecoin mainnet: https://github.com/filecoin-project/specs-actors/issues/517)_
```
vmx: The Rust implementation seems to use the 128-bit variant of the x64 Murmur3 hash: https://github.com/ChainSafe/forest/blob/b9ae7e8f53aeecaf2f3a7546dcdf40bb8b8ff55b/ipld/hamt/src/node.rs#L17

vmx (commented):
Thanks for documenting the Filecoin variation. I haven't checked if the schema matches the Rust implementation as it would be more time consuming than I anticipated.

data-structures/hashmap.md
rvagg: uh oh https://github.com/ipfs/go-hamt-ipld/blob/21886a1edb600df1cc6cead007fcba408f47af91/hash.go#L53

I'm pretty sure the 64 is _not_ a subset of the 128, so if they're both implemented correctly that HAMT shouldn't be able to read the nodes written by the Go HAMT. But, murmur3 is notorious for cross-impl compatibility problems!

vmx: I'll talk with the Forest folks about that.

rvagg:
re serialization in Rust:

* Node: https://github.com/ChainSafe/forest/blob/b9ae7e8f53aeecaf2f3a7546dcdf40bb8b8ff55b/ipld/hamt/src/node.rs#L56 seems to be right, if `(x,y).serialize(..)` is going to be a tuple
* Pointer: https://github.com/ChainSafe/forest/blob/b9ae7e8f53aeecaf2f3a7546dcdf40bb8b8ff55b/ipld/hamt/src/pointer.rs#L28-L45 the complex `{"0":CID}` & `{"1":[...]}` for serialization and deserialization seems to be all good.
* KV: there's no custom serialize or deserialize but I gather that the definition makes it a tuple and it gets all that free: https://github.com/ChainSafe/forest/blob/b9ae7e8f53aeecaf2f3a7546dcdf40bb8b8ff55b/ipld/hamt/src/lib.rs#L37-L38 `K` has to be a serializable thing so I hope that means that it'll always end up as bytes. The Go HAMT strictly assumes string keys, not serializable keys, and it does a conversion of string to bytes (hopefully avoiding unicode questions for the most part). Lotus does this thing where it'll take something that's bytes and convert it to a string for a `Set()` and then internally the HAMT converts back to bytes. I'm a little concerned about the cross-language compatibility of that process but this is where test fixtures are going to come in handy.

👍

data-structures/hashmap.md
austinabell: I'm pretty confident it is, we have compatibility tested against go impl for this. There is no 64 bit hash value btw as far as I can see. Maybe you are mixing up the 64 bit target with the hash bit length?

(I will look a bit more into to confirm though)

Edit: yup the go impl call is calling the 64 bit target hash which provides a 128 bit digest and that 64 bit digest is just half of the 128 bit. Here is the call in murmur3 go (`New64()` is called to setup the hasher):
```
// New64 returns a 64-bit hasher
func New64() hash.Hash64 { return New64WithSeed(0) }

// New64WithSeed returns a 64-bit hasher set with explicit seed value
func New64WithSeed(seed uint32) hash.Hash64 {
	d := (*digest64)(New128WithSeed(seed).(*digest128))
	return d
}
```

rvagg: heh, yeah I just had my head in there too. There's this comment just above what you pasted:

```
// digest64 is half a digest128.
type digest64 digest128
```

and the reference impl only has a 128 for x64 anyway so I guess it's a subset! https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp :thumbsup:

austinabell: This was something I didn't understand though when I was working on it, what would be the reason for not keeping all 128 bytes? Why intentionally limit the max depth? (I did overlook that this could be a potential inconsistency because the go implementation will hit max depth sooner and there doesn't seem to be a negative with keeping around more hashed bytes as they are never persisted)

data-structures/hashmap.md
```diff
@@ -295,3 +298,55 @@ The impact on the data structure layout imposed by the use of buckets in the IPL
...
+* `bitWidth`: The Filecoin HAMT fixes the bit width to `5`, meaning that each node of the HAMT can contain up to `2`<sup>`5`</sup> (`32`) elements containing either buckets or links to child nodes.
```
austinabell, line 313: actually, this caps out at 3, statically. it's the `arrayWidth` variable in the go hamt impl

data-structures/hashmap.md
```diff
@@ -295,3 +298,55 @@ The impact on the data structure layout imposed by the use of buckets in the IPL
...
+* `bucketSize`: The Filecoin HAMT fixes the maximum length of its buckets to `3`, meaning a maximally full HAMT leaf node can contain `32 x 3` (`96`) key/value pairs.
```
austinabell, line 314: yeah each node is capped at 3 statically, might want to reword these. the bit width is potential values to avoid collisions at each level

data-structures/hashmap.md
rvagg: No good reason I think, you have to have a pretty large amount of data to hit max depth though (unless you can force collisions as is being discussed in that SHA2-256 proposal thread). For 64 bits you get a maximum of 12 levels if you're taking 5 at each level (as Filecoin is), with 32 elements at each level, that's 32^12, then 3 at the leaves, which is `55340232221128790000` for a maximally full tree assuming a completely random distribution of hashes .. I think. So we're talking very large numbers with good randomness so maybe it was just a case of :shrug:, it's big enough for now (famous last words).

data-structures/hashmap.md
rvagg: Sorry, I'm not sure what you're suggesting here, can you suggest some text changes that might make it clearer in your view? This text is also building on a lot of content that already exists in this document that explains the algorithm and how these things all work together so it's assuming you're using all of that as a reference and these details are just the Filecoin HAMT specific bits.

rvagg: also, in https://github.com/ipfs/go-hamt-ipld/pull/52 I've proposed renaming `arrayWidth` to align with the terminology here to save confusion, I figure that it's easier to do it in the Go code than in this spec and all the JavaScript code that implements this spec

rvagg:
They've switched to SHA2-256 for `hashAlg` upstream so the whole murmur3 hassle has gone away. Have updated the PR to match.

vmx (approved):

rvagg:
932af7c

PR #283, opened 2020-07-27
dag-cbor: javascript number limitations

rvagg:
Exposing the nakedness of JavaScript's `Number` type and its implications for cross-language data communication.

Is this too much information for an otherwise compact doc? It seems valuable information to know for anyone wanting to cross language boundaries, but maybe I'm overthinking it.

### DIFF
--- a/block-layer/codecs/dag-cbor.md
+++ b/block-layer/codecs/dag-cbor.md
@@ -10,6 +10,8 @@
 * [Implementations](#implementations)
   * [JavaScript](#javascript)
   * [Go](#go)
+* [Limitations](#limitations)
+  * [JavaScript](#javascript-1)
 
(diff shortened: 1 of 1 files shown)


vmx (approved):
I think this is the right place to put this information. That's where I'd expect to find it.

You are not overthinking it, this is really useful. This way others using the JS stuff that know about this number mess within JS don't have to dig deeply themselves, but get the infos right here.

mikeal:
Love it 😍

warpfork:
Yay!

I also wrote some stuff in over in the other docs repo poking about this subject: https://github.com/ipld/docs/blob/master/details/numeric-size.md

Your prose here is a lot better and more precise though.  I think mine over there will need updates to match.

rvagg:
0cbb9c4

PR #286, opened 2020-08-12
Sneak in basic filecoin determinism "spec"

ribasushi:
This is not an official spec, as the retrieval side is still somewhat in flux.
But it reflects the state of the art as of August 2020.

/cc @mikeal ( I don't have perms on that repo )

### DIFF
--- a/block-layer/content-addressable-archives.md
+++ b/block-layer/content-addressable-archives.md
@@ -116,7 +116,7 @@ The remainder of a Section, after length-prefix and CID, comprises the raw byte 
 
 ### Determinism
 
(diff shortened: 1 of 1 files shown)


mikeal:
+1

Can we also document how inline CID’s are currently handled?

rvagg (approved):
going to 🙈 at the squishiness of the terms "first-seen" and "depth-first" and believe that most readers will interpret them in the appropriate way

ribasushi:
> Can we also document how inline CID’s are currently handled?

I'll get this a little later today, once I see how latest go-car behaves in go-ipfs.

PR #287, opened 2020-08-13
feat: more compact representation for nested bytes

mikeal:
This allows for a more compact representation of nested inline bytes in a list. Since you don’t need the length of an inlined binary value this means we can drop the list and int encoding when we inline bytes.

### DIFF
--- a/data-structures/flexible-byte-layout.md
+++ b/data-structures/flexible-byte-layout.md
@@ -15,7 +15,12 @@ type FlexibleByteLayout union {
 
 type NestedByteList [ NestedByte ]
 
-type NestedByte struct {
+type NestedByte union {
+  | Bytes bytes
+  | NestedFBL list
+} representation kinded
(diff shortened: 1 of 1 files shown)


ribasushi (approved):
Seems reasonable, yes...
Still pending that other "rollout of all permutations to ensure we are not missing anything" analysis that I want to do when I have the braincycles.

rvagg (approved):
more regressing back to originally discussed forms, this is fine by me but it's getting more complex to mentally picture all of the permutations so I support @ribasushi's plan to do a detailed analysis of these forms (which would also be an excellent exercise in digesting more of IPLD Schema-thinking I think!).

mikeal:
it’s getting more complicated in the schema but the actual encoded representation is still pretty simple. i’ll want to adopt this change in my implementation in order to understand any complexity cost though.

PR #293, opened 2020-09-08
doc: graphic for readme, adjust content to match

### LINKED ISSUE #244: Better separate codecs and blocks in documentation
I motion to split "block layer" and "codec layer" and do so consistently across all our specs and docs.

Fundamental reason: the work the codec code does is almost entirely about how to encode/linearize/serialize/whatever-you-call-it Data Model trees into flat bytes (and of course vice versa, unfurling flat bytes to DM).  It's very little to do with blocks.  There's two separate things going on here, and we should reserve space to talk about each of them.

Another proximate reason: the current block-focused naming makes people think that a "blockstore" interface is the centerpiece of this layer... And it's really, really not; we mislead people super badly in this regard.  We should refine our terminology so that we don't impart this mis-impression.

Another tertiary reason: putting codecs under the heading of blocks is likely to mislead people into thinking that our codecs can't do streaming operation, which just isn't true, and makes us sound more limited than we are.

And as a quaternary concern: we should also redouble our efforts to make sure there's no possibility of confusing "blocks" in IPLD with "chunks" and "chunking" of files in IPFS / unixfs-applications.  While we use the word "chunking" for the thing that's done to break up large byte ranges, imo neither we nor (just as importantly) other literature in the space at large use either term with 100% consistency, so the possibility of confusion is very real.  At minimum, identifying "blocks" cleanly and separately from "codecs" should help with this.  (At maximum, I wonder sometimes about the word "block" at all; but I don't come prepared with an alternative suggestion, so I'll park that thought.)

---

Here's some short phrases that would describe this relationship:

> Block->Codec->DataModel(<->Schema|ADL)

> You have a codec that transforms the datamodel into a block and back.
> The block is the unit the content addressed identifier is based on.

What do ya'll think?

rvagg:
Closes: https://github.com/ipld/specs/issues/244

View here: https://github.com/ipld/specs/blob/rvagg/readme-graphic/README.md

I've tried to be as light-touch as possible while trying to match the graphic. I don't want to bite off too much in this iteration but it's required a little bit of unwinding of the layer-cake approach. More adjustments can be made later as needed if we can pin this down.


vmx (approved):
Thanks for keep this going/getting this done!

mikeal (approved):

PR #297, opened 2020-09-15
dag-pb: more accurate schema and up-to-date impl description

rvagg:
Round 1,490 of DAG-PB specification cleanup..

I'm working on getting this just right in https://github.com/ipld/js-dag-pb/ and am taking the opportunity to be very strict about Data Model representation. We have the challenge of wanting to ensure that we have previously accepted some flexibility in the way the components of a DAG-PB were constructed (friendly API) but now we want to be much more strict - you _must_ give us an object of exactly the right shape as defined by the Schema such that if we round-trip it it'll come out the same way (while also offering utilities on the side to regain some of the friendly API functionality). 

Thanks to Hannah's efforts in https://github.com/ipld/go-ipld-prime-proto we have a good chance at full hamonisation of the two implementations! go-ipld-prime-proto has a much more strict interpretation of the Data Model form than the previous Go implementation (which offers the shortcut named pathing form). So I want to try and get that impl and the new JS impl aligned and the Schema here reflecting it correctly. I don't think we quite got it right the last time around, which was hampered by having to conform to the varying legacy implementations and hand-waving about Data Model form. I think we can be much clearer now that we have a clearer concept of the Data Model in the new implementations.

I'm looking mainly at https://github.com/ipld/go-ipld-prime-proto/blob/master/coding.go and its use of https://github.com/ipfs/go-merkledag/tree/master/pb to pull in properties to inform some of the Schema changes, I'd appreciate double-checking on that please. I plan to conform JS to this.

Some additional notes inline.

### DIFF
--- a/block-layer/codecs/dag-pb.md
+++ b/block-layer/codecs/dag-pb.md
@@ -15,7 +15,7 @@ message PBLink {
   // binary CID (with no multibase prefix) of the target object
   optional bytes Hash = 1;
 
-  // UTF-8 string name.
+  // UTF-8 string name
   optional string Name = 2;
 
   // cumulative size of target object
@@ -36,6 +36,12 @@ message PBNode {
 The objects link names are specified in the 'Name' field of the PBLink object.
 All link names in an object must either be omitted or unique within the object.
 
+DAG-PB aims to have a canonical form for any given set of data. Therefore, in addition to the standard Protobuf parsing rules, DAG-PB decoders should enforce additional constraints to ensure canonical forms (where possible):
+
+1. Fields must appear in their correct order as defined by the Protobuf schema above, blocks with out-of-order fields should be rejected. It is common for Protobuf decoders to accept out-of-order field entries.
+2. Duplicate entries in the binary form are invalid, blocks with duplicate field values should be rejected. It is common for Protobuf decoders to accept _updates_ to fields that have already been set.
+3. Fields and wire types other than those that appear in the Protobuf schema above are invalid and blocks containing these should be rejected. It is common for Protobuf decoders to skip data in each message type that does not match the fields in the schema.
+
 ## Logical Format
 
 When we handle DAG-PB content at the Data Model level, we treat these objects as maps.
@@ -43,42 +49,47 @@ When we handle DAG-PB content at the Data Model level, we treat these objects as
 This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
 
 ```ipldsch
+type PBNode struct {
+  Links [PBLink]
+  Data optional Bytes
+}
+
 type PBLink struct {
-  Hash Bytes
+  Hash Link
   Name optional String
   Tsize optional Int
 }
-
-type PBNode struct {
-  Links optional [PBLink]
-  Data optional Bytes
-}
 ```
 
-The first node in a block of DAG-PB data will match the `PBNode` type.
+Constraints:
 
-When decoding data with the DAG-PB codec, maps with exactly these fields will result.
+* The first node in a block of DAG-PB data will match the `PBNode` type.
+* `Data` may be omitted or a byte array with a length of zero or more.
+* `Links`:
+  * must be present, even if empty; the binary form makes no distinction between an empty array and an omitted value, in the Data Model we always instantiate an array.
+  * elements must be sorted in ascending order by their `Name` values, which are compared by bytes rather than as strings.
+* `Hash`:
+  * even though `Hash` is `optional` in the Protobuf encoding, it should not be treated as optional when creating new blocks or decoding existing ones, an omitted `Hash` should be interpreted as a bad block
+  * the bytes in the encoding format is interpreted as the bytes of a CID, if the bytes cannot be converted to a CID then it should be treated as a bad block.
+  * the data is encoded in the binary form as a byte array, it is therefore possible for a decoder to read a correct binary form but fail to convert a `Hash` to a CID and therefore treat it as a bad block.
+* When creating data, you can create maps using the standard Data Model concepts, and as long as they have exactly these fields. If additional fields are present, the DAG-PB codec will error, because there is no way to encode them.
 
-When creating data, you can create maps using the standard Data Model concepts,
-and as long as they have exactly these fields, the DAG-PB codec can encode them.
-If additional fields are present, the DAG-PB codec will error, because there is no way to encode them.
+The most recent [JavaScript](https://github.com/ipld/js-dag-pb) implementation strictly exposes this logical format via the Data Model and does not support alternative means of resolving paths via named links as the legacy implementations do (see below). The most recent [Go](https://github.com/ipld/go-ipld-prime-proto) implementation also avoids alternate pathing mechanisms but does not yet support the strict logical format.
 
-## Alternative Pathing
+## Alternative (Legacy) Pathing
 
-While the [logical format](#logical-format) implicitly describes a set of mechanisms for pathing over and through DAG-PB data,
-DAG-PB also enjoys some other special forms of pathing in addition to the Data Model norms, and these are supported by most major applications that use DAG-PB.
+While the [logical format](#logical-format) implicitly describes a set of mechanisms for pathing over and through DAG-PB data in strict Data Model form, legacy implementations afford a means of resolving paths by privileging the `Name` in links.
 
 This alternative pathing is covered here as part of this descriptive spec, but was developed independently of the Data Model and is thus not well standardized.
-It currently differs between implementations. Please see [issue #55] for more information about the harmonization effort. This section describes the current implementations as of September 2019.
+The alternative pathing mechanisms differ between implementations and has been removed from the newer implementations entirely.
 
-The Go and JavaScript implementation both support pathing with link names: `/<name1>/<name2>/…`.
+The legacy [Go](https://github.com/ipfs/go-merkledag/tree/master/pb) and [JavaScript](github.com/ipld/js-ipld-dag-pb) implementations both support pathing with link names: `/<name1>/<name2>/…`.
 
-In Go, this is the only way, which implies that is is impossible to path through nodes that don't name their links. Also neither the Data section nor the Links section/metadata are accessible through paths.
+In the legacy Go implementation, this is the only way, which implies that is is impossible to path through nodes that don't name their links. Also neither the Data section nor the Links section/metadata are accessible through paths.
 
-In the JavaScript implementation, there is an additional way to path through the data. It's based purely on the structure of object, i.e. `/Links/<index>/Hash/…`. This way you have direct access to the `Data`, `Links`, and `size` fields, e.g. `/Links/<index>/Hash/Data`.
+In the legacy JavaScript implementation, there is an additional way to path through the data. It's based purely on the structure of object, i.e. `/Links/<index>/Hash/…`. This way you have direct access to the `Data`, `Links`, and `size` fields, e.g. `/Links/<index>/Hash/Data`.
 
 These two ways of pathing can be combined, so you can access e.g. the `Data` field of a named link via `/<name/Data`. You can also use both approaches within a single path, e.g. `/<name1>/Links/0/Hash/Data` or `/Links/<index>/Hash/<name>/Data`. When using the DAG API in js-ipfs, then the pathing over the structure has precedence, so you won't be able to use named pathing on a named link called `Links`, you would need to use the index of the link instead.
-
 
 ## Canonical DAG-PB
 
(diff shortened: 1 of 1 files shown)


block-layer/codecs/dag-pb.md
```diff
@@ -44,13 +44,13 @@ This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
 
 ```ipldsch
 type PBLink struct {
-  Hash Bytes
-  Name optional String
-  Tsize optional Int
+  Hash optional Link
```
rvagg: https://github.com/ipfs/go-merkledag/blob/749fd8717d46b4f34c9ce08253070079c89bc56d/pb/merkledag.pb.go#L83-L88

I'm not sure on the NodeBuilder side though, I assume that either way you can end up with a "maybe Link, maybe empty"?

block-layer/codecs/dag-pb.md
```diff
@@ -44,13 +44,13 @@ This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
...
+  Name String (implicit "")
```
rvagg: https://github.com/ipfs/go-merkledag/blob/749fd8717d46b4f34c9ce08253070079c89bc56d/pb/merkledag.pb.go#L90-L95

can't have empty string as far as I can tell, round-trip will end up with a `""`

block-layer/codecs/dag-pb.md
```diff
@@ -44,13 +44,13 @@ This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
...
+  Tsize Int (implicit "0")
```
rvagg: https://github.com/ipfs/go-merkledag/blob/749fd8717d46b4f34c9ce08253070079c89bc56d/pb/merkledag.pb.go#L97-L102

same as `Name`, can't have an empty value here

block-layer/codecs/dag-pb.md
```diff
@@ -44,13 +44,13 @@ This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
...
 }
 
 type PBNode struct {
-  Links optional [PBLink]
+  Links [PBLink]
```
rvagg, line 53: https://github.com/ipld/go-ipld-prime-proto/blob/38f564004a47f20fe67a4c36a40aec72c0287b2b/coding.go#L37

We _could_ make this optional, but I'm happy with empty-array, it's cleaner IMO

block-layer/codecs/dag-pb.md
```diff
@@ -44,13 +44,13 @@ This layout can be expressed with [IPLD Schemas](../../schemas/README.md) as:
...
   Data optional Bytes
```
rvagg, line 54: https://github.com/ipfs/go-merkledag/blob/749fd8717d46b4f34c9ce08253070079c89bc56d/pb/merkledag.pb.go#L154-L159

rvagg:
Also pushed in an update that adds the sorting order to `Links` which we've neglected to document before. I don't know if go-ipld-prime-proto does this when encoding, maybe it defers to go-merkledag for it but if it does there might be a non-roundtrippyness problem there where you construct a DAG-PB object in memory, traverse through it assuming it's "correct" but it comes out in a different form when it does a round-trip (same problem in general with strictness in object shapes I'm trying to correct for in JS).

block-layer/codecs/dag-pb.md
rvagg: Is it possible to have a zero-length `Data` element in go-ipld-prime-proto? So zero-length and empty/nil are both valid forms? In JS zero-length is currently not allowed (it's converted to a `null`).

block-layer/codecs/dag-pb.md
mvdan: You can have an empty string. The type of `m.Name` is `*string`, so it can be `nil`, `""`, or any non-empty string. The code checks for `nil`, so you just end up with any string, empty or non-empty.

mvdan: Or maybe I misunderstand what you mean here? :)

block-layer/codecs/dag-pb.md
mvdan: The type in Go is `[]byte`, which indeed can be either `nil` or `[]byte{}` (empty and non-nil list), but I don't think the Protobuf encoding separates those two. From the proto3 language guide:

> For bytes, the default value is empty bytes.

So I think whichever you choose will always be encoded as empty bytes in protobuf, I imagine. But I'm not sure.

block-layer/codecs/dag-pb.md
rvagg: Yeah, this might need some experimental code! I have a bit of "made with Go" in my JS PB tests but maybe more would be good to sort out these specific pieces. There's really two questions I'm interested in:

1. What are the limitations on the IPLD nodes you can create for DAG-PB using go-ipld-prime-proto - can you make one that has a `nil` string as well as a `""` string and a `"foo"` string in the `Name` field?
2. What happens when you encode each of the above cases into protobufs and then decode it back into go-ipld-prime-proto; do you get all of these cases back out, `nil`, `""`, and `"foo"` or does `nil` always become `""` on such a round-trip as I'm suspecting.

And then as a follow-up - if there is variation between what you can _build_ in memory with go-ipld-prime-proto and what happens to it after you round-trip, is this acceptable? I'm taking the stance that this is not acceptable (at least for JS, but the reasoning should be the same for Go I think) because it leads to weird side-effects for users when their "nodes" are different after doing a round-trip through encode/decode. Same question for all of these fields that could be `optional` or have `implicit`s, same for the `Links` array or non-array.

I'm trying to be absolutely clear in JS and I think maybe I need to get dirty with go-ipld-prime-proto to do the same.

rvagg:
OK, did some investigation with Go to see what it's doing at the lowest level and that resulted in https://github.com/ipfs/go-merkledag/pull/58 and some winding back of most of the proposals in here. The only real one that changes the schema now is:

```
Hash optional Link
```

Plus I've added some notes below with further constraints that can't be expressed with the schema. Particularly that `Links: []` is invalid and an empty `Links` array must be `null`. There's no difference at the byte level. But it's possible to have zero-length byte arrays in both `Data` and `Name`, as well as `null` for both. I've added additional constraints around nulls and zero-length byte arrays and the validity of the `Hash` -> CID conversion.

I was going to use go-ipld-prime-proto as a reference but noticed this in the README:

> It does not include robust facilities to support actual creation of DAG protobuf nodes

It really is minimal and defers mostly to go-merkledag and has some very rough rules about what it's doing with the data. So what we define here should be the rules used when go-ipld-prime-proto is properly fleshed out so it's worth getting this right.

I've also added a section about a zero-length block being valid. Both current Go and JS implementations will decode this into `{ Data: null, Links: null }`, so it's important for implementers to be aware this is a case they may need to test for.

rvagg:
ok! I've had my head in this all week and I'm much more comfortable with where this is at now. I've reversed a bit from the previous comment, the schema uses `optional` for these fields and not a kinded union that includes `Null` so `{ Data: null, Links: null }` is not valid according to the schema unless we go with a kinded union. In Go there's `nil` for these fields, but go-ipld-prime has `Null` as a separate thing, and `IsAbsent()` can tell you whether it's there or not, it just doesn't have to be part of a graph. In JavaScript we just don't define it (or `undefined`). So:

* The only major change to the schema as making the `Hash` optional. I don't see a way around this, an empty `PBLink` is possible in the binary format, who knows how many of those are in the wild. `{ Links: [{}] }` is valid.
* I've added a bunch of constraints clarifying things:
  - one notable one is that the `Hash` bytes _must_ be a valid CID. I think that's reasonable, I believe all implementations are treating them as such so even if there are blocks in the wild with bad `Hash` values, they should be rejected already.
  - empty `Links` array is indistinguishable from an omitted one in the binary format, that's an important clarification (and this is different to an empty `Data` byte array which is distinguishable from an omitted one).

Implementation stuff:

* Go "forms" tests is here: https://github.com/ipfs/go-merkledag/pull/58 - it allows a few things that we should disallow in go-ipld-prime-proto - the zero-length Links case (indistinguishable from no Links) and the non-CID Hash.
* New JS implementation is now fully compatible with this spec, it's strict on the way in, and mostly strict on the way out (won't check sorting of Links array is the only bit of non-strictness). It also has a `prepare()` method that lets you re-shape a sloppy input object into a precise Data Model form, and `encode()` is super-strict so you won't end up having an in-memory representation that's different from its round-trip version. I also have a mirror of the "forms" tests to validate these edge cases and their binary representations. https://github.com/ipld/js-dag-pb

rvagg:
OK, after sleeping on it and then talking it through this morning in the meeting and with Mikeal I've decided to even wind back the `optional` for `Hash`. So the only real change to the schema text is that it _must be a `Link`_, that is, it must be interpreted as a `CID` and anything else is invalid. I've added this to my notes, that a block should be rejected. Now a `[{}]` is not a valid `Links` array.

I've also added that a block with "unrecognized" data should also be rejected. This is a _should_ because most PB decoders will silently accept extraneous data. In go-merkledag there's a `XXX_unrecognized` property on both nodes to capture it, but any data in there should signal a bad block.

I'm about to go enforce this on the new JS DAG-PB library. Maybe we'll find some source of data out there that has these bad blocks, but until we see them in the wild we can afford to be strict in our newer implementations.

block-layer/codecs/dag-pb.md
```diff
@@ -36,50 +36,58 @@ message PBNode {
...
 
 ```ipldsch
-type PBLink struct {
-  Hash Bytes
-  Name optional String
-  Tsize optional Int
-}
-
 type PBNode struct {
   Links optional [PBLink]
```
warpfork: If I understood correctly that the protobuf wire format doesn't disambiguate between a lack of a list here vs a zero length list... maybe it would be best to describe it in our schemas as `Links [PBLink]` _without_ the optional modifer on the field?

If we have the `optional` modifier on the field, that means it _adds_ `Absent` to the cardinality (it +1's the cardinality), but we really have no way of *taking away* the empty list as a valid value (if we could, that'd be -1 on the cardinality, and balance things -- but we don't have such an ability).

Just having the codec handle the translation from {empty list}<->{absense of encoded bytes} seems correct.

block-layer/codecs/dag-pb.md
warpfork: I concur that I get the impression from [the protobuf language guide](https://developers.google.com/protocol-buffers/docs/proto#optional) that the wire format will not disambiguate between an absent value here and an empty byte sequence.

I think the [same logic I just used about empty lists](https://github.com/ipld/specs/pull/297/files#r494338322) applies here as well: we might be best off not using the `optional` modifier here.  I don't emotionally _like_ this, because it would feel a lot "cleaner" if our IPLD'ified views could give a clear `Absent` value, but given that we don't have a way to say "empty byte sequences are invalid" and given that the serial format doesn't disambiguate these, I think it's the most logically consistent choice we can make.

rvagg: Not quite, `0x0a00` means - "field 1, wire type 2 (bytes), length 0": `{Data:[]bytes{}}`

Where `0x1200` means - "field 2, wire type 2 (bytes), length 0", but for field 2 it has to drop into parsing the `Links` array with a zero byte array and that yields a single empty link: `{Links:[{}]}`

So in terms of cardinality you can't avoid the absent, zero-length, non-zero-length cases for `Data`.

For `Links` you can have absent but you can't have zero-length.

block-layer/codecs/dag-pb.md
rvagg: Yeah, so I've been wanting to lean this way but the thing that's stopped me is that it creates a lot of situations where you're forcing the use (and allocation) of an empty list where you don't want anything. For decoding a dag-pb block with just data in it you're going to need to throw in an extra array just to satisfy this case. When creating an object manually in our APIs you're going to have to allocate an array. In JavaScript this means that I have to force users to give me `{ Data: [...], Links: [] }` when they're just wanting to ingest terminal data blocks. This isn't a huge inconvenience as I've created this `prepare()` method to help get the forms right before instantiating, but it adds a level of hostility that I'm trying to avoid. I think in ipld-prime you'd have data structures that have an empty array for `Links` when you create a new `PBNode` anyway so I suppose you get around this hostility fairly easily.

But.. I take your point on cardinality. And the reason why I've been tempted in this direction is that there's something to be said on the other end of this API where you _always_ get an array in that position and don't need to check for its existence. If I can write a traversal that safely assumes `Links` is always there then it's much nicer than the incessant checking.

I'm going to ponder this today and maybe try out some code.

rvagg:
OK, I'm going with @warpfork's [argument about `Links` cardinality](https://github.com/ipld/specs/pull/297#discussion_r494338322). An omitted one in the binary form will be interpreted as an empty array. In the Data Model the array must always be present.

I've done another pass of the doc and cleaned up the clarifications to the schema. I've also added notes in the Protobuf schema about our recommendations for strict decoding.

https://github.com/ipld/js-dag-pb now matches this. I'm going to work on a Go form too I think now that I have a good test suite but it'll just be a simple replacement to go-merkledag/pb, I don't think I can manage the ipld-prime piece of it and the question of whether to integrate it in ipld-prime-proto is still up for discussion. So I'll just go with the low-level pieces for now.

PR #307, opened 2020-10-02
Add exploration report about Numbers and IPLD Codecs

vmx:
This is an exploration report about how to serialize IPLD Data Model numbers
into bytes and back.

That report got very different from what I originally thought (which is great). It
got way simpler as all the problems I had are just due to JavaScript ;)

### DIFF
--- a/design/history/exploration-reports/2020.10-data-model-numbers-and-codecs.md
+++ b/design/history/exploration-reports/2020.10-data-model-numbers-and-codecs.md
@@ -1,0 +1,77 @@
+IPLD Data Model numbers and Ipld Codecs
+=======================================
+
+**Author**: Volker Mische ([@vmx])
+
+This is an exploration report about how to serialize [IPLD Data Model] numbers into bytes and back.
+
+
+Canonical representations
+-------------------------
+
+In IPLD it's desirable to have one single representation for every instance of the Data Model. This would lead to a bijective mapping between the Data Model and the encoded bytes. If you have a Data Model instance and you use a specific codec, it should always result in the same bytes, independent of the programming language used. That's important for content-addressing.
+
+
+### Conversion into the Data Model
+
+This looks like a hard problem, e.g. if you look at this IPFS issue with [DAG-CBOR] where there is a [different encoding between JavaScript and Go when numbers are integers]. Though in this issue, the problem is **not** the Data Model to Codec conversion. The problem is the **input data to Data Model conversion**. It took me a long time to figure that out.
+
+When I think of the IPLD Data Model I often think in terms of JSON and having numbers represented as text. I then think about how to represent those things as bytes. If you think about it that way, you have problems like: is `4251.00` and `4251` the same number? If yes, then it should be encoded the same way?
+
+From a Data Model perspective that's a non-issue. You have two kinds of numbers: [Integer] and [Float]. You can clearly distinguish between `Integer(4251)` and `Float(4251)`.
+
+
+### Serializing the Data Model
+
+
+#### Codec with multiple integer/float sizes
+
+Even within the Data Model there might be ambiguities, depending on Codec you use. If your Codec supports more than one integer type (which is often the case), you need to decide which one to choose. The obvious way is what CBOR in its [canonical representation] is doing:
+
+> Integers must be as small as possible.
+
+This means that a codec that supports 8-bit to 64-bit integers will encode `8472` always as `0x2118` and never as `0x00002118`.
+
+For floating point numbers (I always mean [binary IEEE-754 floating point numbers], others don't matter in modern day computing) the same issue arises. If your codec supports more than one floating point number type, e.g. 32-bit single precision and 64-bit double precision, a number like [`0.199951171875`] could be represented as `0x3e4cc000` (32-bit) or `0x3fc9980000000000` (64-bit). You can, just as with integers, require the smallest lossless representation of a float.
+
+Such a conversion to the smallest possible float makes sense for the sake of having a canonical representation, it doesn't make sense for space savings. Only about 0.00000002% of all the 64-bit floats can be represented as 32-bit floats. So it's not that likely that your data contains exactly those floats.
+
+
+#### Codecs with one type for integers and one for floats
+
+If you restrict your Codec to a single integer and a single float type, the conversion from the Codec into the IPLD Data Model becomes much simpler, as you obviously don't have to deal with different types.
+
+For integers, you could just a sensibly large integer, e.g. 64-bit and always serialize only into that. In case you want to save bytes, you can use variable sized integers like [LEB128].
+
+For floats the only sensible way is using IEEE-754 double precision float only. They are well supported by almost all programming languages.
+
+
+Problems with JavaScript
+------------------------
+
+Creating a canonical representation for numbers is possible when you distinguish between integers and floats. The IPLD Data Model has that property, hence it's not a problem designing a Codec that has a canonical representation of numbers. Though, there is the problem with JavaScript. Currently, most (all?) IPLD Data Model implementations work with native JavaScript types without any wrappers around them. The problem is that historically JavaScript only has a [`Number`] type, which is always a 64-bit IEEE-754 float. You cannot distinguish between `4251.00` and `4251`. Both are always floats.
+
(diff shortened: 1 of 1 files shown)


design/history/exploration-reports/2020.10-data-model-numbers-and-codecs.md
```diff
@@ -0,0 +1,77 @@
...
+
+> Integers must be as small as possible.
+
+This means that a codec that supports 8-bit to 64-bit integers will encode `8472` always as `0x2118` and never as `0x00002118`.
+
+For floating point numbers (I always mean [binary IEEE-754 floating point numbers], others don't matter in modern day computing) the same issue arises. If your codec supports more than one floating point number type, e.g. 32-bit single precision and 64-bit double precision, a number like [`0.199951171875`] could be represented as `0x3e4cc000` (32-bit) or `0x3fc9980000000000` (64-bit). You can, just as with integers, require the smallest lossless representation of a float.
+
+Such a conversion to the smallest possible float makes sense for the sake of having a canonical representation, it doesn't make sense for space savings. Only about 0.00000002% of all the 64-bit floats can be represented as 32-bit floats. So it's not that likely that your data contains exactly those floats.
```
rvagg, line 37: Hah! that's a tiny percentage. I wonder if there's problems translating that into practice and whether "in practice, the types of floats generally represented by computer programs fall within .... certain bounds ...". Maybe in practice it's not as lop sided as this.

But this whole "smallest possible float" representation in CBOR sure is a major hassle and quite wasteful during encoding because you need to try encode+decode in 16, then 32 and then fall back to 64. With the most likely path, as you've suggested, being 64. Which is why I keep on coming back to this problem and wondering whether we're doing it wrong in JS or wrong in Go ... perhaps always-64 is the right choice?

rvagg (approved):

design/history/exploration-reports/2020.10-data-model-numbers-and-codecs.md
vmx: Using the "smallest possible float" makes sense for CBOR itself, but I don't think it makes sense for **DAG**-CBOR. I think we should switch to "always 64-bit". I think it would make encoders and decoders way simpler.

design/history/exploration-reports/2020.10-data-model-numbers-and-codecs.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+### Conversion into the Data Model
+
+This looks like a hard problem, e.g. if you look at this IPFS issue with [DAG-CBOR] where there is a [different encoding between JavaScript and Go when numbers are integers]. Though in this issue, the problem is **not** the Data Model to Codec conversion. The problem is the **input data to Data Model conversion**. It took me a long time to figure that out.
+
+When I think of the IPLD Data Model I often think in terms of JSON and having numbers represented as text. I then think about how to represent those things as bytes. If you think about it that way, you have problems like: is `4251.00` and `4251` the same number? If yes, then it should be encoded the same way?
+
+From a Data Model perspective that's a non-issue. You have two kinds of numbers: [Integer] and [Float]. You can clearly distinguish between `Integer(4251)` and `Float(4251)`.
```
mikeal, line 21: Should we consider changing this? We could just say “numbers” instead of integer and float. At the end of the day, the data model just tries to de-serialize into common types, but if a specific type difference is required in a specific language and application that’s going to be mapped on above our Data Model. It’s not in conflict with the Data Model to de-serialize all integers as BigInt if that’s what the application requires, nor is in conflict to convert a list of integers to floats if what you need to do on them is floating point math. It may be better/simpler to say “number” is the Data Model type and leave it up to languages to figure out how each serialization maps to their type system.

That would leave us free to say “there is only one way to encode 4251 and you can’t do it as a float” in formats we design for a very high degree of hash consistency. **Formats** may still distinguish between integers and floats, and that’s fine, but the Data Model doesn’t have to.

vmx: That's a strong "no we don't change this" from me. It is soooo good that we have this distinction in the Data Model. JavaScript is the one who is wrong here. I'd hope that this exploration report makes that clear. In case it doesn't I'm happy to see if I can put together a follow up report.

vmx:
We can still keep discussing things, I'm just merging this one so that I can link to it.

design/history/exploration-reports/2020.10-data-model-numbers-and-codecs.md
mikeal: I’m not going to try and defend JavaScript’s number type system, it’s fairly indefensible, but for me this isn’t about JavaScript and more about what the data model’s view of “lossless” means.

Finding the smallest binary representation of a number is a lossy operation from the point of view of almost any number type system. Even VARINT loses some type information you may have had on your numbers that you’ll have to add back if you want to maintain the same integer type. For instance, if you have a 64b integer type of a small number value it’s probably going to round-trip to a smaller number type that you’ll then need to convert back to 64b.

Depending on the language, it’s often going to be necessary to do some form of type conversion to satisfy the needs of the application that is working with the de-serialized state.

So the question is, does unifying these into a single number type incur a form of loss that makes the type effectively unusable? That answer might be yes, I’m leaning towards yes (in agreement with you), but I want to explore exactly why.

At the end of that day, this is all just serialized binary data that we’re applying logic to in order to convert to useful types. With integers, we have pretty simple and recurrent patterns for working with any loss the format may cause during serialization. I don’t think we have that for floats, and as I keep being reminded, floats are **NOT INTEGERS** and can’t be represented effectively if you’re trying to normalize and reconcile them with integers. They are effectively a different type from integers in **every** programming language, ~~including JavaScript (Float32Array),~~ because there’s just no way for a language to work around the fact that the math operations for them are implemented by a separate set of instructions.

vmx: > Finding the smallest binary representation of a number is a lossy operation

Yes. It depends on the definition of "lossy". I wouldn't call it lossy, I would call it mutation or transformation. Though I totally get what you mean.

>  if you want to maintain the same integer type

This exploration report is about the Data Model only. And that's important here. I've also mixed the Data Model end native language encoding in my mind in the past, but I think we shouldn't. If you think about the Data Model only, there are no different integer types, there is only a single one.

> floats are **NOT INTEGERS** and can’t be represented effectively if you’re trying to normalize and reconcile them with integers

I'm not sure what you mean. Converting an integer to a varint is the same kind of transformation as if you convert it to a double (keeping the bit size limitations in mind, e.g. < 53 bit for a 64-bit float).

rvagg: This is really about where we want to draw a line between utility and fidelity. We can use the same logic in this discussion to further explore:

 * Why do we bother with a `String` type? They're all just byte arrays in the end and strings are only really interesting in that they can represent things that humans can read. They're essentially sugar with weird rules for conversion (utf-8).
 * Why bother with a `Boolean` type? C has made do with `0` and `1` forever, JavaScript makes good use of "truthy" and "falsy" in the same spirit.
 * Maps really are a pain, we should rule them out in the representation layer and rely on Schemas to _transform_ them for users from tuple representation (and other) into maps in memory. This is what Filecoin is doing after all and it really does solve a lot of problems and saves space when encoded.

I think it's legitimate to argue for `Number` over `Int` and `Float` as long as we're boiling down all of `int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`, `uint64` into a single type, why not throw `float16`, `float32` and `float64` into the same basket? I know @warpfork's argument here is about cardinality - when you treat `1.0` and `1` as the same thing you reduce cardinality and lose information. But you're doing that anyway when you reduce all of the int types to `Int`.

But from a utility perspective, I think we JavaScript programmers should be the ones arguing that greater utility lies in differentiating between integers and floats, `x % 1 === 0` _is a **really** crappy thing to have to do_ (same goes for `Number.isInteger()` which is just sugar over 💩) and `Number.MAX_SAFE_INTEGER` is equally crappy.

"My system deals with floating point numbers in _this_ place and deals with integers in _that_ place" is pretty useful. I also like that we can shove floats into their own box and push it off to the side, and recommend that users to likewise with serialisation. `Number` is just a lowest common denominator which ends spoiling integers.

PR #308, opened 2020-10-05
DAG-JSON: be more specific about numbers

vmx:
The IPLD Data Model distinguishes between integers and floats,
so should DAG-JSON. The easy way is to do what popular dynamically
typed programming languages are doing. If the number only contains
digits, it's an integer, if it contains a decimal point, it's a
float.

This PR was triggered by my exploration report about numbers [1],
where I realized that there are also underspecified things when
parsing DAG-JSON as input.

[1]: https://github.com/ipld/specs/blob/19e572e4f0b71ed89b5eb4ea3132a1398cb96c05/design/history/exploration-reports/2020.10-data-model-numbers-and-codecs.md

### DIFF
--- a/block-layer/codecs/dag-json.md
+++ b/block-layer/codecs/dag-json.md
@@ -22,7 +22,7 @@ All kinds of the IPLD Data Model except Bytes and Link are supported natively by
 
 #### Numbers
 
-Numbers are a special case. JSON only has a single number type, though many dynamically typed languages (e.g. Python, Ruby, PHP) distinguish between integers and floats when parsing JSON. A number consisting only of digits is parsed as integer, if it contains a dot, it's parsed as a float. In DAG-JSON the same method is used to represent integers and floats.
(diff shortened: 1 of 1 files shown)


block-layer/codecs/dag-json.md
```diff
@@ -20,9 +20,14 @@ producing the same data end up with matching block hashes.
 
 All kinds of the IPLD Data Model except Bytes and Link are supported natively by JSON.
 
+#### Numbers
+
+Numbers are a special case. JSON only has a single number type, though many dynamically typed languages (e.g. Python, Ruby, PHP) distinguish between integers and floats when parsing JSON. A number consisting only of digits is parsed as integer, if it contains a dot, it's parsed as a float. In DAG-JSON the same method is used to represent integers and floats.
```
mvdan: for the sake of consistency, perhaps wrap this paragraph like the others?

mvdan (approved):
SGTM. Have you thought about signs? Namely, the JSON `-123` should still be considered an integer, and it's more than just digits. Be mindful of `-0` though, which technically needs to be a float to not be collapsed into the (positive) `0`.

block-layer/codecs/dag-json.md
vmx: ~I wanted to keep the diff small.~

vmx: Sorry, you meant the opposite. There are already paragraphs that aren't wrapped. So it's mixed already anyway.

warpfork (approved):

PR #310, opened 2020-10-12
dag-cbor: more strictness items and additional clarifications

rvagg:
Some additional items that have occurred to me during my weekend cbor hacking. Some of these changes don't add rules, it's just further clarity.

* Clarify position on tags - just because the CBOR spec mentions some official-looking tags doesn't mean they're supported. And just because some of the tags are used for seemingly "reasonable" things (like bignums and bigfloats and other ways of doing bytes) doesn't mean they're supported.
* "Simple values" are not supported. There's a rabbit hole in major type 7 so I included an exclusive list just for that.
* `Undefined` shouldn't be supported .. which may be a surprise to some JS folks, it's not in our data model and we can't have it in our encodings. (This should be clarified further in our Data Model docs, we have to rule out any kind of sparse arrays for instance, `['foo',,,'bar']`. That shouldn't be possible but currently is in js-*-dag-cbor:

```
> dagCbor.util.deserialize(dagCbor.util.serialize([1,,,,3]))
[ 1, undefined, undefined, undefined, 3 ]
```

(JSON does the right thing here in JS btw and coerces them to `null` but coercion is probably the wrong thing to be doing there!).

### DIFF
--- a/block-layer/codecs/dag-cbor.md
+++ b/block-layer/codecs/dag-cbor.md
@@ -20,10 +20,10 @@ DAG-CBOR uses the [Concise Binary Object Representation (CBOR)] data format, whi
 
 ## Format
 
-The CBOR IPLD format is called DAG-CBOR to disambiguate it from regular CBOR. Most CBOR objects are valid DAG-CBOR. The primary differences are:
-  * tag `42` interpreted as CIDs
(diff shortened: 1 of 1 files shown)


vmx (approved):
Thanks. I'm currently operating in a similar space, so I fully support those changes.

warpfork (approved):
:100:

PR #311, opened 2020-10-12
dag-cbor: resolve the float question - 64-bit

rvagg:
(PR on a PR, apologies)

Let's resolve this! I think we're all coming around to the always-64-bit representation being the appropriate course of action. It's both easier and most likely to be safest over time. It can be wasteful for some common numbers, but I suspect the Venn diagram of people concerned about byte shaving and those wanting floats in their blocks might have a very small intersecting area.

### DIFF
--- a/block-layer/codecs/dag-cbor.md
+++ b/block-layer/codecs/dag-cbor.md
@@ -44,9 +44,9 @@ Therefore the DAG-CBOR codec must:
 1. Use no tags other than the CID tag (`42`). A valid DAG-CBOR encoder must not encode using any additional tags and a valid DAG-CBOR decoder must reject objects containing additional tags as invalid.
    * This includes any of the initial values of the tag registry in [section 2.4 of the CBOR specification], such as dates, bignums, bigfloats, URIs, regular expressions and other complex, or simple values whether or not they map to the [IPLD Data Model].
 2. The only usable major type 7 minor types are those for encoding double-precision Floats (`27`), True (`20`), False (`21`) and Null (`22`).
-	 * "Simple values" are not supported. This includes all registered or unregistered simple values that are encoded with a major type 7.
-	 * Undefined (`23`) is not supported.
-	 * All floating point numbers must be encoded as double-precision (64 bit). Half, and single precision encoding is not supported.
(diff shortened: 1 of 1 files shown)


vmx (commented):
Could you somewhere mention IEEE-754 to be extra precise? Perhaps even "binary double-precision IEEE-754 floats (64-bit) or so?

rvagg:
done, pushed fixup at 764c992:

> DAG-CBOR encodes all floating point numbers as the binary form of double-precision floats (64-bit IEEE 754); i.e. as major type 7, minor 27.

block-layer/codecs/dag-cbor.md
```diff
@@ -43,32 +43,34 @@ Therefore the DAG-CBOR codec must:
...
+	 * All floating point numbers must be encoded as double-precision (64 bit). Half, and single precision encoding is not supported.
+3. Use most of the canonical CBOR encoding defined by the suggestions in [section 3.9 of the CBOR specification]. A valid DAG-CBOR decoder should reject objects not following these restrictions as invalid. Specifically:
    * Integer encoding must be as short as possible.
    * The expression of lengths in major types 2 through 5 must be as short as possible.
    * The keys in every map must be sorted lowest value to highest. Sorting is performed on the bytes of the representation of the keys.
      - If two keys have different lengths, the shorter one sorts earlier;
      - If two keys have the same length, the one with the lower value in (byte-wise) lexical order sorts earlier.
    * Indefinite-length items are not supported, only definite-length items are usable.
+	 * Floating point numbers _do not_ follow the smallest-possible representation rule but use a consistent 64-bit encoding (see above).
```
warpfork: This bullet seems too indented to me?

Ah.  You've been smitten by mixed tabs and spaces.

warpfork (approved):
LGTM.

I had minor qualms about diverging from the CBOR canonicalization recommendation, but I've come to agree with the team's prevailing opinion that the number of cases where this matters is low, and the implementation complexity cost is unnecessarily high for such little utility gained.

block-layer/codecs/dag-cbor.md
rvagg: arghflkahjdl! this is the cost of doing both JS & Go with the same editor ... fixed, thanks

PR #313, opened 2020-10-13
fix link and expand on map basic type.

prataprc:
Since the default diff is line-oriented, and this markdown has each paragraph in a single line, this might help:

* [gh](https://cli.github.com/) command is needed.
* cd ipld/specs/
* gh pr checkout 313
* gh diff --color-words HEAD~

Normally I create an alias `diffw` for `diff --color-words`.

### DIFF
--- a/data-model-layer/data-model.md
+++ b/data-model-layer/data-model.md
@@ -7,10 +7,11 @@ which describes what data is representable in IPLD --
 for example, booleans, integers, textual strings, maps and lists, etc.
 
 While the Data Model describes these representations in the abstract,
-[Codecs](Codecs) specify exactly how these data are transcribed into serialized bytes.
-(Another component of the IPLD specifications, [Schemas](../schemas), provide
-additional optional tooling on top of the Data Model which can further refine,
-describe, and constrain the range of acceptable data values.)
+[Codecs](../block-layer/codecs) specify exactly how these data are transcribed
+into serialized bytes. (Another component of the IPLD specifications,
+[Schemas](../schemas), provide additional optional tooling on top of the Data
+Model which can further refine, describe, and constrain the range of acceptable
+data values.)
 
 Motivation
 ----------


data-model-layer/data-model.md
```diff
@@ -113,7 +114,8 @@ Values contained in lists can be accessed by their ordinal offset in the list.
 Map is a recursive kind.
 
 Values in maps are accessed by their "key".  Maps can also be iterated over,
-yielding key+value pairs.
+yielding key+value pairs, and such iteration shall not assume any order to
```
rvagg: oh, this is treading into contested waters here, I don't know if we have agreement yet about map ordering while in the data model, we still don't even have agreement about the _right_ way to view ordering at the codec layer.

This _might_ be safe to say, I'd certainly want to reserve this property for some scenarios, but I don't think we can assert that there isn't "any order" (in the Go sense of "no stable ordering for you!").

rvagg:
Something funky happened to your branch @prataprc, would you mind doing a rebase and force push to clean it up? I don't think the wording you have about iteration order is going to work, there's something to be said about ordering but it's not that the order **isn't** stable. So maybe pull that bit out for now and we'll get the link fix in first.

warpfork:
Not... sure what's going on here with the history, seems like maybe a rebase gone funky has made this appear to contain a lot more commits than intended... but that aside...

I'd be opposed to the change that Rod already highlighted.  Many things *do* have defined iteration orders.  Some systems, like Selectors, also depend on stability (n.b., only stability -- not necessarily sortedness) in iteration orders in order to expose predictable/deterministic behavior at higher levels.  (And that, yet further, is sometimes critical to performance -- for example, Graphsync, which builds on Selectors, would have serious issues with memory required for state tracking if it had to receive information in unpredictable orders.)

We need to get this into the specs repo, but my latest writing on the ordering situation can currently be found here as a gist if this helps clarify anything: https://gist.github.com/warpfork/9315eac9848fa2c2ab36174272a80d9b

prataprc:
My bad, was trying to slice the modifications per file and track it via separate branches.
Thanks for the clarification about order preserving maps.

rvagg:
:thumbsup: thanks

PR #314, opened 2020-10-13
hashmap: clean up references to hashAlg string

rvagg:
The additional clean-up I suggested in the original PR changing it from string to int.


warpfork (approved):
I am indifferent to this but approve :)

mvdan (approved):

PR #315, opened 2020-10-13
hashmap: schema changes to root node, tuple nodes, use of 'Any'

rvagg:
BREAKING change to the block representation in two ways:
 - nested root node, now within `hamt` property (I can't recall who suggested this but the name is a nice clue about what the block contains if you're looking at the map at the root).
 - tuple representation for nodes

Also (cosmetic):
 - use of 'Any' instead of custom 'Value' that does the same thing

### DIFF
--- a/data-structures/hashmap.md
+++ b/data-structures/hashmap.md
@@ -101,15 +101,14 @@ See [IPLD Schemas](../../schemas) for a definition of this format.
 type HashMapRoot struct {
   hashAlg Int
   bucketSize Int
-  map Bytes
-  data [ Element ]
+  hamt HashMapNode
 }
 
 # Non-root node layout
 type HashMapNode struct {
   map Bytes
   data [ Element ]
-}
+} representation tuple
 
 type Element union {
   | HashMapNode map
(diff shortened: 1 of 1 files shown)


rvagg:
note that the first node is inline, so they look like this:

root:

```
{
  "hashAlg": 18,
  "bucketSize": 3,
  "hamt": [
    <bytes>,
    [ ... data ... ]
  ]
}
```

whereas intermediates and leaves look like:

```
[
  <bytes>,
  [ ... data ... ]
]
```

mvdan (approved):
Nice!

data-structures/hashmap.md
```diff
@@ -332,21 +320,10 @@ type Bucket [ BucketEntry ]
 
 type BucketEntry struct {
   key Bytes
-  value Value
+  # There is currently no limitation on the types available for storage as long
```
warpfork: Tiniest nit: I think this would read a little better as prose after this code block than as a comment on the value field.

warpfork (approved):
Reads wonderfully.  :+1:

data-structures/hashmap.md
rvagg: done

PR #316, opened 2020-10-13
hashmap: switch default hashAlg to SHA2-256

rvagg:
I kind of want to see some performance numbers on this and am tempted to add back in Murmur as an optional but explain the reasons why you might choose one over the other. Jeromy said (regarding this switch that Filecoin made) that SHA2-256 was a bit faster than Murmur on the CPUs that Filecoin has been focusing on (which have SHA2 instructions). But for JavaScript, the delta could be quite large in the other direction but I just don't know. Certainly if you want a cryptographically secure HAMT then go SHA2-256 ... and maybe a secure default is something we should be shipping anyway because these kinds of choices are nuanced—hash collision attacks are a consistent feature of programs and runtimes and we don't seem to be getting better at identifying where they are likely to show up, maybe we do our users a favour by opting for secure by default.

### DIFF
--- a/data-structures/hashmap.md
+++ b/data-structures/hashmap.md
@@ -255,7 +255,7 @@ These defaults are descriptive rather than prescriptive. New implementations may
 
 ### `hashAlg`
 
-* The default supported hash algorithm for writing IPLD HashMaps is the x64 form of the 128-bit [MurmurHash3](https://github.com/aappleby/smhasher) (identified by the multihash name ['murmur3-128'](https://github.com/multiformats/multicodec/blob/master/table.csv)). Note the x86 form will produce different output so should not be confused with the x64 form. Additionally, [some JavaScript implementations](https://cimi.io/murmurhash3js-revisited/) do not correctly decompose UTF-8 strings into their constituent bytes for hashing so will not produce portable results.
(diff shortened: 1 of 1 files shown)


mvdan:
No strong opinion here, but I agree that we should be secure by default and drive this decision with benchmarks.

warpfork:
> maybe we do our users a favour by opting for secure by default.

:100: 

IMO DoS vectors are perfectly real threats, and especially so in data structures that are containing relatively "untrusted" user input... which is the case a decentralized data structure *always* finds itself in.  Using cryptographic hashes for the sharding is a very meaningful improvement to the stability of this structure, and we should do it -- if not always, then at least *absolutely* by default.

warpfork:
This seems like almost an auto-merge to me, reviewing this again today?

We kinda... [really don't like murmur](https://github.com/multiformats/go-multihash/issues/135), it turns out, and hadn't even specified _what that is_ particularly well when we looked at it again recently.  So I'm _extremely_ enthusiastic about removing it from this spec file in favor of _literally anything_.

I think we're also ending up with implementations parameterizing this pretty consistently, so the text about what's recommended here isn't especially load bearing and there just isn't a reason to be giving a massive amount of caution around change.

PR #317, opened 2020-10-13
Fix typos and few rewording.

prataprc:
Sharing my thoughts via these PRs. Since by far this seem to be a passive and non-intrusive way to touch base with specification authors.

### DIFF
--- a/concepts/type-theory-glossary.md
+++ b/concepts/type-theory-glossary.md
@@ -128,7 +128,7 @@ Product types usually show up in prose descriptions of data as the word "and".
 
 For example, say: "I have a boolean and another boolean and another boolean" --
 you might write code for this as `struct{ a bool, b bool, c bool }`,
-and its cardinality would be "|2*2*2|" -- "|8|" --
+and its cardinality would be "|2\*2\*2|" -- "|8|" --
 two for each of the booleans, and multiplied because each of them may vary independently,
 and thus each of the booleans adds another "dimension" to the state space described by this combined value.
 
@@ -138,7 +138,7 @@ and "Furniture" (it has members "chair", and "desk") --
 we could say "I have an Animal, *and* it's on top of a Furniture" --
 and then I could have any one of of *six* states
 ("cat on a chair", "cat on desk", "dog on a chair", etc):
-the cardinality is "|3*2|".
+the cardinality is "|3\*2|".
 
 ### infinite recursives
 
@@ -179,7 +179,7 @@ you have a [sum type](#sum-types) (more specifically, a [maybe type](#maybe-type
 and even if the cardinality of string is "infinite", it's still useful to declare
 the cardinality of your string is "a" (an arbitrary placeholder),
 and then the cardinality of "string or null" is "|a+1|".
-(In practice: how often have you had an API bug burn you on the different
+(In practice: how often have you had an API bug burn you on the difference
 between null and an empty string, or the difference between null and absent?
 This is why we keep counting cardinality even after an infinity shows up!)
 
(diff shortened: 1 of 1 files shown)


concepts/type-theory-glossary.md
```diff
@@ -216,8 +216,13 @@ This "open membership" is commonly known as **interfaces**.
 ### interfaces
 
 Interfaces have the concept of "open membership", and are also typically
-combined with some definition of "features" that are required if something
-will be considered to implement the interface.
+combined with some definition of "features", also called as traits or behavior
+in some languages, that are required if something will be considered to
+implement the interface.
+
+Like all values, an interface also have well-defined memory footprint in
```
rvagg: "an interface will also have ..."

rvagg (changes requested):
seems good to me sans small wording nit
@warpfork was originally responsible for this doc so his eye over it would be good

concepts/type-theory-glossary.md
warpfork: This paragraph is a bit... far altogether.  I'd think I'd rather take it back out.

The rest of this document stays pretty far away from talking about memory layouts (certainly at least until it gets to the "in language X..." parts, much later), so it's odd to bring up memory footprints here.

I also just don't think I'd say interfaces have a well-defined memory footprint in a program.  That's a pretty expansive claim to try to make for all programing languages.

warpfork (changes requested):
I like most of this, and thank you for the good eye on the typos (and formatting!).  Do you think we could take that paragraph about memory back out though?  Sans that, I'd be happy to merge this :)

prataprc:
Thanks for the review. Removed the paragraph and re-based it.

rvagg (approved):

concepts/type-theory-glossary.md
```diff
@@ -403,6 +404,6 @@ The core features of IPLD Schemas map very closely to the categories of types we
 IPLD Schemas don't have a concept of [interfaces](#open-vs-closed-membership).
 At least, *not exactly* -- not *within* the schema.
 Instead, the *whole Schema* is treated as an interface:
-since you can take some data and take a schema, and attempt to unify them to get typed data if the schema "matches"...
+since you can take some data and take a schema, and attempt to unify them to get typed value if the schema "matches"...
```
warpfork: So sorry, I found one more when I read it again this morning.  I think this either needs to be "data" or "values" (plural).  I think I'd rather keep it "data" to agree with the earlier part of the sentence :)

prataprc: No worries :). I have changed it back to _data_.

Anyhow, working with Rust past few years, I found this nomenclature,

* _data_, one or more opaque blocks of bytes.
* _value_, data with its semantics understood.
* _type_, set of values belonging to same category.

feels more precise. If I understand correctly, in this context when we have data and schema unified, i.e, when data matches a schema, we are effectively instantiating a value out of that data, whose type is fully understood by the implementing language.

Btw, one more ask related to **schema-matching**, is this something similar to erlang/haskell pattern matching ? If so, both shape of the schema and type of its field considered ? If there is a document explaining the schema matching exist, we can link it here.

warpfork: Ah, I kind of like that nomenclature when you explain it like that.

It has somewhat limited ability to be communicated by implication, I'm afraid, because some of those words are so synonymous colloquially that a reader might not reliably be able to detect from usage that the words are meant to be load-bearing of different meaning.  But when it's laid out clearly it's very nice.

I'll think more on this in the future.  :)

warpfork:
Thanks much :D

concepts/type-theory-glossary.md
warpfork: Oh!  Regarding schema-matching: there's some documentation available in this in the vicinity of https://specs.ipld.io/schemas/migration.html#version-detection .  It's likely that documentation is due for some expansions too, though!  And it's probably going to feel somewhat similar to pattern matching in various programing languages, yep.

It's all structural -- there really isn't any types associated with data until _after_ this matching happens, so structural information is all it's got to work with.

(I guess it would be possible to perform schema matching *again* on some other data that had already gone through one schema... but it would be a little odd and I wouldn't expect people to often find any reason to do this in practice.  And the main reason it happens to work is that even schema-processed data still implements the same interfaces as the raw Data Model... so, it would only be reasonable that even if someone did try to do a schema match on other already-schema'd data, it should still treat it as the data alone, for consistency.)

PR #322, opened 2020-10-20
hashmap: remove support for inlined child nodes & resolve kinded union problem

rvagg:
The original design had the ability to inline child nodes, so you could end up with a partial or complete inlined structure in a single block. But I have no code that actually does this and in practice I'm not considerably more sceptical that there is even a use-case for this. If you wanted to inline data that was stored in a HAMT into a single block, you could just use a data model map and save a lot of space and lot of complexity.

This resolves the broken kinded union that was introduced in the recent change to this spec in #315 where it had a `map` for a `tuple` struct but also already had a `list`, so it just didn't work anyway.

### DIFF
--- a/data-structures/hashmap.md
+++ b/data-structures/hashmap.md
@@ -111,7 +111,6 @@ type HashMapNode struct {
 } representation tuple
 
 type Element union {
-  | HashMapNode map
   | &HashMapNode link
   | Bucket list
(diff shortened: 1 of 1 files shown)


mvdan (approved):
Looks good! I agree that a HAMT with inline nodes makes little sense.

PR #324, opened 2020-10-20
Discussion: IPLD Data Model

vmx:
This is triggered by the "Strings" discussion. Please note that the work "String" doesn't even appear in the document (except for the appendix).

My hope that this view on the IPLD Data Model will give a new perspective on the "what are Strings" discussion.

I decided to push it as a PR, so that we can discuss it here and even have inline comments.

This isn't really meant to be merged. Hopefully the outcome of this discussion will lead to something that then can be merged.

### DIFF
--- a/design/history/exploration-reports/2020.10-data-model.md
+++ b/design/history/exploration-reports/2020.10-data-model.md
@@ -1,0 +1,91 @@
+IPLD Data Model
+===============
+
+The IPLD Data model is about data. Data is usually encoded in bytes. Programming languages use types to differentiate between different kinds of byte sequences. They help with interpreting what those bytes are about. They give a meaning. In the IPLD Data Model we therefore have so-called Kinds, which represent an abstraction layer on top of bytes, which then can be translated into corresponding programming language types.
+
+The conversion into programming language types is often straight forward, but not always. Though the goal is that IPLD Data Model Kinds are general enough, so that they can be mapped intuitively to programming language native types.
+
+The Data Model is the layer between the programming language and the serialization which is done by the IPLD Codecs. We aim for having data round-trip between the programming language and the Data Model and between the Data Model and the Codec. These are distinct steps.
+
+Going from programming language to Data Model (and back) is programming language specific. It depends on how a programming language works and which types are implemented. Going from Codec to Date Model is Codec specific. Some Codecs might implement only a subset of the Data Model Kinds or perhaps even a superset.
+
+On the Data Model layer we don't talk about memory layouts or how things should be serialized. We only talk about the properties of the Kinds. How they are implemented is then up to the library. The memory layout and type depends on the programming language, the serialization is the concern of IPLD Codecs. We try to minimize the number of Kinds to the minimum, while still having a useful system, where every Kind has its distinct properties.
+
+
+Primitive Kinds
+---------------
+
+### Null
+
+The `null` Kind has a cardinality of 1.
+
+
+### Boolean
+
+A boolean is either `true` or `false`. It has a cardinality of 2.
+
+
+### Integer
+
+A numeric value without fractions. It can be positive or negative.
+
+
+### Float
+
+A numeric value that represents a fraction.
+
+
+### Link
+
+A link is an identifier to some content-addressed data. It has the same properties as a CID.
+
+
+Compound Kinds
+--------------
+
+### Byte-Array
+
+A fixed sized sequence of 8-bit bytes. We don't define a primitive `Byte` Kind, that would be redundant as it can be represented as a single element Byte-Array.
+
+
+### Text
+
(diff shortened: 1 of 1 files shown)


mikeal:
The language I recommended for the Data Model spec from our last call was:

```
Strings SHOULD be valid UTF8. Strings MAY be contain invalid UTF8 
characters and codecs implementations should document a strategy for 
handling such strings when they are encountered in languages that cannot support 
non-UTF8 characters in the native String type.
```

For map keys it’s a little trickier, we should probably have a more detailed discussion about how to document that and where it should go. My suggestion was:

1. Data Model specifies that maps must have a single and consistent key type but it’s up to the codec and the language it’s implemented in to define what that type is.
2. Codecs should document these choices and call out any language/library differentiation we know about (this list will grow over time as more implementations are added).

prataprc:
Let us say a map is defined like `Map<K>`, where K is parametrised key-type and the value is enumeration of all supported kinds in data-model. It might end up like this:

```rust
struct Map<K> where K: Hash {
    ...
}
```

If we don't have a common denominator on how K can be represented in bytes, can we have a consistent/uniform index() operation for the same map value across different implementation ?

Btw, one more question. Other than serialization, other than hash-digest on serialized block, other than schema-matching, and other than indexing operation within list and map kinds, should IPLD worry about any other operational behaviour of _data_ ?

design/history/exploration-reports/2020.10-data-model.md
```diff
@@ -0,0 +1,91 @@
...
+--------
+
+**Q: Why is the `Text` Kind Unicode and not UTF-8?**
+
+**A:** The data model doesn't talk about memory layouts. Your programming language might encode Unicode as UTF-16 (e.g. Java, JavaScript) and that's totally fine. Your Codec then defines how to serialize things, which is then likely UTF-8, but any other encoding would also be possible
+
+
+**Q: My programming language only supports strings as map keys**
+
+**A:** Depending on your language that might not be a problem. The Data Model talks about Kinds in abstract. This means that a Kind does not necessarily be the same programming language type in all cases.
```
rvagg, line 82: The byte-keys thing seems to be the most controversial component in this PR, I assume that's the primary intent since it's the largest diversion from what we have now.

The logic used here, and the example included below, could be flipped the other way to support our current standard: maps have string keys except in the languages that have trouble with the loose-unicode view, in which case it's welcome to interpret them as bytes and use them that way as long as users are made aware of the differences.

vmx:
@prataprc Your idea is more general. What we have in mind is keeping things as simple as possible. And I expect being more specific and less generic to be easier to reason about and implement.

vmx:
> The language I recommended for the Data Model spec from our last call was:
> 
> ```
> Strings SHOULD be valid UTF8. Strings MAY be contain invalid UTF8 
> characters and codecs implementations should document a strategy for 
> handling such strings when they are encountered in languages that cannot support 
> non-UTF8 characters in the native String type.
> ```

My proposal here explicitly does *not* talk about memory layout. It also does *not* talk about what Codecs should do. And I think this is the sweat spot about it.

This means that from a pure Data Model perspective Text with invalid Unicode will be an invalid.  In a document that talks about Codecs could then talk about what to do if you encounter invalid Unicode characters.

I've two more points:

 - I also don't get, why would you create invalid Unicode `Text`? Why is that desirable? Why should it be supported and not just lead to an error?
 - If we go for a "it MUST be valid Unicode" definition, how would it be different from our definition of a CID? Or do we also plan to do special things about invalid CIDs?

design/history/exploration-reports/2020.10-data-model.md
vmx: > The byte-keys thing seems to be the most controversial component in this PR

I had hope that wouldn't be the case. I had hoped that the most controversial part is that I talk about the Data Model in an abstract way, which is independent of the programming language or the Codec. So far we often convoluted those concerns and I hope we can make clearer distinctions.

vmx:
Talking about Codecs: Do we expect that non Unicode `Text` round-trips properly in a Codec? I.e. I have a Text with non Unicode and transform it into my Data Model/programming language and I then transform it back. Does it need to just work or could it error?

If it *could* error then it aligns what I'm proposing here. Only because non-Unicode `Text` would be an invalid Data Model Kind, doesn't mean that your implementation might not be able to deal with those invalid things in a special way. This is what I'm after. I want to have a simple/happy "common case" (and yes, I really think Unicode is the common case for text) code path, that is easy to implement across most of programming languages.

vmx:
>     1. Data Model specifies that maps must have a single and consistent key type but it’s up to the codec and the language it’s implemented in to define what that type is.

This aligns with what I'm saying here. This means that from a Data Model perspective they key can be arbitrary bytes, hence is the `Byte-Array` Kind.

vmx:
I've added a few questions and answers that came up on IRC by @warpfork.

vmx:
I'm closing this one, it did serve good enough as discussion starter.

PR #331, opened 2020-11-06
doc: string honesty

mikeal:
Ok, here’s the language I think we can all agree is the state of the world as far as strings are concerned.

I hope to put to bed additional discussion of data model guarantees in any particular direction when it comes to strings.

### DIFF
--- a/data-model-layer/data-model.md
+++ b/data-model-layer/data-model.md
@@ -112,7 +112,7 @@ Applications that only serialize valid UTF8 in string values will have less comp
 issues than applications that do not.
 
 Codec implementations that can de-serialization and roundtrip
-arbitrary byte data in strings will see less bug reports from people working with data produced by
+arbitrary byte data in strings will see fewer bug reports from people working with data produced by
 applications that serialize arbitrary byte data into strings.
 
 #### Bytes kind
--- a/data-model-layer/data-model.md
+++ b/data-model-layer/data-model.md
@@ -108,7 +108,7 @@ Applications **SHOULD** only encode UTF8 data into string values and use byte va
 arbitrary bytes, but IPLD libraries may not provide these guarantees and rely on the application, or often the
 programming language itself, to do so instead.
 
-Applications that only serialize valid UTF8 in string values will have less compatibility
+Applications that only serialize valid UTF8 in string values will have fewer compatibility
 issues than applications that do not.
 
 Codec implementations that can de-serialization and roundtrip
--- a/data-model-layer/data-model.md
+++ b/data-model-layer/data-model.md
@@ -91,9 +91,29 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or 
 
 #### String kind
 
-String is a scalar kind.  Its cardinality is infinite -- strings do not have a length limit.
+Strings do not have any guarantees or requirements with regard to encoding.
 
-Strings should be UTF-8 text, in [NFC](https://www.unicode.org/reports/tr15/#Norm_Forms) canonicalization.
+All the hazards you typically find among programming languages and libraries working with binary
+serialization of strings apply to strings in the IPLD data model. IPLD's data model is conceptual,
+it takes the world as it is, and the world of strings has widely known compatibility issues.
+
+* Some languages/libraries guarantee a string encoding (typically UTF8), some do not.
+* Some languages/libraries can handle arbitrary byte data in strings, some do not.
+
+While some codec specifications will define a required encoding it should be noted that in practice
+many codec implementations leave this kind of validation and sanitizaton up to the consumer (application
+code) and it is typical to find arbitrary byte data in strings even in codecs that explicitely forbid it.
+
+Applications **SHOULD** only encode UTF8 data into string values and use byte values when they need
+arbitrary bytes, but IPLD libraries may not provide these guarantees and rely on the application, or often the
+programming language itself, to do so instead.
+
+Applications that only serialize valid UTF8 in string values will have less compatibility
+issues than applications that do not.
+
+Codec implementations that can de-serialization and roundtrip
+arbitrary byte data in strings will see less bug reports from people working with data produced by
+applications that serialize arbitrary byte data into strings.
 
 #### Bytes kind
 


data-model-layer/data-model.md
```diff
@@ -91,9 +91,29 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or
...
+
+Applications **SHOULD** only encode UTF8 data into string values and use byte values when they need
+arbitrary bytes, but IPLD libraries may not provide these guarantees and rely on the application, or often the
+programming language itself, to do so instead.
+
+Applications that only serialize valid UTF8 in string values will have less compatibility
+issues than applications that do not.
+
+Codec implementations that can de-serialization and roundtrip
+arbitrary byte data in strings will see less bug reports from people working with data produced by
```
ribasushi:
```suggestion
arbitrary byte data in strings will see fewer bug reports from people working with data produced by
```

data-model-layer/data-model.md
```diff
@@ -91,9 +91,29 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or
...
+
+While some codec specifications will define a required encoding it should be noted that in practice
+many codec implementations leave this kind of validation and sanitizaton up to the consumer (application
+code) and it is typical to find arbitrary byte data in strings even in codecs that explicitely forbid it.
+
+Applications **SHOULD** only encode UTF8 data into string values and use byte values when they need
+arbitrary bytes, but IPLD libraries may not provide these guarantees and rely on the application, or often the
+programming language itself, to do so instead.
+
+Applications that only serialize valid UTF8 in string values will have less compatibility
```
ribasushi:
```suggestion
Applications that only serialize valid UTF8 in string values will have fewer compatibility
```

ribasushi (approved):
Minor fixes, bugs my OCD

warpfork (approved):
I'm going to give this a :+1: , but also document some concerns and wishes I have about our ability to create clear specifications that will lead implementers to produce solid and interoperable libraries.

I think our specs provide the most value if they provide the most clarity possible.  I think this provides a lot less clarity, and creates less vocabulary for us to describe support levels, than the text I propose in https://gist.github.com/warpfork/8dd88ffb90bdd40eff9945cf7a67fc15#ipld-strings .

I would be much happier if we made clear statements about what we expect to be supported, and I don't think it's useful to be anything other than extremely direct above the importance of supporting non-unicode byte sequences in strings given that we already know of data in the wild that will cause errors from libraries not designed well enough to support that kind of data.  To evade this issue will increase the odds of new library authors writing libraries that fail to account for this data, and I don't think that's a good move for enriching our ecosystem.

I also think most of this text would be better in a "[motivations](https://github.com/ipld/team-mgmt/blob/0a08df1206639eed67e25fcaa8978a4e04955be3/meeting-notes/2020/2020-10-08--ipld-specs.md#more)" document.  That would clear the way for shorter, clearer statements here.

That said, all of this text is a solid explanation of the world we live in.  I'm satisfied with it.

mikeal:
@warpfork we’ll definitely want a deeper document on strings. I’d like that doc to provide better guidance for implementers that can clearly explain the tradeoffs and also list the existing implementations and how they handle each case. I’ll work on that (pulling from the doc you linked) once this is agreed upon and merged, but I agree with you on the need for something with clearer guidance for implementers.

rvagg (approved):

data-model-layer/data-model.md
```diff
@@ -91,9 +91,29 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or
...
+All the hazards you typically find among programming languages and libraries working with binary
+serialization of strings apply to strings in the IPLD data model. IPLD's data model is conceptual,
+it takes the world as it is, and the world of strings has widely known compatibility issues.
+
+* Some languages/libraries guarantee a string encoding (typically UTF8), some do not.
+* Some languages/libraries can handle arbitrary byte data in strings, some do not.
+
+While some codec specifications will define a required encoding it should be noted that in practice
+many codec implementations leave this kind of validation and sanitizaton up to the consumer (application
+code) and it is typical to find arbitrary byte data in strings even in codecs that explicitely forbid it.
```
vmx, line 105: To me this language is too weak. I'd right (the emphasis shouldn't be part of this PR, it's just to make it easier to parse if you read my comment):

While *most* [I'd say all, but I can see that that's too strong] define a required encoding, it should be noted in practice that *some* codec implementations leave this kind of validation and sanitization up to the consumer (application
code) and you *might* find arbitrary byte data in strings even in codecs that explicitly forbid it.

If you want a detailed write up, let me know. The short version is, I've looked into the all the language independent codecs I could find and *all* of them specify how strings should be like (JSON, CBOR, MessagePack, ProtocolBuffers, Cap'n'proto, CUE, Apache Thrift, Apache Avro, Flatbuffers, BARE, BSON, Borsh, Amazon Ion).

data-model-layer/data-model.md
```diff
@@ -91,9 +91,29 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or
...
+
+Applications **SHOULD** only encode UTF8 data into string values and use byte values when they need
+arbitrary bytes, but IPLD libraries may not provide these guarantees and rely on the application, or often the
+programming language itself, to do so instead.
+
+Applications that only serialize valid UTF8 in string values will have fewer compatibility
+issues than applications that do not.
```
vmx, line 112: This is again not strong enough for me. I would write:

Applications that only serialize valid UTF8 in string values will be more interoperable than applications that do not.

As a single data point (I'm sure there would be more, but I'm not sure if it would be time well spent): If you naively write a string containing arbitrary bytes e.g. `\x01\x61\x91` with the Google provided libraries, you will get different data. Go will write it as expected, but JS will write `\x01\x61\c2\c91` due to strings being UTF-16. I'm sure there are workarounds for that. I just want to make clear that "bytes in strings" sounds straight forward, easy to support and less error prone that UTF-8 string. It's certainly not the case when you think about interoperability between languages.

vmx (commented):
In other conversations I said that I'd prefer if you wouldn't put too much focus on the exact wording. But those were more like general ideas and on issues. I think for written text it is important, hence I propose a few changes. And of course my wording isn't ideal either, so I'm happy to change that if we can find better language to transport my intentins.

The intention behind those changes is to encourage people to use UTF-8 in strings so that they get more interoperable systems.

vmx:
Based on my comments and @warpfork comments I think the point of friction is what each of us considers what leads to greater interoperability. And also what we consider valid/invalid. I try to summarize, please @warpfork correct me if I got your points wrong:

 - @warpfork: Libraries that don't validate every piece of data and also allow invalid data are more versatile. They work even if the producer of the data didn't implement things fully truth to some specs. It leads to better systems as they won't error on subtle that might not even be a problem.
 - @vmx: If data is produced which isn't spec compliant, it shouldn't be expected to work properly with libraries that are implemented according to the spec. Interoperability should be guaranteed for cases where all actors comply to the specs. If you implement your library according to the spec, it is interoperable with everyone else complying to the spec. If not it's a bug. If a spec is based on existing specs it should not contradict existing constraints in order to provide maximum interoperability even on the lowest level.

prataprc:
> Applications that only serialize valid UTF8 in string values will be more interoperable than applications that do not.

Interoperablity with past and present is one thing, but interoperability with future is a different beast altogether.

> The short version is, I've looked into the all the language independent codecs I could find and all of them specify how strings should be like (JSON, CBOR, MessagePack, ProtocolBuffers, Cap'n'proto, CUE, Apache Thrift, Apache Avro, Flatbuffers, BARE, BSON, Borsh, Amazon Ion).

May be this is a solid case for not doing it _their_ way ;) when everyone make the same design choice, everyone wins or everyone looses.

Btw, I am only following the string-encoding discussion that happens in github, so pardon me if I sound too pedagogical.

----

Not the first time I am seeing string poping up as difficult data-type. Languages like Rust will simply call each string-encoding as different type altogether. But I guess IPLD is not a type system so we don't have that luxury.

Per my understanding, `string` data-type is going to be as color-ful as there are humans and human languages. And honestly I like to have it that way. We should remember that not all linguistically different population are equally empowered with computer technology in their native-tongue. But it will happen eventually, at which point each one of them could have their own "string-data-type". For now utf8 is the largest repository of linguistic analysis from programming PoV and has backward compatibility with ASCII.

Personally, one of the reason I am attracted to IPLD / IPFS is that it wants to be future proof. It is a super difficult problem.

Also I see "interoperability" being mentioned in two different context. One, that happens after a handshake between actors, and the other that happens without-a-handshake.

mvdan (approved):
LGTM. Overall, I agree that the stronger wording belongs in the extra doc to guide implementers. I think we should all agree that this is a step in the right direction, merge it, and focus on writing the extra doc.

PR #333, opened 2020-11-18
filecoin: split basic types and state from chain.md

rvagg:
@ribasushi @willscott would you mind checking that I've done this appropriately? The main thing to look at is how I've taken the huge chain.md and extracted state.md from it. So chain.md is quite small now and points to both state.md and messages.md where those two come in to play. So the "main chain" should be fully described in chain.md according to how I understand how people describe the "main chain".

There's no data structure changes in here, just splitting docs and adding some new text to link it all together.


data-structures/filecoin/basic_types.md
```diff
@@ -0,0 +1,94 @@
...
+# returns
+type CborEncodedReturn Bytes
+```
+
+## Crypto Types
+
+```ipldsch
+type Signature union {
+  SignatureSecp256k1 1
+  SignatureBLS 2
```
willscott, line 92: i don't see where we note that both of these concrete signature types are Bytes?

willscott (approved):

data-structures/filecoin/basic_types.md
rvagg: it's implied by `representation byteprefix`

... which was introduced specifically for this case in Filecoin when we started looking at Schemafying their specs, @warpfork has never really liked this fwiw https://specs.ipld.io/schemas/authoring-guide.html#byteprefix-unions-for-bytes

ribasushi:
Looks correct overall. A few nits since this is the first time I am reading this in depth:

### Changes requested
- `s/final/initial/` ?
  https://github.com/ipld/specs/blame/65d65c426e711e3800c9ae5d70192e4775af5d1b/data-structures/filecoin/chain.md#L46
- missing closing ```
  https://github.com/ipld/specs/blame/65d65c426e711e3800c9ae5d70192e4775af5d1b/data-structures/filecoin/chain.md#L136-L138

rvagg:
> `s/final/initial/`

I went with "terminal", I'm uncomfortable with "initial" because this is viewing the chain backward, you can't _start_ from the "initial", you have to work back to it (hence the original "final"). Reasonable?

ribasushi (approved):
Perhaps `terminal Genesis block` => `foundational Genesis block`, but not loving it either. Words are hard.

Good to merge as-is

rvagg:
OK, I'm scrapping that entirely and going a different direction:

```
# The type hint here, `BlockHeader`, holds true **except** for the case of the
# Genesis block, which is a different format entirely. Note there is a
# `BlockHeader` "bootstrap genesis" but its `Parents` is a single CID pointing
# to the actual Genesis block which is described by the `Genesis` type.
```

So we have this "bootstrap genesis" which is what the specs are going to refer to as the "genesis". I'm going to merge this but seek further clarification on how core folks talk about this quirk.

PR #336, opened 2020-11-20
filecoin: clarify the height=0 parent block situation

rvagg:
After some feedback from Filecoin folks.


rvagg:
updated based on feedback, using "genesis" and "height 0" for now

PR #338, opened 2020-11-23
feat: support filecoin hamt in same schema

mikeal:
This is an attempt to support the Filecoin HAMT in the same schema. We should discuss the implications of this in the IPLD call because I’m not sure how well a double union will work in practice.

### DIFF
--- a/data-structures/hashmap.md
+++ b/data-structures/hashmap.md
@@ -101,7 +101,7 @@ See [IPLD Schemas](../../schemas) for a definition of this format.
 type HashMapRoot struct {
   hashAlg Int
   bucketSize Int
-  hamt HashMapNode
+  hamt &HashMapNode
 }
 
 # Non-root node layout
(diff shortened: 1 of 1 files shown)


mikeal:
@rvagg @mvdan where did we land with this? Did this end up working for the new Go implementation of did we find some other reason we can’t make the schemas compatible?

willscott:
i believe they are compatible for the forthcoming actor hamts in spec-actors v3

mvdan:
Yes, luckily their v3 now [uses a kinded union](https://github.com/filecoin-project/go-hamt-ipld/blob/9b8f85f2e3ade8de68d7758ec7201154e39e882e/hamt.go#L109-L112) for their equivalent to our Element :)

Something like what you did here would be useful if we want to support filecoin HAMTs older than v3 with our ADL, but as far as I know that's not planned right now. So I think we should close this PR for the time being, without losing the diff in case we want to dig it up later.

Issue #342, opened 2020-12-19
Representation of Infinity and NaN in DAG-JSON


patrsc:
DAG-JSON is supposed to support the full IPLD data model. Does this mean that the IEEE 754 special values Infinity, -Infinity, and NaN should also be supported? Currently these values are mapped to `null` in the JS implementation of DAG-JSON. It is possible to represent these values in DAG-CBOR (also the JS implementation supports it). A possible way to encode them could be the following:

````
{"/": {"float": "Infinity"}}
{"/": {"float": "-Infinity"}}
{"/": {"float": "NaN"}}
````

prataprc:
This is a recurring issue with JSON. Hope JSON gets extended with something like [json5](https://json5.org/)

rvagg:
I don't know if we've properly addressed whether "float" in the data model maps directly to IEEE 754 and therefore whether it should even include infinites and not-a-number. Instinctively I would say that these are not ideal forms to be encoding anyway, but I don't know if I can justify that when IEEE 754 is so widely supported so these specials already have wide utility.

@ipld/core we should probably resolve that base question first, then whether the proposal above is a good idea. It seems like a reasonable approach to me if we accept the place of IEEE 754 specials in the data model.

_(mostly though I'd say to just avoid floats entirely in your encoded data, they're a minefield)_

patrsc:
I agree and as far as I know IEEE 754 is the most widely used floating point representation: it is heavily used e.g. in scientific computing and engineering. Also most programming languages support it, so in my opinion it should be available in the IPLD data model to make it also useful in these domains.

rvagg:
I'm trying to get broader engagement on this question from others (we might need to be patient due to the holiday period). So far this seems to me to be the question space to consider:

1. Do we allow IEEE 754 to bleed into the Data Model and therefore we support these notions of `Infinity` `-Infinity` and `NaN`
  a. we could rule them out and reject them at encode/decode time so they aren't even usable
  b. we could treat them as magic values for a specific codec + language combination that you happen to get free but don't bet on them being transferable (this is implicitly the current approach and probably needs to be explicit if we adopt this)
  c. or we could make affordances in other codecs to support them (DAG-JSON first, as per that linked issue)
2. If yes to the above, do we treat these things as part of "float", or do they become new kinds, like "null" is its own kind?
      - this might mean either saying that, by "float", we mean IEEE 754 and they roll up into the "float" kind and we bind ourselves by IEEE 754's limitations and affordances

[...]

PR #344, opened 2021-01-11
data-model: clarifications for numbers - float & int

rvagg:
Ref: https://github.com/ipld/specs/issues/342

The main thing this does is rule out IEEE 754 specials `NaN`, `Infinity` and `-Infinity` from the data model. 👍 or 👎 to resolve that issue here.


data-model-layer/data-model.md
```diff
@@ -85,9 +85,42 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or
...
+It is important to consider codec and language limitations that may be imposed
+on the serialization of integers from the IPLD Data Model. For example:
+
+* Some codecs, such as DAG-CBOR, will assume that integers must be within the
+  64-bit signed range and reject anything larger.
+* IPLD libraries, such as go-ipld-prime, limit their in-memory representation
+  of the integer kind to the signed 64-bit range.
+* JavaScript has difficulties safely handling and representing integers outside
+  of the 53-bit unsigned range and differentiating between integers and floats
+	where there is no fractional component.
```
mvdan: mixing tabs and spaces? :)

data-model-layer/data-model.md
```diff
@@ -85,9 +85,42 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or
...
 
 #### Float kind
 
+Float is a scalar kind and refers to a number with a fractional component,
+represented with a decimal point. This may also include numbers where the
+fractional component is zero (although the distinction between these numbers
+and integers is blurred in some cases—specifically the JavaScript memory model).
```
mvdan, line 107: maybe spaces around the dash? otherwise it almost reads like a single word. but I'm not an expert at technical writing :)

data-model-layer/data-model.md
```diff
@@ -85,9 +85,42 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or
...
+
+While Float is a formal kind in the IPLD Data Model, **it is recommended that
+Float values be avoided when developing systems on IPLD** (and
+content-addressable systems in general) due to the broad scope for introducing
+variability in byte representations. Alternative methods for encoding
+fractional numbers with greater precision and less variability should be
+considered where possible.
```
mvdan: do we want to give some examples here? for example, off the top of my head it's not obvious how I would store a "big int" or a "big float" with arbitrary precision in IPLD. A base-10 string? Some sort of binary encoding in bytes?

data-model-layer/data-model.md
```diff
@@ -85,9 +85,42 @@ Boolean is a scalar kind.  Its cardinality is two -- either the value 'true' or
...
+
+There is not a strict 1:1 mapping between the Float kind and any particular
```
mvdan, line 109: But we still do support a subset of IEEE 754 though, right? Otherwise it almost seems like Float is little more than a bunch of arbitrary bytes, which might contain practically anything from a spec point of view.

mvdan (approved):
SGTM, thanks for working on this. Putting on my "new to IPLD" hat, I have some minor suggestions.

vmx (approved):

data-model-layer/data-model.md
rvagg: sigh .. the perils of having JS & Go people share docs, and me being a Go+JS person using the same editor for both, will fix

data-model-layer/data-model.md
rvagg: in the code font on here it looks like an en-dash but it's an em-dash, which is longer and shouldn't have spaces around it when used in sentences like this. If you look at it in preview it comes out properly: https://github.com/ipld/specs/blob/rvagg/ieee754-specials/data-model-layer/data-model.md

data-model-layer/data-model.md
rvagg: Yes, this is a tricky point - we're talking about the data model here which is about what you interact with programmatically, so (1) that's a step away from codecs, where the rubber hits the road on the bytes->something question, and (2) this is a matter for programming languages and IPLD libraries.

I've added this to try and get closer to the point:

> However, for practical purposes, the IPLD Float kind will be implemented using IEEE 754 primitives and byte layout in most languages and codecs due to its ubiquity.

I hope that helps clarify, along with some other words I've added.

data-model-layer/data-model.md
rvagg: I've added some examples but would appreciate some sanity checking on them!

rvagg:
Please review my latest commit, I've shuffled things a bit and added more wording to clarify the relationship with IEEE754, added some more to the reasoning behind us not recommending the use of Float and added some examples of possible replacements for it.

data-model-layer/data-model.md
mvdan: gotcha, sounds good. I guess in here it's monospace.

data-model-layer/data-model.md
mvdan: Yeah, I think that wording does help.

data-model-layer/data-model.md
```diff
@@ -106,21 +106,43 @@ represented with a decimal point. This may also include numbers where the
...
+IEEE 754 is supported (`NaN` in particular introduces considerable variability
+in byte representations). These should not be supported by IPLD tooling.
+
+While Float is a formal kind in the IPLD Data Model, **it is recommended that
+Float values be avoided when developing systems on IPLD** (and
+content-addressable systems in general) due to:
+ * The broad scope for introducing variability in byte representations.
+ * The ambiguity introduced in some languages that may cause round-trip
+   discrepancies; specifically JavaScript which does not clearly disambiguate
+	 between "float" and "integer" in its memory model.
```
mvdan: it happened again :P

mvdan (approved):
LGTM!

vmx (approved):
I really like what you've added in the new commit!

PR #345, opened 2021-01-11
dag-cbor: additional strictness details

rvagg:
Some additional clarifications, depends on #344 in its current form.

* 64-bit float always (resolved late 2020)
* IEEE 754 specials not allowed
* additional clarification about indefinite length
* additional clarification about tags

### DIFF
--- a/block-layer/codecs/dag-cbor.md
+++ b/block-layer/codecs/dag-cbor.md
@@ -44,30 +44,19 @@ Therefore the DAG-CBOR codec must:
 1. Use no tags other than the CID tag (`42`). A valid DAG-CBOR encoder must not encode using any additional tags and a valid DAG-CBOR decoder must reject objects containing additional tags as invalid.
    * This includes any of the initial values of the tag registry in [section 2.4 of the CBOR specification], such as dates, bignums, bigfloats, URIs, regular expressions and other complex, or simple values whether or not they map to the [IPLD Data Model].
 2. The only usable major type 7 minor types are those for encoding Floats (`25`, `26`, `27`), True (`20`), False (`21`) and Null (`22`).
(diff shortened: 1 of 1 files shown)


block-layer/codecs/dag-cbor.md
```diff
@@ -49,25 +49,14 @@ Therefore the DAG-CBOR codec must:
...
-
- * Floating point values must be encoded as the smallest of 16-, 32-, or 64-bit floating point that accurately represents the value, even for integral values.
-
-[refmt], for Go (used via [ipld-cbor] and [ipld-prime]), uses a consistent 64-bit approach:
-
- * All floating point values must be encoded as 64-bit floating point, even for integral values.
-
-One of these approaches will be chosen and the libraries for the other language will be adjusted or replaced to harmonize.
+5. Floating point values are always encoded in 64-bit, double-precision form, regardless of whether they can be represented as half (16) or single (32) precision.
+6. IEEE 754 special values `NaN`, `Infinity` and `-Infinity` are not accepted as they do not appear in the IPLD Data Model. Therefore, tokens `0xf97c00` (`Infinity`), `0xf97e00` (`NaN`) and `0xf9fc00` (`-Infinity`) and their 32-bit and 64-bit variants, should not appear, or be accepted in DAG-CBOR binary form.
```
ribasushi: How do we reconcile `are not accepted` with the language below that goes `these implementations encode and decode them just fine"

LGTM otherwise

ribasushi (approved):

block-layer/codecs/dag-cbor.md
rvagg: have changed to "should not be accepted"

vmx (approved):
Thanks!

PR #347, opened 2021-01-13
dag-pb: update, cleanup, make consistent, mark "Final"

rvagg:


block-layer/codecs/dag-pb.md
```diff
@@ -33,10 +41,14 @@ message PBNode {
 }
 ```
 
-The objects link names are specified in the 'Name' field of the PBLink object.
-All link names in an object must either be omitted or unique within the object.
+Historical Note: The ordering (Links then Data) of the PBNode message is due to
+a bug in the initial protobuf encoder that was used in the first implementation
```
warpfork: Much as I find this detail of protobuf odd and distateful, is it accurate to call it a bug?  Or is it just (extremely) errata?

(It's good to mention this detail, of course.  Just wondering about how protobuf considers ordering details like this.)

warpfork (approved):
:ship:

block-layer/codecs/dag-pb.md
rvagg: 🤷. This bit is just lifted from a lower section that I've removed. I kind of want to this bit too but it seems like the kind of thing that ended up in here because someone said "we should note that in the spec" some distant time ago.

rvagg: came in 4b97496c which landed in #201 which included its source, #115

rvagg: I've removed this block

PR #348, opened 2021-01-13
exploration report on encryption in IPLD

warpfork:
I've written a short (sorta) informal (relatively) exploration report on recent conversations we've had about encryption in IPLD.

Like any exploration report, this is probably incomplete, it's single author, it's subject to change, it's not a totally definitive stance of the entire IPLD team yet, etc.  But... we might take this content and move it towards being something more ratified and serious, and perhaps eventually into a spec.

There's some other conversations going on about encryption right now in various venues, so I wanted to scrape this together to maybe give a home to accumulating some shared vocabulary about the more advanced concepts being discussed.  We've also had some interesting new ideas recently about how to connect encryption primitives to the rest of the IPLD universe, so those are discussed here.

Comments welcome on this, of all sorts.  Usually an exploration report is meant to _stay_ a single author thing and go on the record that way, but in this case, I'd love lots of feedback of all kinds, and will also probably edit this to include notes from some meetings over the next couple of day.

### DIFF
--- a/design/history/exploration-reports/2021.01-encryption.md
+++ b/design/history/exploration-reports/2021.01-encryption.md
@@ -1,0 +1,262 @@
+Encryption and IPLD, 2021
+=========================
+
+This is an exploration report about the role and relationship of encryption relating to IPLD,
+gathering some thoughts and recent updates in early 2021.
+It's meant to be useful as a reference piece for further discussion at this time.
+
+This document takes input from tons of people.
+It's written by warpfork in the immediate aftermath of close conversation with Mikeal,
+but also has tons of input from Carson of Textile (via the 2021.01.11 IPLD Weekly meeting),
+and also draws on other notes exchanged over time with project such as the Ceramic Network, 3Box, Peergos, Qri, and others.
+(Even if you don't see your name here, it's likely you've contributed something --
+this topic has just been a long time brewing, so attributing all inspirations completely is now hard!)
+Thank you to all these folks for their efforts.
+
+
+Overview
+--------
+
+People frequently want to implement encryption as part of decentralized systems.
+So, it's no surprise that it's also frequent that people want a story for how encryption and IPLD should interact.
+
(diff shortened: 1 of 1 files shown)


warpfork:
Prototype code for some of this already is pushed and reviewable in https://github.com/multiformats/js-multiformats/pull/59 , for anyone who likes a more literal and fully worked thing to grind over.

design/history/exploration-reports/2021.01-encryption.md
```diff
@@ -0,0 +1,224 @@
...
+
+Encryption Codecs
+-----------------
+
+The IPLD team is now ready to plan (and implement) encryption which is signaled by multicodec indicators
+(and thus works anywhere CIDs are used),
+and works in the natural way an IPLD codec is expected to work.
+
+(This is a big change in stance.
+Previously, we've considered it unclear whether codecs)
```
rvagg: whether codecs ..... "are the right place to do this"?

design/history/exploration-reports/2021.01-encryption.md
```diff
@@ -0,0 +1,224 @@
...
+
+There's a couple of details about how we expect this to work which are recent realizations,
+and so this document might be nearly the first description of them:
+
+
+### encryption codecs use multicodec indicators
```
rvagg, line 54: this section might want to be updated - the current approach is to have a single new `ipld` codec which encodes the cipher as a multicodec integer, so we have one new codec (for now) and a bunch of ciphers in the table.

warpfork:
I updated this to end with a note that discussion is ongoing in the other PR.  I'd like to freeze the rest as "exploration report".  (The filename has January in it; it's an accurate statement about where thoughts were in January.  I'm unlikely to further revisit this anytime soon unless there's a fresh impetus.)

I'll merge this tomorrow if there's no objections, just to get it out of the list of work that looks like it's in-progress.

vmx (approved):

PR #353, opened 2021-01-20
dag-cbor: cleanup, reference RFC 8949, include @ipld/dag-cbor, mark final

rvagg:


vmx (approved):

block-layer/codecs/dag-cbor.md
```diff
@@ -37,93 +39,103 @@ In DAG-CBOR, map keys must be strings, as defined by the [IPLD Data Model]. Othe
...
+[CBOR Tags Registry]: https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml
+[IPLD content identifiers (CIDs) in CBOR]: https://github.com/ipld/cid-cbor/
+[section 3.4 of RFC 8949]: https://tools.ietf.org/html/rfc8949#section-3.4
+[section 4.2 of RFC 8949]: https://tools.ietf.org/html/rfc8949#section-4.2
+[section 3.9 of RFC 7049]: https://tools.ietf.org/html/rfc7049#section-3.9
+[Simple Values]: https://tools.ietf.org/html/rfc8949#section-2.1
+[section 5.1 of RFC 8949]: https://tools.ietf.org/html/rfc8949#section-5.1
+[@ipld/dag-cbor]: https://github.com/ipld/js-dag-cbor/
+[multiformats]: https://github.com/multiformats/js-multiformats/
+[`BigInt`]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
```
mvdan, line 134: TIL these can contain special markdown characters

block-layer/codecs/dag-cbor.md
```diff
@@ -37,93 +39,103 @@ In DAG-CBOR, map keys must be strings, as defined by the [IPLD Data Model]. Othe
...
+Note that inability to clearly differentiate between integers and floats in JavaScript may cause problems with round-trips of floating point values. See the [IPLD Data Model] and the discussion on [Limitations](#limitations) below for further discussion on JavaScript numbers and recommendations regarding the use of floats.
 
 ### Go
 
-[ipld-cbor] and [ipld-prime] adhere to this specification, with the following caveats:
+**[go-ipld-cbor]** and **[go-ipld-prime]** adhere to this specification, with the following caveats:
 
- * Strictness is not yet enforced on decode, blocks encoded that don't follow the strictness rules are not rejected
- * IEEE 754 special values `NaN`, `Infinity` and `-Infinity` are accepted for decode and encode.
+ * Strictness is not enforced on decode; blocks encoded that do not follow the strictness rules are not rejected.
```
mvdan, line 85: out of curiosity, do we have an issue about this for ipld-prime?

block-layer/codecs/dag-cbor.md
```diff
@@ -37,93 +39,103 @@ In DAG-CBOR, map keys must be strings, as defined by the [IPLD Data Model]. Othe
...
+ * [IEEE 754] special values `NaN`, `Infinity` and `-Infinity` are accepted for decode and encode.
```
mvdan, line 86: do we want to change this, given the recent discussions on the topic? @warpfork

mvdan (approved):

block-layer/codecs/dag-cbor.md
ribasushi: same question, LGTM otherwise

ribasushi (approved):

block-layer/codecs/dag-cbor.md
mikeal: I think we should leave out statements about when/how strictness is applied. The spec should define the strictness and implementations that want to violate the strictness can document when and where they do so.

We should strive for an equilibrium in which strictness is assumed to be the norm and is explicitly called out when not applied.

rvagg: > strictness is assumed to be the norm and is explicitly called out when not applied

that's what this section is doing, are you suggesting I remove this?

block-layer/codecs/dag-cbor.md
```diff
@@ -37,93 +39,103 @@ In DAG-CBOR, map keys must be strings, as defined by the [IPLD Data Model]. Othe
...
+2. Use the "Deterministically Encoded CBOR" rule suggestions defined in [section 4.2 of RFC 8949] except for map key ordering, which follow the original rules as defined in [section 3.9 of RFC 7049]. Therefore, a valid DAG-CBOR encoder should produce encoded forms that adhere to the following rules, and a valid DAG-CBOR decoder should reject encoded forms not adhering to the following rules:
    * Integer encoding must be as short as possible.
    * The expression of lengths in major types 2 through 5 must be as short as possible.
    * The expression of tag numbers (specifically only `42`) must be as short as possible for major type 6. Therefore, for valid DAG-CBOR, the only tag token that can appear is `0xd82a` - where `0xd8` is "major type 6 with 8-bit integer to follow" and `0x2a` is the number `42`.
-   * The keys in every map must be sorted lowest value to highest. Sorting is performed on the bytes of the representation of the keys.
+   * The keys in every map must be sorted length-first by the byte representation of the string keys, where:
```
warpfork, line 52: I still hold that map sorting is an antifeature and it's almost certainly a mistake to specify it with a "must".

We've got lots of data in the wild that doesn't do this already, and lots of reasoning why it's ecosystemically dubious even to try to force it in the future.  I've written https://gist.github.com/warpfork/9315eac9848fa2c2ab36174272a80d9b#reasons about this and still feel it holds in its entirety.

I strongly suspect the outcomes of strict map sorting will be: people are pushed to use CBOR (and not DAG-CBOR) for handling any legacy data; and/or we'll see the emergence of a DAG-CBOR-HARDER codec in the future which is identical to this spec in every way but is order-preserving.  Or: we'll simply see almost every implementation not enforce this rule (in which case, what was the point?).

I would much rather we move this to a section on canonicalization which makes sorting *recommend* when the user data *is known to lack a meaningful ordering*.

I'm also not going to hold the PR up on this, because the previous text also stated sorting, and this text just states it better.

block-layer/codecs/dag-cbor.md
warpfork: It seems we should probably remove (or at least put behind a flag) our support for these "numbers", yes.  (I have no idea what our code currently does with these, in detail.)

block-layer/codecs/dag-cbor.md
warpfork: Probably need to break this down into a couple things:

- The NaN/Infinite issue
- Float encoding size (I think we accept variable size, but only emit 64bit size, iirc)
- Map sorting (but [I have opinions about whether this is a good idea](https://github.com/ipld/specs/pull/353/files#r561683154))

block-layer/codecs/dag-cbor.md
rvagg: I'm only using foot-links here because they're already in this doc, otherwise I _hate_ this style of MD linking

block-layer/codecs/dag-cbor.md
rvagg: Sorry, I'm getting mixed signals in this thread. For clarity - refmt supports `NaN`, `Infinity` and `-Infinity` but I'm assuming this will change at some point in the future, perhaps when ipld-prime absorbs more refmt functionality and does its own CBOR. I don't view this as particularly urgent work and don't think these listed items should be viewed as marks against the libraries, just informational for users.

@mikeal @warpfork are you suggesting I remove comments about how the various libraries conform or don't conform from here? That would seem to be a negative for this document, they're informational for users. If you have suggestions for how to reframe the wording then let me know.

block-layer/codecs/dag-cbor.md
rvagg: OK, let me first acknowledge your overall objection here @warpfork, but this particular change to the text is only clarifying what the current status is, not changing the current status. It would be nice if we could find a way to split out this objection from this particular PR rather than hold up a clean-up with a structural objection. How can I adjust to help this happen? Do you want me to remove the change to "Final"?

I have thoughts on how we might inch toward a reconciliation on the views here, but I really don't want to bikeshed in this PR.

block-layer/codecs/dag-cbor.md
mvdan: Really? I actually find it quite tidy and nice for reading the "plaintext English" without having to render the doc.

block-layer/codecs/dag-cbor.md
warpfork: Yeah, if we leave the final off for the moment being, I'd be happy overall.

block-layer/codecs/dag-cbor.md
warpfork: I am happy about how this is currently phrased.  (Github's UI may have caused me and Mikeal tunnel vision earlier; some views "helpfully" cut off the line introducing this list stating that it's a list of exceptions to the rule.  I think we're all in agreement.)

rvagg:
Turned it back to "Draft" before merging as per comment above.

PR #356, opened 2021-01-25
dag-json: cleanup and fully specify reserved space

rvagg:
I've been working ok figuring out how to codify the rules that we discussed sometime last meeting .. during some meeting. I can't find the particular meeting by looking at the minutes in https://github.com/ipld/team-mgmt so I'm doing this from memory. But the main points were:

* Strictly only allow certain forms in the `{"/":...}` namespace so that tokenizing parsers didn't have to buffer and unroll too much data but can quickly reject form without getting too deep.
* ~~Remove the multibase prefix from the encoded bytes.~~

After working through some of this, I've flipped on my opinion of the second point and would rather it stay as proper multibase, even if it's still strictly base64. It adds clarity and there's scope for using this for further validation if we want to. It's a single byte and I think it increases the clarity and reduces the number of edge cases. So I haven't included that change here, but discussion is welcome if you feel strongly!

The bit to pay most attention to below is "The Reserved Namespace" where I've tried to enumerate the rules. It turns out to be quite hard to explain it with crystal clarity, but that could be me so suggestions welcome!

Along with this, as exploration, I've been working on a version of JS DAG-JSON that does tokenized decoding and uses the same base encoder/decoder as the new JS DAG-CBOR codec, just by swapping out the backend. There's some consistency benefits here in having the same code paths and same affordances for applying data model strictness. Still uncertain whether I'll actually propose that we merge this, although I did uncover some bugs in the current implementation in the process, but the code is here: https://github.com/ipld/js-dag-json/compare/rvagg/cborg (see the innards of `DagJsonTokenizer` for the token parsing rules that arise from the reserved space rules).

### DIFF
--- a/block-layer/codecs/dag-json.md
+++ b/block-layer/codecs/dag-json.md
@@ -44,11 +44,13 @@ See further discussion on [Floats in the Data Model](../../data-model.md#float-k
 
 ### Bytes
 
-The Bytes kind is represented as an object with `"bytes"` as key and a [Multibase](https://github.com/multiformats/multibase) Base64 encoded string as value. The Base64 encoding is the one described in [RFC 4648, section 4](https://tools.ietf.org/html/rfc4648#section-4) without padding, hence the Multibase prefix is `m`.
+The Bytes kind is represented as an object with `"bytes"` as key and a Base64 encoded string as value. The Base64 encoding is the one described in [RFC 4648, section 4](https://tools.ietf.org/html/rfc4648#section-4) without padding.
+
+_Note that a previous version of this specification and some implementations used a [Multibase](https://github.com/multiformats/multibase) prefix `m` for the bytes, this has been removed from the specification and the Base64 encoded bytes **should not** be prefixed._
 
 
 ```javascript
-{"/": { "bytes": String /* Multibase Base64 encoded binary */ }}
+{"/": { "bytes": String /* Base64 encoded binary */ }}
 ```
 
 ### Links
@@ -69,7 +71,7 @@ Maps with the first key of `"/"` are considered the **reserved namespace** in DA
 The two forms used in the reserved namespace are:
 
  * **CID**: a map with the single key `"/"`, whose value is a string, must contain a valid CIDv0 in Base58 string form **or** CIDv1 in Base32 string form. Such a map whose string does not properly represent such a CID should should be rejected as invalid DAG-JSON.
- * **Bytes**: A map with the single key `"/"`, whose value is a map with the single key `"bytes"`, whose value is a string, must contain a valid multibase Base64 encoded byte array. Such a construction whose string does not properly represent a multibase base64 encoded byte array should be rejected as invalid DAG-JSON.
+ * **Bytes**: A map with the single key `"/"`, whose value is a map with the single key `"bytes"`, whose value is a string, must contain a valid Base64 encoded byte array. Such a construction whose string does not properly represent a Base64 encoded byte array should be rejected as invalid DAG-JSON.
 
 #### Parse rejection modes in the reserved namespace
 
@@ -106,7 +108,8 @@ There is no mechanism for escaping otherwise valid JSON data that takes these fo
 
 The legacy **[ipld-dag-json](https://github.com/ipld/js-ipld-dag-json)** implementation adheres to this specification, with the following caveats:
  * The reserved namespace rules above are not strictly applied. Decoding maps with the forms of Bytes and Links but with additional entries in inner or outer maps will be successfuly decoded as Bytes or Links but the extraneous entries will be ignored.
+ * Bytes are encoded with their Multibase Base64 prefix `m` as per a previous version of this specification.
 
 ### Go
 
-**[go-ipld-prime]** adheres to this specification.
+**[go-ipld-prime]** adheres to this specification except for the sorting of map keys, which retain their assembled order.
--- a/block-layer/codecs/dag-json.md
+++ b/block-layer/codecs/dag-json.md
@@ -112,4 +112,6 @@ The legacy **[ipld-dag-json](https://github.com/ipld/js-ipld-dag-json)** impleme
 
 ### Go
 
-**[go-ipld-prime]** adheres to this specification except for the sorting of map keys, which retain their assembled order.
+**[go-ipld-prime]** adheres to this specification with the following caveats:
+ * Map keys are not sorted, they retain their assembled order.
+ * Encoded forms are pretty-printed, i.e. do not have whitespace stripped.
--- a/block-layer/codecs/dag-json.md
+++ b/block-layer/codecs/dag-json.md
@@ -4,9 +4,17 @@
 
 DAG-JSON supports the full [IPLD Data Model](../../data-model-layer/data-model.md).
 
+DAG-JSON uses the [JavaScript Object Notation (JSON)] data format, defined by [RFC 8259](https://tools.ietf.org/html/rfc8259).
+
 ## Format
 
-### Serialization
+The native JSON IPLD format is called DAG-JSON to disambiguate it from regular JSON. Most simple JSON objects are valid DAG-JSON. The primary differences are:
+
+ * Bytes and Links are supported with special use of single-key (`"/"`) map.
+ * In limited cases, maps with the key `"/"` other than those used to encode Bytes and Links, are disallowed.
+ * Maps are sorted by key.
+
+## Serialization
 
 Codec implementors **MUST** do the following in order to ensure hashes consistently match for the same block data.
 
@@ -16,24 +24,25 @@ Codec implementors **MUST** do the following in order to ensure hashes consisten
 This produces the most compact and consistent representation which will ensure that two codecs
 producing the same data end up with matching block hashes.
 
-### Natively supported kinds
+## Supported Kinds
 
-All kinds of the IPLD Data Model except Bytes and Link are supported natively by JSON.
+All [IPLD Data Model Kinds](../../data-model.md#kinds) except Bytes and Link are supported natively by JSON.
 
-#### Numbers
+Bytes and Links use extensions specific to DAG-JSON. They are implemented as an map, where the single key is a slash (`"/"`) and the value contains the kind's data.
 
-Numbers are a special case. JSON only has a single number type, though many dynamically typed programming languages (e.g. Python, Ruby, PHP) distinguish between integers and floats when parsing JSON. A number consisting of an optional leading sign and only digits is parsed as integer, if it contains a decimal point, it's parsed as a float. In DAG-JSON the same method is used to represent integers and floats.
+### Numbers
 
-Contrary to popular belief, JSON as a format supports Big Integers. It's only
-JavaScript itself that has trouble with them. This means JS implementations
-of DAG-JSON can't use the native JSON parser and serializer if integers bigger
-than `2^53 - 1` should be supported.
+JSON only has a single number type. Many dynamically typed programming languages (e.g. Python, Ruby, PHP) distinguish between integers and floats when parsing JSON. JavaScript does not since all numbers are represented internally as IEEE 754 floats. A JSON number consisting of an optional leading sign (`-`) and only digits is parsed as integer, if it contains a decimal point, it's parsed as a float. DAG-JSON the same method is used to represent integers and floats.
 
-### Other kinds
+Data Model floats that do not have a fractional component should be encoded without a decimal point, and will therefore not be distinguishable from an integer during round-trip.
 
-The non-natively supported kinds are wrapped in an object, where the key is a slash (`"/"`) and the value is the actual kind.
+Contrary to popular belief, JSON as a format supports Big Integers. It's only JavaScript itself that has trouble with them. This means JS implementations of DAG-JSON can't use the native JSON parser and serializer if integers bigger than `2^53 - 1` need to be supported.
 
-#### Bytes kind
+`Infinity`, `NaN` and `-Infinity` are not natively supported by JSON and are not supported by the IPLD Data Model.
+
+See further discussion on [Floats in the Data Model](../../data-model.md#float-kind), including a recommendation to avoid floats where possible when producing and consuming content addressed data.
+
+### Bytes
 
 The Bytes kind is represented as an object with `"bytes"` as key and a [Multibase](https://github.com/multiformats/multibase) Base64 encoded string as value. The Base64 encoding is the one described in [RFC 4648, section 4](https://tools.ietf.org/html/rfc4648#section-4) without padding, hence the Multibase prefix is `m`.
 
@@ -42,9 +51,9 @@ The Bytes kind is represented as an object with `"bytes"` as key and a [Multibas
 {"/": { "bytes": String /* Multibase Base64 encoded binary */ }}
 ```
 
-#### Link kind
+### Links
 
-A link is represented as a base encoded CID. CIDv0 and CIDv1 are encoded differently.
+A Link kind is represented as a base encoded CID. CIDv0 and CIDv1 are encoded differently.
 
  - CIDv1 is represented as a Multibase Base32 encoded string. The Base32 encoding is the one described in [RFC 4648, section 6](https://tools.ietf.org/html/rfc4648#section-6) without padding, hence the Multibase prefix is `b`.
  - CIDv0 is represented in its only possible Base58 encoding. The Base58 encoding is the one described in [Base58 draft](https://tools.ietf.org/html/draft-msporny-base58).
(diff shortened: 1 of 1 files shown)


block-layer/codecs/dag-json.md
```diff
@@ -42,13 +51,49 @@ The Bytes kind is represented as an object with `"bytes"` as key and a [Multibas
...
+ * Maps where the first key is `"/"` and its value is a map with more than one key where the first key of the inner map is `"bytes"` whose value is a string. e.g. `{"/":{"bytes":"foo","bar":"baz"}}`.
+   * Where a key exists in the inner map that sorts before `"bytes"`, the map is valid, e.g. `{"/":{"abar":"baz","bytes":"foo"}}`.
+   * Where the value of the inner map's `"bytes"` entry is not a string, the map is valid, e.g. `{"/":{"bytes":true},"bar":"baz"}`.
+ * Maps with more than one key, where the first key is `"/"` and its value is a map where the first key of the inner map is `"bytes"` whose value is a string. e.g. `{"/":{"bytes":"foo"},"bar":"baz"}`.
+   - Where a key exists that sorts before `"/"`, the map is valid, e.g. `{"0bar":"baz","/":{"bytes":"foo"}}`.
+   - Where the value of the `"bytes"` entry in the inner map is not a string, the map is valid, e.g. `{"/":{"bytes":true},"bar":"baz"}`
+
+Further:
+
+ * A map with the single key `"/"`, whose value is a string, must contain a valid CIDv0 base58 string or CIDv1 base32 string. Such a map whose string does not properly represent such a CID should should be rejected as invalid DAG-CBOR.
```
rvagg: One possible change is that we could shift the "valid CID" rule up to the above section, thereby making it so that if the string in `{"/":String}` _is not_ a valid CID, then a decoder should just interpret it as a standard map.

I think I'm a soft -1 on this because it's possible that you might get a CID that you can't decode (e.g. CIDv2 when you only have a CIDv1 library?) and an error might be better than silently passing it through.

block-layer/codecs/dag-json.md
```diff
@@ -42,13 +51,49 @@ The Bytes kind is represented as an object with `"bytes"` as key and a [Multibas
...
+ * A map with the single key `"/"`, whose value is a map with the single key `"bytes"`, whose value is a string, must contain a valid multibase base64 encoded byte array. Such a construction whose string does not properly represent a multibase base64 encoded byte array should be rejected as invalid DAG-CBOR.
```
rvagg: Similar to above, we could use the "must be base64" to allow through values that are not base64 as standard maps rather than rejecting. But again I'm a soft -1, just to remove the dark corners of the spec where weird things can happen.

I'd be more in favour of finding an escaping mechanism for data model forms that break our reserved space rules if we care enough about the forms we can't represent.

block-layer/codecs/dag-json.md
```diff
@@ -42,13 +51,49 @@ The Bytes kind is represented as an object with `"bytes"` as key and a [Multibas
...
+
+There is no mechanism for escaping otherwise valid JSON data that takes these forms. For this reason, it is recommended that the `"/"` key should be avoided in Data Model maps where DAG-JSON may be used in order to avoid such conflicts.
+
+## Implementations
```
rvagg, line 103: This section isn't quite correct but we could get it all cleaned up before merging so that it is correct. `@ipld/dag-json` currently has the same caveats as `ipld-dag-json` as well as a couple of bugs. go-ipld-prime doesn't do bytes at all yet I think, maybe not even links, last I checked. It got stalled on the reserved space discussions. There was a branch with the beginnings of the work necessary but I can't see that anymore.

block-layer/codecs/dag-json.md
```diff
@@ -16,24 +24,25 @@ Codec implementors **MUST** do the following in order to ensure hashes consisten
...
 
-All kinds of the IPLD Data Model except Bytes and Link are supported natively by JSON.
+### Numbers
 
-#### Numbers
+JSON only has a single number type. Many dynamically typed programming languages (e.g. Python, Ruby, PHP) distinguish between integers and floats when parsing JSON. JavaScript does not since all numbers are represented internally as IEEE 754 floats. A JSON number consisting of an optional leading sign (`-`) and only digits is parsed as integer, if it contains a decimal point, it's parsed as a float. DAG-JSON the same method is used to represent integers and floats.
 
-Numbers are a special case. JSON only has a single number type, though many dynamically typed programming languages (e.g. Python, Ruby, PHP) distinguish between integers and floats when parsing JSON. A number consisting of an optional leading sign and only digits is parsed as integer, if it contains a decimal point, it's parsed as a float. In DAG-JSON the same method is used to represent integers and floats.
+Data Model floats that do not have a fractional component should be encoded without a decimal point, and will therefore not be distinguishable from an integer during round-trip.
```
vmx: I would rather say: "Data Model floats that do not have a  fractional component should be encoded **with** a decimal point so that they are distinguishable from integers during round-trip"

block-layer/codecs/dag-json.md
```diff
@@ -42,13 +51,49 @@ The Bytes kind is represented as an object with `"bytes"` as key and a [Multibas
...
+
+### The Reserved Namespace
+
+Maps with the first key of `"/"` are considered the **reserved namespace** in DAG-JSON as they are used to represent Bytes and Links. There are special rules that restrict certain data forms from being properly encoded in DAG-JSON. These rules allow for the clean representation of Bytes and Links as well as efficient operation of tokenizing decoders. A tokenizing decoder should not need to buffer and back-track more than 4 tokens upon detection of a map that is not properly encoding Links or Bytes.
+
+Specifically, data with the following forms are strictly not valid DAG-JSON and should be rejected by encoders and decoders:
+
+ * Maps with more than one key, where the first key is `"/"` and its value is a string. e.g. `{"/":"foo","bar":"baz"}`.
+   - Where a key exists that sorts before `"/"`, the map is valid, e.g. `{"0bar":"baz","/":"foo"}`.
+   - Where the value of the `"/"` entry is not a string, the map is valid, e.g. `{"/":true,"bar":"baz"}`
```
vmx: It took me awhile to grasp that the main item is the rule to reject something and that the sub-items are rules that would not be rejected.

Perhaps it could be expended and we give the rules even names. Example:

### Multiple keys

Rejection criteria: Maps with more than one key, where the first key is `"/"` and its value is a string. 

Examples for invald DAG-JSON:
 - `{"/":"foo","bar":"baz"}`.
 - `{"/":"foo","bar":true}`.

Examples for valid DAG-JSON:
 - `{"0bar":"baz","/":"foo"}`: There is a key that sorts before `"/"`. The value following the `"/"` will *not* be interpreted as link.
 - `{"/":true,"bar":"baz"}`: There value after the `"/"` is not a string.
 - `{"/":"foo"}: There is only a the single map key `"/"`. The value is expected to be a valid base encoded CID.

block-layer/codecs/dag-json.md
```diff
@@ -42,13 +51,49 @@ The Bytes kind is represented as an object with `"bytes"` as key and a [Multibas
...
+ * Maps where the first key is `"/"` and its value is a map with more than one key where the first key of the inner map is `"bytes"` whose value is a string. e.g. `{"/":{"bytes":"foo","bar":"baz"}}`.
+   * Where a key exists in the inner map that sorts before `"bytes"`, the map is valid, e.g. `{"/":{"abar":"baz","bytes":"foo"}}`.
+   * Where the value of the inner map's `"bytes"` entry is not a string, the map is valid, e.g. `{"/":{"bytes":true},"bar":"baz"}`.
+ * Maps with more than one key, where the first key is `"/"` and its value is a map where the first key of the inner map is `"bytes"` whose value is a string. e.g. `{"/":{"bytes":"foo"},"bar":"baz"}`.
+   - Where a key exists that sorts before `"/"`, the map is valid, e.g. `{"0bar":"baz","/":{"bytes":"foo"}}`.
+   - Where the value of the `"bytes"` entry in the inner map is not a string, the map is valid, e.g. `{"/":{"bytes":true},"bar":"baz"}`
+
+Further:
```
vmx: What about putting the "Further" section higher up. First explain the "simple" rules on what you expect to be valid and then getting into the nitty-gritty details on how you could construct data that would still be valid but isn't really a good idea.

warpfork:
I went digging for our discussion notes (and merged some stale PRs) and found them here:

https://github.com/ipld/team-mgmt/blob/master/meeting-notes/2020/2020-11-05--ipld-specs.md#dag-json-bytes-and-links-discuss-parsing-rules-edge-cases

block-layer/codecs/dag-json.md
```diff
@@ -16,24 +24,25 @@ Codec implementors **MUST** do the following in order to ensure hashes consisten
...
-JavaScript itself that has trouble with them. This means JS implementations
-of DAG-JSON can't use the native JSON parser and serializer if integers bigger
-than `2^53 - 1` should be supported.
+Contrary to popular belief, JSON as a format supports Big Integers. It's only JavaScript itself that has trouble with them. This means JS implementations of DAG-JSON can't use the native JSON parser and serializer if integers bigger than `2^53 - 1` need to be supported.
 
-### Other kinds
+`Infinity`, `NaN` and `-Infinity` are not natively supported by JSON and are not supported by the IPLD Data Model.
 
-The non-natively supported kinds are wrapped in an object, where the key is a slash (`"/"`) and the value is the actual kind.
+See further discussion on [Floats in the Data Model](../../data-model.md#float-kind), including a recommendation to avoid floats where possible when producing and consuming content addressed data.
```
warpfork, line 43: _this is more of a comment than a question..._

I really like this phrasing.  Very well said about the recommendation and its key reason.  Nice.

block-layer/codecs/dag-json.md
rvagg: just revisiting this from my backlog -- @vmx @warpfork you both seem to be suggesting we encode like `{"foo":10.0}`. Do you think there might be a risk in some JSON decoder _somewhere_ not being able to parse such a construct? I think that a design-goal of dag-json is still that it should be a subset of JSON and therefore be parseable by any generic parser. Would inserting additional `.0` create a risk that we're straying too far from "standard"?

warpfork: Yeah, I don't think this is a stretch.  As far as I know, a trailing `.0` should be accepted in just about any generic JSON parser.  One would really have to go out of their way to reject that, it seems to me, and I doubt anyone has.

block-layer/codecs/dag-json.md
rvagg: I think I've addressed this, but worth a review

rvagg:
I've addressed the reviews so far and I think this is ready for landing. The reserved namespace rules are hard to be clear about but I've done the shuffle recommended by @vmx and separated things out a little more to hopefully make it easier to digest.

I went with the `.0` rule for floats with no fractional component - although this is no help to JavaScript but Go can implement it.

I also have a working JavaScript version that I'm pretty much ready to release. It uses the same backend as the DAG-CBOR codec but with JSON semantics. We've lost some speed in the process but we get all the reserved namespace rules and all the strictness.

@warpfork care to review the changes?

block-layer/codecs/dag-json.md
```diff
@@ -16,24 +24,25 @@ Codec implementors **MUST** do the following in order to ensure hashes consisten
...
+JSON only has a single number type. Many dynamically typed programming languages (e.g. Python, Ruby, PHP) distinguish between integers and floats when parsing JSON. JavaScript does not since all numbers are represented internally as IEEE 754 floats. A JSON number consisting of an optional leading sign (`-`) and only digits is parsed as integer, if it contains a decimal point, it's parsed as a float. DAG-JSON the same method is used to represent integers and floats.
 
-Contrary to popular belief, JSON as a format supports Big Integers. It's only
-JavaScript itself that has trouble with them. This means JS implementations
-of DAG-JSON can't use the native JSON parser and serializer if integers bigger
-than `2^53 - 1` should be supported.
+Data Model floats that do not have a fractional component should be encoded **with** a decimal point, and will therefore be distinguishable from an integer during round-trip. (Note that since JavaScript still cannot distinguish a float from an integer where the number has no fractional component, this rule will not impact JavaScript encoding or decoding).
```
vmx, line 37: I'm not sure if the note about JavaScript is needed as it does exactly that when encoding/decoding JSON:

```console
> JSON.stringify(2.1)
'2.1'
> JSON.stringify(2.0)
'2'
> JSON.parse('2.1')
2.1
> JSON.parse('2.0')
2
```

vmx (approved):
The rewrite of the "Parse rejection modes" reads great now!

block-layer/codecs/dag-json.md
rvagg: yeah, you're making my point @vmx. The spec is saying that `2.0` should be encoded as `2.0` but JavaScript can only do `2` since we can't distinguish. Then in the reverse the spec says that a `2.0` should be decoded as a "float" but JavaScript still can't distinguish so a `2` and a `2.0` would come out as the same thing, regardless. Maybe one day we'll retain the information that "this was a float in the encoded form" but there's no infrastructure to attach that information today.

So the note basically says "JavaScript can't conform to this part of the spec". The spec says "... floats that **do not** have a fractional component should be encoded **with** a decimal point".

rvagg:
As per a brief discussion today:

* go-ipld-prime just had `"bytes"` implemented @ https://github.com/ipld/go-ipld-prime/pull/166 but this was done without the multibase prefix as per the current version of the spec
* We have previously discussed removal of the multibase prefix but this spec change didn't include that -- see the original post for a discussion on that and why it was removed from scope.
* The current Rust implementation has a multibase prefix and Forest has some test fixtures using dag-json, and the current JavaScript dag-json has a multibase prefix too.

So either we should fix go-ipld-prime to match spec or we just bite the bullet and remove it and get it done everywhere—I can do this in JavaScript pretty quickly after merging this.

@warpfork @vmx @willscott (and anyone else) 👍 for _just remove multibase prefix_ and 👎 for _leave the spec alone darn it!_.

vmx:
My reasons for having the multibase prefix:

 - consistency: the CIDs are also Multibases. The rule would be: all base encoded things are Multibases (how could would it be if that would be true across all our specs?)
 - we do self-describing things: to me having a multibase isn't necessarily a signal that any arbitrary Multibase is allowed, it's for being self-described.
 - not performance: if you care about performance, don't use DAG-JSON. An even if you care about allocations, for base encodings you need to over-allocate slightly anyway, so just allocate one byte more for the prefix
 - why change now: generally we are hesitate to change things in specs that are already out there, unless we really have to. We don't know who else has implemented that spec already. So why changing the status quo if it doesn't really make a practical difference (except for one of our own libraries we control that didn't follow the spec).

All that said, I don't have a strong opinion and can live with either way.

rvagg:
* https://github.com/ipld/go-ipld-prime/pull/168 has a multibase prefix
* https://github.com/ipld/js-dag-json/pull/36 has the same test case ipld-prime has

at first I couldn't get them making the same bytes and thought there must be a multibase incompatibility (the horror!) but it's `[]byte('deadbeef')` being deceptive cause that's actually a string, not a hex value in spite of the nice hex characters..

also, sorting, ipld-prime don't sort and that engages hostilities in @warpfork and is not fun for compatibility fixtures :facepalm: and I kind of don't want to have that argument over and over so will skip it for now.

But basically, it's not a big deal to do it in go-ipld-prime. It's also not a big deal to strip it in `@ipld/dag-json`, but I'm also a soft -1 to changing the spec and leaving the prefix in there for reasons stated in this thread and restated by @vmx. But, it's not a big enough deal if someone else has particularly strong feelings and justification.

warpfork:
Arguments against inserting multibase in this bytes field, just to have them written:

- it will become "another one of those bytes" that anyone wanting compactness will question and question and question again -- just like the extra "zero" byte in CIDs in DAG-CBOR which confuses, surprises, and annoys *every* single person who ever has cause to notice it.
  - I've stopped saving references to every time this has happened about that zero in DAG-CBOR, but it's... many times.  People (me included) repeatedly implement DAG-CBOR without it, and then go "what?!" when a fixture points out that it's needed; or people think it's a bug, and try to remove it again later on in development cycles!  And this regularly comes from people working "down the hall" from us, not just distant strangers.  Even if we nevermind the excess bytes on the wire, the number of human hours this has wasted over time does not sit well with me.  I'd like to not do this again.
- having a multibase indicator would suggest it's acceptable to use other multibase indicators here... which it is not; in turn, it is not desirable to all multiple multibases to be used here because it would provide negative value if DAG-JSON were to be *more* expressive than the IPLD Data Model in this position.
- **holistically**: I think multibases are to be used when you've got a string and it's fairly user-facing, and there are no other disambiguators or indicators in the area that could help you know what that value is.  **This is not the scenario we have in the middle of dag-json**: we already have 1) the multicodec indicator, from the outside, 2) all the information/confidence provided by the success of parsing as dag-json throughout the block, 3) the information/confidence provided by a single-key map with a "/" key, and 4) the information/confidence provided by a single-key map with a "bytes" key, and then 5) the confidence provided by the success of parsing the value therein as base64.  This is a *huge* amount of confidence and a *ton* of non-accidental structure, and it seems to me heaping on another byte just to "be sure" it's clear what base is expected here would be wild overkill.
  - The (admittedly rather fine) distinction I'm trying to make here is that the "natural form" of this data is bytes -- it's not a string of some multibase, it's bytes -- and dag-json is in the business of specifying how those bytes are encoded in the serial format.  I think that we have a goal of using indicators like multibase in contexts where there are not yet indicators; but in the the *middle* of a codec's serial forms is *not* such a context.
    - Can you imagine if we put a "zero" prefix on every bytes scalar in DAG-CBOR, too, just to make it _really_ clear that they're bytes?
- **We argued about this before** in https://github.com/ipld/team-mgmt/blob/master/meeting-notes/2020/2020-11-05--ipld-specs.md#votes.
  - I distinctly remember *trying* to hold a vote and do an actual formal count -- and now do regret not **forcing** that! -- but as the notes say, and we reviewed and merged, in video, we had entirely assenting voices for removing multibase.  There were more than a half dozen people there, fully engaged, and having just collaboratively reviewed the entire subject.
  - That meeting had *more* people and more of their attention dedicated to it than this re-review today has got, so I'd prefer to stick with it.

rvagg:
fine, you appear to have a stronger opinion than either of us (although I'm not sure I'm convinced it's all that strong)

have pushed a new version that removes it and will do likewise in the JS impl

rvagg:
(PTAL & merge if :thumbsup:)

PR #357, opened 2021-02-22
exploration report about fixtures

warpfork:
Nitty gritty not-fun-but-it-matters engineering stuff: how should we go about test fixtures that we want to be cross-language?

We've got some smatterings of this stuff, but we need a lot more, so I'd like to think about which direction we want to go.

### DIFF
--- a/design/history/exploration-reports/2021.02-cross-language-fixtures.md
+++ b/design/history/exploration-reports/2021.02-cross-language-fixtures.md
@@ -1,0 +1,160 @@
+let's talk about how we do fixtures
+===================================
+
+This is an exploration report to talk about ways we could do test fixtures that is cross-language and generally "feels nice".
+
+We have some work in this area already, but we need a lot more of it.
+Maybe some of it can adopt some standard conventions to save work and increase readability (and writeability).
+
+
+Requirements
+------------
+
+- separate document segments should be clearly delinated, and allow separate formats.
+	- (e.g., no, one homogeneus JSON document full of more JSON documents is not actually ideal.)
+- separate document segments should be able to be combined into one file and edited together.
+	- it just saves a lot of time and hastle in practice.
+	- avoids the question of "which of these files are related?" without spaming tons of directories (which would otherwise be common in practice).
+	- if document segments have to be split into wholy separate files, one can't copy-paste snippets of fixture easily.  (Means: bug reports and discussions in issues may be likely to invent an ad-hoc syntax for document segments, resulting in more work to turn those things into fixtures.)
+- easy to edit as a human.  Introducing new segments should ideally be context-free.
+	- writing these fixtures needs to be as low-friction as practically possible.
+	- e.g.: "put cursor at end of file, press enter, press dash three times" -- something like this complexity level is the goal.
+- should preserve whitespace precisely.
+	- it would be problematic for parser fixtures if whitespace isn't preserved!  (Mind touchy things like trailing linebreaks, especially.)
+- must be able to frame incomplete sub-documents, and not require knowledge of the parser for them.
+	- I need to use this to write fixtures for parsers _and their error modes_, which means sub-documents might be _intentionally_ unparsable.
+- nonrequirement: able to represent binary.
+	- as close as possible is certainly better.
+	- this can't be a goal simply because it runs directly crosswise to "easy to edit as a human".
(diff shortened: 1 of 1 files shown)


design/history/exploration-reports/2021.02-cross-language-fixtures.md
```diff
@@ -0,0 +1,141 @@
...
+---------
+
+- selector fixtures in json: https://github.com/ChainSafe/ipld-traversal-vectors
+	- notes on outcome: awesome... except when people eyeball it, the fact that the top level maps are containing "control plane" info (which part of the fixture it is) is non-obvious.
+- schema DSL parser fixtures: https://github.com/ipld/js-ipld-schema/tree/master/test/fixtures and https://github.com/ipld/go-ipld-schema/tree/master/test/fixtures
+	- notes on outcome: the parts that use yaml: awesome... although a lot of the yaml files are resorting to heredocs.
+		- heredocs aren't always preserved by yaml parse and reserialize, so that's a knock against the "programatically updatable" goal.
+		- This author is frankly just not a fan of yaml.  The parsers are... complex.
+	- notes on outcome: the parts that don't use yaml: awesome... except sometimes there's two files that are siblings, and they're related, and you just have to "know" that.
+- (other project) "interactive" fixtures in noise: https://github.com/pipedown/noise/blob/master/repl-tests/limit.noise
```
vmx, line 52: The prior art to that one is https://dev.mysql.com/doc/dev/mysql-server/latest/PAGE_MYSQL_TEST_RUN.html

design/history/exploration-reports/2021.02-cross-language-fixtures.md
```diff
@@ -0,0 +1,160 @@
...
+	- writing these fixtures needs to be as low-friction as practically possible.
+	- e.g.: "put cursor at end of file, press enter, press dash three times" -- something like this complexity level is the goal.
+- should preserve whitespace precisely.
+	- it would be problematic for parser fixtures if whitespace isn't preserved!  (Mind touchy things like trailing linebreaks, especially.)
+- must be able to frame incomplete sub-documents, and not require knowledge of the parser for them.
+	- I need to use this to write fixtures for parsers _and their error modes_, which means sub-documents might be _intentionally_ unparsable.
+- nonrequirement: able to represent binary.
+	- as close as possible is certainly better.
+	- this can't be a goal simply because it runs directly crosswise to "easy to edit as a human".
+	- we may want to make a convention around base64 (?) and possible around comments in that format (?) ... but this can be implemented separately from the overall document container format.
```
aschmahmann, line 29: multibase might be nice here, could contain some flag that indicates the data is multibase encoded.

warpfork: I think the key on this point is that it's almost something we're going to solve be nesting it inside the container format.

E.g. precisely zero things we can pick up off the shelf are going to support multibase.

But to pick up something off the shelf, like txtar just to pick a random example, and then put either b64 or multibase inside it, is pretty trivial.

design/history/exploration-reports/2021.02-cross-language-fixtures.md
```diff
@@ -0,0 +1,160 @@
...
+-------
+
+### txtar
+
+Txtar is a simple textual format, with reference implementation at https://pkg.go.dev/golang.org/x/tools/txtar .
+
+- Pros:
+	- fits the requirements.
+	- needs verification: clarity of trailing linebreaks.
+- Cons:
```
aschmahmann, line 73: It's probably fine if you're careful since it's obvious, but from what I can tell your fixture cannot contain the characters `-- test --` since it will be treated as if you're starting a new fixture and there are no escape sequences defined in the format.

design/history/exploration-reports/2021.02-cross-language-fixtures.md
warpfork: Added entry.  :+1:

design/history/exploration-reports/2021.02-cross-language-fixtures.md
warpfork: Glad you pointed this out.  I was so enthusiastic about txtar on my first glance at it that I somehow forgot to check this.

Updated document with notes about band size for all the options.

warpfork:
Added a few more critical notes to this pointed out by comments (thank you!), and will now merge -- this is just an exploration report; no binding decisions have been made here; and it seems the exploration is done or at least paused for now so it's time to merge :)

PR #359, opened 2021-03-26
data-structures: update filecoin's HAMT version

mvdan:
(see commit message)

### DIFF
--- a/data-structures/hashmap.md
+++ b/data-structures/hashmap.md
@@ -47,7 +47,7 @@ The IPLD HashMap is constructed as a [hash array mapped trie (HAMT)](https://en.
 * Peergos [CHAMP](https://github.com/Peergos/Peergos/blob/master/src/peergos/shared/hamt/Champ.java) implementation
 * [IAMap](https://github.com/rvagg/iamap) JavaScript implementation of the algorithm
 * [ipld-hashmap](https://github.com/rvagg/js-ipld-hashmap) JavaScript IPLD frontend to IAMap with a mutable API
-* [go-hamt-ipld](https://github.com/ipfs/go-hamt-ipld) Filecoin Go HAMT implementation used by the [Lotus](https://lotu.sh/) client. See the appendix for how this implementation differs from this specification.
+* [go-hamt-ipld](https://github.com/filecoin-project/go-hamt-ipld) Filecoin Go HAMT implementation used by the [Lotus](https://lotu.sh/) client. See the appendix for how this implementation differs from this specification.
 
 ## Summary
 
@@ -285,7 +285,7 @@ As yet, there is no known hash collision attack vector against IPLD data structu
 
 The [Filecoin Project blockchain](https://filecoin-project.github.io/specs/) makes use of a HAMT that uses the same HAMT with buckets and CHAMP mutation semantics as outlined in this document. It encodes directly as [DAG-CBOR](../block-layer/codecs/dag-cbor.md) but uses a different block layout to the one specified here. This section documents the specific ways that the Filecoin HAMT variant differs from this specification. IPLD HashMap implementations may be able to implement a form that provides compatibility with Filecoin when requested by the user.
 
-The reference Go implementation for the Filecoin HAMT is used by the [Lotus](https://lotu.sh/) client and is available at <https://github.com/ipfs/go-hamt-ipld>.
+The reference Go implementation for the Filecoin HAMT is used by the [Lotus](https://lotu.sh/) client and is available at <https://github.com/filecoin-project/go-hamt-ipld>. Note that this document reflects v3 of the Go module, whose latest version is `v3.0.1` at the time of writing.
 
 ### Implicit and fixed parameters
 
--- a/data-structures/hashmap.md
+++ b/data-structures/hashmap.md
@@ -305,57 +305,43 @@ Rather than encoding the HAMT node's `map` field as the full sequence of byte-to
 
 The IPLD schema describing the Filecoin HAMT varies from the IPLD HashMap [schema](#Schema) and also between version of the Filecoin Actors, so any implementation needing to read Filecoin HAMT blocks will need to handle its specific layout (both layouts if historical data is important to create / read).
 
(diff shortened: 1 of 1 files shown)


mvdan:
I mentioned this two months ago, and I actually did the patch locally... then forgot to send the PR :see_no_evil:

mvdan:
Also, the repo has since moved to https://github.com/filecoin-project/go-hamt-ipld, and the old ipfs repo location doesn't have the v3 tag, so I'll fix the link too.

willscott (approved):

rvagg (approved):

data-structures/hashmap.md
```diff
@@ -290,7 +290,7 @@ As yet, there is no known hash collision attack vector against IPLD data structu
 
 The [Filecoin Project blockchain](https://filecoin-project.github.io/specs/) makes use of a HAMT that uses the same HAMT with buckets and CHAMP mutation semantics as outlined in this document. It encodes directly as [DAG-CBOR](../block-layer/codecs/dag-cbor.md) but uses a different block layout to the one specified here. This section documents the specific ways that the Filecoin HAMT variant differs from this specification. IPLD HashMap implementations may be able to implement a form that provides compatibility with Filecoin when requested by the user.
 
-The reference Go implementation for the Filecoin HAMT is used by the [Lotus](https://lotu.sh/) client and is available at <https://github.com/ipfs/go-hamt-ipld>.
+The reference Go implementation for the Filecoin HAMT is used by the [Lotus](https://lotu.sh/) client and is available at <https://github.com/filecoin-project/go-hamt-ipld>. Note that this document reflects v3 of the Go module, whose latest version is `v3.0.1` at the time of writing.
```
rvagg: did we actually get it updated for the v3 form? I don't even remember now

mvdan: You're absolutely right, I did not actually do that :facepalm:

This is what happens when I pick up a months-old branch and I forget that I hadn't finished it. Element was changed from keyed to kinded for v3, for example.

warpfork:
This seems thoroughly approved -- good to merge?

rvagg:
Can't be merged as is - the comment about being v3 is wrong, it's v2-compatible. It probably needs a change to have the v3+ version in here which is just a changed from the keyed to a kinded union but it'd be nice to show both. There's more overlap with the spec form now but still minor differences to note.

Either @mvdan or I need to get back to this and spend a tiny bit of time making sure it reflects reality and also documents the historical <=v2 forms as well.

mvdan:
I've made the PR a draft to be clearer. If it makes maintenance difficult, I don't have a problem with closing it as I can't say I'll finish it this week or next.

rvagg:
Rebased and added v3 details and some additional clarifications (like the `big.Int` bitfield). This is ready for review and potentially merging. PTAL @mvdan @warpfork.

data-structures/hashmap.md
```diff
@@ -2,35 +2,37 @@
 
 **Status: Prescriptive - Draft**
 
-* [Introduction](#Introduction)
```
rvagg, line 5: sorry for the noise here, editor settings changed and I'm just going with it.

data-structures/hashmap.md
```diff
@@ -295,27 +297,66 @@ The Filecoin HAMT _does not_ use an explicit root block (`HashMapRoot`) to encod
...
-An IPLD schema representing the Filecoin HAMT varies from the IPLD HashMap [schema](#Schema) so any implementation needing to read Filecoin HAMT blocks will need to handle its specific layout:
+The IPLD schema describing the Filecoin HAMT varies from the IPLD HashMap [schema](#Schema) and also between version of the Filecoin Actors, so any implementation needing to read Filecoin HAMT blocks will need to handle its specific layout (both layouts if historical data is important to create / read).
+
+The **v3** form of the Filecoin HAMT has the same IPLD layout as the IPLD HashMap except for the lack of an explicit root node to describe the parameters. A Filecoin HAMT's root is `HashMapNode`, which is the same as all other nodes and parameters are implicit.
+
+Note that there is currently no limitation on the types available for storage as `value`s as long as they can be decoded from the bytes. In practice, the Filecoin HAMT is used to store inline objects rather than links to objects.
+
+#### Actors v0.9 and v2 form
+
+Prior to **v3** the `Element` union was keyed with `"0"` and `"1"`.
```
mvdan: Did you consider keeping v2 as just this "keyed" note relative to v3? Because the schema is otherwise identical, so it's a bit repetitive. Assuming most people will be interested in v3, I'd lean towards keeping the spec doc shorter. The flip side is people wanting to copy-paste the v2 schema, so I agree with keeping this as-is if you were concerned about copy-pasting.

mvdan (commented):
LGTM, though it seems I can't approve my own PRs :) You'll have to approve.

data-structures/hashmap.md
rvagg: 0h, yeah, it's kind of verbose. And another problem is repeated symbols so if the .md is validated (both our schema implementations can extract schema from markdown docs) then you get overlapping symbols.
What I did in the Filecoin schema docs, inherited from @willscott's naming in statediff is versioned the symbols and I could do that here and only have the overlapping entries unversioned.  So we could have `FilecoinHAMTNodeV0` and `FilecoinHAMTNodeV3` as the different roots and they'd have `V0` and `V3` `Element` types but share `Bucket` and `Element` types. Gets kind of ugly but could be expressed without duplication and be properly unique throughout the doc.
@warpfork what think you in terms of expressiveness?

rvagg:
Latest commit unifies the schemas into one, renames the types to be what Filecoin calls them, and tries to explain the differences inline. PTAL @mvdan @warpfork and let me know if you think this is clear enough. Rendered @ https://github.com/ipld/specs/blob/hashmap-filecoin-v3/data-structures/hashmap.md#block-layout

rvagg:
merged with minor updates in new docs site https://github.com/ipld/ipld/pull/103, will tombstone this doc now

PR #360, opened 2021-03-26
block-layer/codecs: dag-pb can't enforce PBNode ordering

mvdan:
(see commit message)

### DIFF
--- a/block-layer/codecs/dag-pb.md
+++ b/block-layer/codecs/dag-pb.md
@@ -45,9 +45,10 @@ message PBNode {
 
 DAG-PB aims to have a **canonical form** for any given set of data. Therefore, in addition to the standard Protobuf parsing rules, DAG-PB decoders should enforce additional constraints to ensure canonical forms (where possible):
 
-1. Fields must appear in their correct order as defined by the Protobuf schema above, blocks with out-of-order fields should be rejected. It is common for Protobuf decoders to accept out-of-order field entries.
-2. Duplicate entries in the binary form are invalid, blocks with duplicate field values should be rejected. It is common for Protobuf decoders to accept _updates_ to fields that have already been set.
-3. Fields and wire types other than those that appear in the Protobuf schema above are invalid and blocks containing these should be rejected. It is common for Protobuf decoders to skip data in each message type that does not match the fields in the schema.
+1. Fields in the `PBLink` message must appear in the order as defined by the Protobuf schema above, following the field numbers. Blocks with out-of-order `PBLink` fields should be rejected.  (Note that it is common for Protobuf decoders to accept out-of-order field entries, which means the DAG-PB spec is somewhat stricter than may be seen as typical for other Protobuf-based formats.)
+2. Fields in the `PBNode` message must be encoded in the order as defined by the Protobuf schema above. Note that this order does not follow the field numbers. The decoder should accept either order, as IPFS data exists in both forms.
+3. Duplicate entries in the binary form are invalid; blocks with duplicate field values should be rejected. (Note that it is common for Protobuf decoders to accept repeated field values in the binary data, and interpret them as _updates_ to fields that have already been set; DAG-PB is stricter than this.)
+4. Fields and wire types other than those that appear in the Protobuf schema above are invalid and blocks containing these should be rejected. (Note that it is common for Protobuf decoders to skip data in each message type that does not match the fields in the schema.)
 
 ## Logical Format
 


willscott:
i'm not sure i agree with the conclusion of not having a canonical form?
just because we're willing to decode non-canonical stored representations doesn't mean there isn't a canonical form

willscott (approved):

mvdan:
I guess that depends on whether you define encoders that use a different order as spec-compliant or not. This is why I said the encoding "should" use a specific order, instead of "must".

If we don't get a review from Rod by the middle of next week, I can always remove that sentence to unblock a merge.

rvagg:
Yeah, I'm with @willscott on this one.

I think it should be "**must** be encoded ...", but "The decoder should accept either order", and also that there **is** a canonical form for given data. It's just that there exists some historical data which a decoder may encounter which is not in that canonical form. What we want here is for any new/current codec implementation to produce that canonical form, so round-trips of this historical data will end up as different bytes & with a different CID. We're not committing to preserving non-canonical forms in round-trips either (which is one way you could read this current wording I think).

mvdan:
Fair enough, that's a good point. Your wording does imply that someone using a "field number order" protobuf encoder is actively breaking the spec, but I think we're fine with that for the sake of strongly encouraging canonical forms.

mvdan:
Updated :) I'll prepare a branch for go-codec-dagpb on top of the performance PR, since otherwise I think I'll run into conflicts with myself.

block-layer/codecs/dag-pb.md
```diff
@@ -45,9 +45,10 @@ message PBNode {
 
 DAG-PB aims to have a **canonical form** for any given set of data. Therefore, in addition to the standard Protobuf parsing rules, DAG-PB decoders should enforce additional constraints to ensure canonical forms (where possible):
 
-1. Fields must appear in their correct order as defined by the Protobuf schema above, blocks with out-of-order fields should be rejected. It is common for Protobuf decoders to accept out-of-order field entries.
-2. Duplicate entries in the binary form are invalid, blocks with duplicate field values should be rejected. It is common for Protobuf decoders to accept _updates_ to fields that have already been set.
-3. Fields and wire types other than those that appear in the Protobuf schema above are invalid and blocks containing these should be rejected. It is common for Protobuf decoders to skip data in each message type that does not match the fields in the schema.
+1. Fields in the `PBLink` message must appear in the order as defined by the Protobuf schema above, following the field numbers. Blocks with out-of-order `PBLink` fields should be rejected. It is common for Protobuf decoders to accept out-of-order field entries.
```
warpfork:
```suggestion
1. Fields in the `PBLink` message must appear in the order as defined by the Protobuf schema above, following the field numbers. Blocks with out-of-order `PBLink` fields should be rejected.  (Note that it is common for Protobuf decoders to accept out-of-order field entries, which means the DAG-PB spec is somewhat stricter than may be seen as typical for other Protobuf-based formats.)
```

warpfork: I wonder if the "following the field numbers" part should be moved to a paranthetical or second sentence also, since I notice the line right below this has to make a "Note [...]" of the fact it *doesn't* follow the field numbers.

(But of course these are all nits on very fine details of prose, just striving for clarity but saying the same things already said.)

block-layer/codecs/dag-pb.md
```diff
@@ -45,9 +45,10 @@ message PBNode {
...
+2. Fields in the `PBNode` message must be encoded in the order as defined by the Protobuf schema above. Note that this order does not follow the field numbers. The decoder should accept either order, as IPFS data exists in both forms.
+3. Duplicate entries in the binary form are invalid, blocks with duplicate field values should be rejected. It is common for Protobuf decoders to accept _updates_ to fields that have already been set.
```
warpfork:
```suggestion
3. Duplicate entries in the binary form are invalid; blocks with duplicate field values should be rejected. (Note that it is common for Protobuf decoders to accept repeated field values in the binary data, and interpret them _updates_ to fields that have already been set; DAG-PB is stricter than this.)
```

(if I'm understanding protobuf correctly, that is)

block-layer/codecs/dag-pb.md
```diff
@@ -45,9 +45,10 @@ message PBNode {
...
+4. Fields and wire types other than those that appear in the Protobuf schema above are invalid and blocks containing these should be rejected. It is common for Protobuf decoders to skip data in each message type that does not match the fields in the schema.
```
warpfork: I'd kind of like to move the contextualizing commentary about protobuf norms to a paranthetical here, too, since it's more about protobuf norms than it is about DAG-PB.

warpfork (approved):
I'm really appreciative of the time spent to verify all these details and make sure our spec discusses them accurately and well!  :smile: 

---

Thinking-out-loud(-but-writing-it-down) for a bit, just for posterity:

- Most of the time (e.g., as I have with `*json` and `*cbor`), I think I'd argue that if the variation is common in the wild, and has any chance of being loadbearing, we should admit it and be able to round-trip it through the data model.
- ...But in this case, I think the former is sorta-true (but presumably there's a strong predominance to one ordering), and the latter we definitely know to be untrue (it's *not* allowed to be load-bearing (other than that it affects the hash)).

Resolutions which come from that:
- I have a harder time than usual imagining a practical call to have support for loading the non-canonical order and re-emitting that with our libraries.
- ... and therefore, yes, I feel comfortable with the strength of the wording proposed here about the necessity of encoding following the strict order.
- Basically +1 to what @rvagg  said.

---

I made a couple suggestions on the prose, mostly about clarifying some comments that are about protobuf norms, but the whole thing reads quite well already.  Nice work.

block-layer/codecs/dag-pb.md
mvdan: ack, done

mvdan:
I've applied all three of Eric's nits. It seems like we all agree, so I'm moving ahead. Thanks!

Issue #343, opened 2020-12-19
Confusion about envelope (and inline) representation of unions


patrsc:
I am confused about validation of envelope unions in IPLD schemas: The authoring guide (schemas/authoring-guide.md) suggests that

```json
{
  "msg": "All good",
  "tag": "progress",
  "payload": {
    "percent": 0.6,
    "last": "61626378797a"
  }
}
```

would pass validation in the example given there (https://github.com/ipld/specs/blob/master/schemas/authoring-guide.md#envelope), while the specification (https://github.com/ipld/specs/blob/master/schemas/representations.md#union-envelope-representation) looks (to me) like

```json
{
  "msg": "All good",
  "payload": {
    "tag": "progress",
    "payload": {
      "percent": 0.6,
      "last": "61626378797a"
    }
  }
}
```
would be the correct representation. Which representation is the correct one? If the first one is correct, I think it would be good to give such an example also in the specs in order to avoid confusion. [This](https://github.com/rvagg/js-ipld-schema-validator) IPLD schema validator treats the second representation as the correct one and refuses the first one.

Also the inline representation suffers from the same confusion.

rvagg:
hmm, I think @vmx pointed out some problems with the inline union docs that were certainly wrong but I don't recall whether we fixed that or not, or even noted it in this repo (I can't find it with a quick search and commits don't show anything!).

The envelope union is meant to have the contents _within_ a "content" key in a map, with the "discriminant" key telling you which type of the union it is. So the first example you've shown matches one with a `contentKey` of `tag` and `discriminantKey` of `payloa`, with the `"progress"` union having a form that being `type Progress struct { percent Float; last String }`. The second example would only work if you're focusing on the top level `payload` as a node to be a union. I don't think I see that example arising out of the docs you've pointed to but it sounds like we have ambiguity about it that we should fix. Perhaps you can help with some wording changes or point out the bits that are problematic?

The inline union is similar but pulls everything up into the current map. So it ditches `contentKey` because the map itself is the holder of the content. It just has one special key that is used as the `discriminantKey` to tell you what to interpret the map itself as. So the `payload` content is just pulled up into the parent map and the `tag` tells us how to interpret that map.

If it helps, the source of truth I use for all the code around this are mirrored here for JS and Go:

* https://github.com/ipld/js-ipld-schema/blob/master/test/fixtures/bulk
* https://github.com/ipld/go-ipld-schema/tree/master/test/fixtures/bulk

[...]

PR #367, opened 2021-05-02
Fixes to examples in schema authoring guide.

warpfork:
Attempted to fix up several of the examples in the schema authoring guide.  I tried to mostly leave the prose as-is, but updated the sections that seemed incongruous without updates.

(Where I updated prose, I added per-sentence line-breaking.)

I think this should fix some things pointed out in https://github.com/ipld/specs/issues/343.

### DIFF
--- a/schemas/authoring-guide.md
+++ b/schemas/authoring-guide.md
@@ -538,21 +538,25 @@ If we are designing the data layout for our example protocol (rather than consum
 
 #### Keyed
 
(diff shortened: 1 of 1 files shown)


rvagg (approved):
😬 these problems have been noted more than once, just never fixed. Thanks for fixing and making them clearer!

warpfork:
Thanks for the quick review :D

PR #372, opened 2021-06-22
tombstone for hashmap.md

rvagg:
first one, is this a pattern we're happy with moving forward @warpfork ?


warpfork:
Thank you :bow:

PR #374, opened 2021-07-12
tombstones for schema docs

rvagg:
have also put `# ..` and `{"error":"..."}` into .ipldsch and .json docs


schemas/examples.ipldsch.json
```diff
@@ -1,67 +1,3 @@
...
-					"fields": {
-						"fooField": {
-							"rename": "foo_field"
-						}
-					}
-				}
-			}
-		}
-	}
+  "error": "This document has moved: https://ipld.io/docs/schemas/"
```
willscott: The link here does not take you to an actual example of a json representation of a schema. This is strictly less useful.

schemas/schema-schema.ipldsch
```diff
@@ -1,704 +1 @@
...
-## assigns it to another type.
-##
-## The DSL defines a TypeCopy as `type NewThing = CopiedThing`, where
-## "CopiedThing" refers to a `type` defined elsewhere in a schema and is not
-## one of TypeKind or an inline type descriptor (`{}`, `[]`, `&`).
-##
-type TypeCopy struct {
-	fromType TypeName
-}
+# This document has moved: https://ipld.io/docs/schemas/
```
willscott: We should really maintain this in code in a canonical place, and i think this is the right place, so that other language implementations can base their schema parsing off of this definition.

The schema-schema is not linked or visible on the ipld.io docs page, so this is not a reasonable deprecation

warpfork: The bug here is that the linking in the new site didn't pick up the schema-schema file in its new home there.  It's a good callout, but we should just fix that there.

warpfork (approved):
I am fond of having less replicas of this information, and appreciate the redirections :+1:

willscott:
> I am fond of having less replicas of this information

we need a canonical raw instance of these files for use in APIs. that needs to be clearly and easily available. The PR in this form doesn't do that.

warpfork:
We have a canonical raw instance of these files.

It's https://github.com/ipld/ipld/blob/master/specs/schemas/schema-schema.ipldsch.json .

The linking in the website generation _was_ broken.  I'm fixing that in https://github.com/ipld/ipld/pull/116 .

rvagg:
OK so I've been specific in the redirects for .ipldsch and .ipldsch.json documents to their actual counterparts on ipld.io. I have left the others general, like I did for the CAR redirect, simply to give us flexibility to reorganise things over there and not have to come back here to fix up links every time we might do a minor shuffle. I'd rather we not shuffle (and am still not enthused about any of this stuff moving, but whatever) and keep stable URLs, but I think we've proven as an org that we're not super good at that kind of stability (maybe we just need to point to CIDs of these things instead).

Thoughts?

PR #376, opened 2021-09-08
Update README to denote deprecation.

BigLep:
Being done as part of https://github.com/ipld/ipld/issues/124


warpfork:
I took this on separately and have applied a big heap of diffs to the main branch now :)  but thanks for the kick in the shorts to get it done!

PR #377, opened 2021-11-10
Signal ADLs in selectors

willscott:
Introduce an `ExploreInterpretAs` wrapper in selectors to allow points in a traversal where the current `Node` should be `Reified` using identified, well-known ADLs.

Example:
```
{
	"~": {
               "c": 0xabc00002,
               ">": { ".": {} }
        }
}
```


willscott:
wrong repo

PR #378, opened 2022-06-28
feat: redirect to ipld.io/specs/, with some specifics

rvagg:
Uses a (custom) vuepress plugin to overwrite all generated pages with a new redirect
html that does a JS redirect in 5 seconds after explaining it. Some pages have
specific redirects to their new versions, some are done in groups (e.g.
schemas pages all go to /specs/schemas/), and the remainder just go to
ipld.io/specs.

Closes: https://github.com/protocol/infra/issues/818

---

I unarchived this repo to get this in. After merge, the Github Action in here will build the site and push to gh-pages which will replace the old HTML with this new one. I'll re-archive after merge.

Most interesting 2 pieces in this PR are the redirect template, and the list of redirects in the array. I've mostly plucked out important pages and easy redirects. Many are just going to get a general redirect to ipld.io/specs though.

---
e.g. of the dag-cbor page:

Screenshot 2022-06-28 at 4 08 29 pm

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -52,5 +52,8 @@ module.exports = {
   extendMarkdown: md => {
     // use more markdown-it plugins!
     md.use(linkfix)
-  }
+  },
+  plugins: [
+    require('./redirects.js')
+  ]
 }
--- a/.vuepress/redirect_template.html
+++ b/.vuepress/redirect_template.html
@@ -3,6 +3,8 @@
 <head>
   <meta charset="UTF-8">
   <title>This page has moved...</title>
+  <meta http-equiv="refresh" content="7;url={{redirectUrl}}">
+  <link rel="canonical" href="{{redirectUrl}}">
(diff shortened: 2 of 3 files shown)


lidel (approved):
Pushed small fix in https://github.com/ipld/specs/pull/378/commits/f2b40a87b87524d6cc2b2a7bee0a56373b3a5b67:
- html redirect after 7s in case browser had JS disabled
- [canonical link](https://en.wikipedia.org/wiki/Canonical_link_element) will be read by google crawlers and merge old links into new ones on results pages
- `location.replace` is bit better because it also replaces entry in browser history

I was not able to build locally – probably requires ancient Node version, hopefully Github Action still works :crossed_fingers: 

Otherwise, lgtm, thank you @rvagg !

rvagg:
beautiful, thanks for your expertise on this one @lidel

rvagg:
seems to be working as expected

